PR #33, opened 2015-10-01
set a default value for the status

antoine-de:
I don't understand how the status  can be nullable and have a none
value, but since we don't handle all IRE flow for the moment, we can
have 'none'


buildbotctp:
Build finished. 25 tests run, 0 skipped, 0 failed.

PR #35, opened 2015-10-05
Add a contributor for realtime_updates

kinnou02:
this way we can choose who receive this info


kirin/ire/ire.py
```diff
@@ -42,6 +42,7 @@ def _make_rt_update(data):
     Create an RealTimeUpdate object for the query and persist it
     """
     rt_update = model.RealTimeUpdate(data, connector='ire')
+    rt_update.contributor = current_app.config.get('CONTRIBUTOR')
```
antoine-de: it's optional ?

PR #37, opened 2015-10-06
Get activate real_time_update objects

azime:
Select all real_time_update objects by contributor, start and end date.

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -117,7 +117,6 @@ class StopTimeUpdate(db.Model, TimestampMixin):
         self.arrival_status = other.arrival_status
 
 
-
 associate_realtimeupdate_tripupdate = db.Table('associate_realtimeupdate_tripupdate',
                                     db.metadata,
                                     db.Column('real_time_update_id', postgresql.UUID, db.ForeignKey('real_time_update.id')),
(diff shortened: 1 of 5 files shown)


kirin/core/model.py
```diff
@@ -192,3 +190,16 @@ def __init__(self, raw_data, connector,
         self.status = status
         self.error = error
         self.received_at = received_at
+
+    def __repr__(self):
+        return '<RealTimeUpdate %r>' % self.id
+
+    @classmethod
+    def all(cls, contributor, start_date=datetime.datetime.now().date(), end_date=None):
```
TeXitoi: Using now as default seems strange. A coherent default, for me, would be -infinity

kirin/core/model.py
```diff
@@ -192,3 +190,16 @@ def __init__(self, raw_data, connector,
...
+        query = cls.query.filter_by(contributor=contributor)
+        query = query.join(cls.trip_updates)
```
kinnou02: All the indexes for this request exist?

kirin/core/model.py
kinnou02: We will want more than one contributor

azime:
Index added

PR #42, opened 2015-10-15
Revert trip id fallback

pbougue:
reverts #41
To be merged after delivery of navitia in internal (navitia v1.28)


xlqian:
ok for me

PR #45, opened 2015-10-22
read delay in ire

antoine-de:
http://jira.canaltp.fr/browse/NAVP-353

### DIFF
--- a/kirin/ire/model_maker.py
+++ b/kirin/ire/model_maker.py
@@ -1,3 +1,4 @@
+# coding=utf-8
 # Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
 #
 # This file is part of Navitia,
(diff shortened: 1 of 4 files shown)


antoine-de:
Jenkins, test this please

kirin/ire/model_maker.py
```diff
@@ -222,3 +244,40 @@ def get_navitia_stop_time(self, downstream_point, nav_vj):
...
+
+        Note: if the status ('Etat') is 'supprimé' the delay is 0
+
+        return a pair with the new datetime and it's status
+        """
+        if xml is None:
+            return datetime.combine(date, navitia_time), 'none'
+
+        if get_value(xml, 'Etat') == 'supprimé':
+            delay = 0
```
kinnou02: it's not what we want, we must not override the previous state if we had one!

PR #46, opened 2015-10-26
Delay handling

antoine-de:
Handle delay on StopTimeUpdates

refactoring of the handle.merge to make it works.

We need to merge the info from 3 sources:
* the navitia base schedule
* the trip update already in the bd (potentially not existent)
* the incoming trip update
The result is either the db_trip_update if it exists, or the new_trip_update (it is updated as a side
effect)
The mechanism is quite simple:
* the result trip status is the new_trip_update's status
        (ie in the db the trip was cancelled, and a new update is only an update, the trip update is
        not cancelled anymore, only updated)
* for each navitia's stop_time and for departure|arrival:
    - if there is an update on this stoptime (in new_trip_update):
            we compute the new datetime based on the new information and the navitia's base schedule
    - else if there if the stoptime in the db:
            we keep this db stoptime
    - else we keep the navitia's base schedule

**Important Note**:
we DO NOT HANDLE changes in navitia's schedule for the moment
it will need to be handled, but it will be done after

fix http://jira.canaltp.fr/browse/NAVP-354


kinnou02:
Jenkins, test this please

kirin/core/handler.py
```diff
@@ -70,36 +70,96 @@ def handle(real_time_update, trip_updates, contributor):
...
-            #TODO handle past midnigth
-            if departure_time:
-                departure = datetime.datetime.combine(trip_update.vj.circulation_date, departure_time)
-            if arrival_time:
-                arrival = datetime.datetime.combine(trip_update.vj.circulation_date, arrival_time)
-
-            st = StopTimeUpdate(navitia_stop['stop_point'], departure, arrival)
-            #we want to keep the order
-            trip_update.stop_time_updates.insert(idx, st)
+        nav_departure_time = navitia_stop.get('departure_time')
```
kinnou02: You should check pick_up and drop_off too

migrations/versions/51d2d334dd4f_.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '51d2d334dd4f'
+down_revision = 'd6469f134d7'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('stop_time_update', sa.Column('order', sa.Integer(), nullable=False))
```
kinnou02: You need to do this in three step:
  - delete all stop time since there are false
  - add the column with a server default
  - remove the server default

antoine-de:
Jenkins, test this please

kirin/core/handler.py
```diff
@@ -70,36 +70,97 @@ def handle(real_time_update, trip_updates, contributor):
...
+    we DO NOT HANDLE changes in navitia's schedule for the moment
+    it will need to be handled, but it will be done after
+    """
+    res = db_trip_update if db_trip_update else new_trip_update
+    res_stoptime_updates = []
+
+    res.status = new_trip_update.status
+    if new_trip_update.message:
+        res.message = new_trip_update.message
+    res.contributor = res.contributor
```
TeXitoi: I don't this this line is so much useful.

kirin/core/handler.py
```diff
@@ -70,36 +70,97 @@ def handle(real_time_update, trip_updates, contributor):
...
+
+    The mechanism is quite simple:
+        * the result trip status is the new_trip_update's status
+            (ie in the db the trip was cancelled, and a new update is only an update, the trip update is
+            not cancelled anymore, only updated)
+
+        * for each navitia's stop_time and for departure|arrival:
+            - if there is an update on this stoptime (in new_trip_update):
+                we compute the new datetime based on the new information and the navitia's base schedule
+            - else if there if the stoptime in the db:
```
TeXitoi: else if there **is** ?

kirin/core/handler.py
```diff
@@ -70,36 +70,97 @@ def handle(real_time_update, trip_updates, contributor):
...
+        nav_departure_time = navitia_stop.get('departure_time')
+        nav_arrival_time = navitia_stop.get('arrival_time')
+
+        departure = arrival = None
+        if nav_departure_time:
+            departure = datetime.datetime.combine(new_trip_update.vj.circulation_date, nav_departure_time)
+        if nav_arrival_time:
+            arrival = datetime.datetime.combine(new_trip_update.vj.circulation_date, nav_arrival_time)
+
+        #TODO handle past midnight
```
TeXitoi: Doesn't it works out of the box ?

antoine-de: no idea :smile:

pytest.ini
```diff
@@ -1,2 +1,3 @@
 [pytest]
 addopts = --doctest-modules
+norecursedirs = .venv migrations
```
TeXitoi: no new line at EOF

PR #47, opened 2015-10-28
Use a named queue for load_realtime

kinnou02:
This way multiple kirin can share the queue on split the work between
them


antoine-de:
Jenkins, test this please

PR #52, opened 2015-11-03
receive_at was initialised one time only at startup

kinnou02:


kirin/core/model.py
```diff
@@ -212,10 +212,10 @@ class RealTimeUpdate(db.Model, TimestampMixin):
...
 
-    def __init__(self, raw_data, connector, status=None, error=None, received_at=datetime.datetime.now()):
+    def __init__(self, raw_data, connector, status=None, error=None, received_at=None):
         self.id = gen_uuid()
         self.raw_data = raw_data
         self.connector = connector
         self.status = status
         self.error = error
-        self.received_at = received_at
+        self.received_at = received_at if received_at else datetime.datetime.utcnow()
```
pbougue: Is it possible to test that? hard to test the "now", but maybe that the date is not None, and otherwise that the date is ok?

kinnou02: It like testing the language itself.

PR #59, opened 2015-12-07
Trip update consistency

pbougue:
First version testing consistency of trip_update (mostly delays) just before persisting and issuing gtfs-rt

JIRA: http://jira.canaltp.fr/browse/NAVP-395

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -45,6 +45,54 @@ def persist(real_time_update):
     model.db.session.commit()
 
 
+def manage_consistency(trip_update):
+    """
+    receive a TripUpdate, then manage and adjust it's consistency
+    returns False if trip update cannot be managed
+    """
+    current_order = 0
+    previous_stu = None
+    for stu in trip_update.stop_time_updates:
+        # rejections
+        if stu.order != current_order:
+            return False
+
+        if not stu.arrival and not stu.departure:
+            return False
+
+        # modifications
+        if not stu.arrival:
+            stu.arrival = stu.departure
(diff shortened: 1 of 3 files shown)


kirin/core/handler.py
```diff
@@ -45,6 +45,54 @@ def persist(real_time_update):
...
 
 
+def manage_consistency(trip_update):
+    """
+    receive a TripUpdate, then manage and adjust it's consistency
+    returns False if trip update cannot be managed
+    """
+    current_order = 0
+    previous_stu = None
+    for stu in trip_update.stop_time_updates:
```
antoine-de: you can do  `for stu, current_order in enumerate(trip_update.stop_time_updates):`

kirin/core/handler.py
```diff
@@ -45,6 +45,54 @@ def persist(real_time_update):
...
+        # rejections
+        if stu.order != current_order:
+            return False
```
antoine-de: I think we have to log the bad cases to be able to trace them

kirin/core/handler.py
```diff
@@ -45,6 +45,54 @@ def persist(real_time_update):
...
+
+        if not stu.arrival and not stu.departure:
+            return False
+
+        # modifications
+        if not stu.arrival:
+            stu.arrival = stu.departure
+            if not stu.arrival_delay:
+                stu.arrival_delay = stu.departure_delay
```
antoine-de: same here, I think we need to log when we change the results

pbougue:
Jenkins, test this please

PR #64, opened 2016-01-12
handle pastmidnight vjs

antoine-de:
past-midnight were not handled in kirin, creating weird delays (spanning on 4 days for my example).

We need to clear the kirin db with those weird delays

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -27,6 +27,7 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 import logging
+from datetime import timedelta
 import pytz
 import kirin
 from kirin import gtfs_realtime_pb2
@@ -199,6 +200,8 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
         res.stop_time_updates = []
         return res
 
+    last_nav_dep = None
+    circulation_date = new_trip_update.vj.circulation_date
     for navitia_stop in navitia_vj.get('stop_times', []):
         stop_id = navitia_stop.get('stop_point', {}).get('id')
         new_st = new_trip_update.find_stop(stop_id)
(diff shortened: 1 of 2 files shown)


kirin/core/handler.py
```diff
@@ -209,13 +212,17 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
...
 
-        departure = arrival = None
-        if nav_departure_time:
-            departure = _get_datetime(new_trip_update.vj.circulation_date, nav_departure_time, timezone)
+        arrival = departure = None
         if nav_arrival_time:
-            arrival = _get_datetime(new_trip_update.vj.circulation_date, nav_arrival_time, timezone)
-
-        #TODO handle past midnight
+            if last_nav_dep and last_nav_dep > nav_arrival_time:
```
pbougue: Isn't there any way to manage vj longer than 24h ?

pbougue:
Seems OK to me.
Maybe add a test on pass midnight between 2 stops... But the code is almost the same.

antoine-de:
yes I wondered about this as well but the code would be copy/pasted

as you wish

kirin/core/handler.py
```diff
@@ -209,13 +212,17 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
...
+                # last departure is after arrival, it's a past-midnight
+                circulation_date += timedelta(days=1)
+            arrival = _get_datetime(circulation_date, nav_arrival_time, timezone)
+        if nav_departure_time:
+            if nav_arrival_time and nav_arrival_time > nav_departure_time:
```
pbougue: You can have `None` in `nav_arrival_time`... In which case I don't know how it will be handled.

xlqian: @pbougue `nav_arrival_time` is actually tested before the comparison

PR #66, opened 2016-02-08
Remove an idle transaction in load_realtime

kinnou02:
Flask magically handle db transaction by associating a session to
the flask context. Flask script create one context for the duration of
the command, so the session and by extension the transaction is kept for
all the lifetime of load_realtime, so theoretically indefinitely.

If we remove the session after we are done with it, the transaction if
rollback and we are happy.

Refer to: http://jira.canaltp.fr/browse/NAVITIAII-2027


pbougue:
Ok for me

PR #67, opened 2016-02-25
More exception trace in db

pbougue:
handle other exceptions in db + longer timeout for navitia

related to JIRA: http://jira.canaltp.fr/browse/NAVITIAII-2059

### DIFF
--- a/kirin/ire/ire.py
+++ b/kirin/ire/ire.py
@@ -70,6 +70,7 @@ class Ire(Resource):
 
     def __init__(self):
         self.navitia_wrapper = make_navitia_wrapper()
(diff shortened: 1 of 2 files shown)


kirin/ire/ire.py
```diff
@@ -70,6 +70,7 @@ class Ire(Resource):
 
     def __init__(self):
         self.navitia_wrapper = make_navitia_wrapper()
+        self.navitia_wrapper.timeout = current_app.config.get('NAVITIA_TIMEOUT', 5)
```
kinnou02: You should add it to the default_settings

pbougue: Done

PR #74, opened 2017-01-31
Fix case of ignored delay at 0

kinnou02:
timedelta at zero are False :(

### DIFF
--- a/tests/integration/ire_test.py
+++ b/tests/integration/ire_test.py
@@ -84,6 +84,66 @@ def test_ire_post_no_data():
         assert len(TripUpdate.query.all()) == 0
         assert len(StopTimeUpdate.query.all()) == 0
 
+def check_db_ire_96231_normal():
+    with app.app_context():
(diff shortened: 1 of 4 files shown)


tests/integration/ire_test.py
```diff
@@ -84,6 +84,63 @@ def test_ire_post_no_data():
...
+        assert second_st.message == 'Affluence exceptionnelle de voyageurs'
+
+        # last stop is gare de Basel-SBB, delay's only at the arrival
+        last_st = db_trip_delayed.stop_time_updates[-1]
+        assert last_st.stop_id == 'stop_point:OCE:SP:TrainTER-85000109'
+        assert last_st.arrival == datetime.datetime(2015, 9, 21, 16, 39)
+        assert last_st.arrival_status == 'update'
+        assert last_st.arrival_delay == timedelta(minutes=0)
+        #The departure is consistent with arrival
+        try:
```
antoine-de: can you add a small additional comment here ?

I fear it will be difficult to understand this in the futur :confused:

PR #81, opened 2017-05-17
hardcode the component name

antoine-de:
for kibana


Hoshiyo:
Already into tag field, so don't need it

PR #83, opened 2017-09-05
first rough implem for gtfs-rt

antoine-de:
WIP with a lots of empty boxes

to use, get a tripupdate and you can do (for the moment)

`http POST :5001/gtfs_rt @tripUpdates.pb`

need tests and still lots of TODOs

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -221,7 +221,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):
     raw_data = db.Column(db.Text, nullable=True)
 
     trip_updates = db.relationship("TripUpdate", secondary=associate_realtimeupdate_tripupdate, cascade='all',
-                                   lazy='joined', backref=backref('real_time_updates', cascade='all'))
+                                   lazy='select', backref=backref('real_time_updates', cascade='all'))
 
     __table_args__ = (db.Index('realtime_update_created_at', 'created_at'),)
 
--- a/kirin/default_settings.py
+++ b/kirin/default_settings.py
@@ -20,6 +20,12 @@ NAVITIA_TOKEN = os.getenv('KIRIN_NAVITIA_TOKEN', None)
 CONTRIBUTOR = os.getenv('KIRIN_CONTRIBUTOR', 'realtime.ire')
 
 
+# TODO better conf for multi GTFS-RT
+NAVITIA_GTFS_RT_INSTANCE = os.getenv('KIRIN_NAVITIA_GTFS_RT_INSTANCE', 'sherbrooke')
+NAVITIA_GTFS_RT_TOKEN = os.getenv('KIRIN_NAVITIA_GTFS_RT_TOKEN', None)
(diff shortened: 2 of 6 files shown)


kirin/default_settings.py
```diff
@@ -20,6 +20,12 @@
 CONTRIBUTOR = os.getenv('KIRIN_CONTRIBUTOR', 'realtime.ire')
 
 
+# TODO better conf for multi GTFS-RT
+NAVITIA_GTFS_RT_INSTANCE = os.getenv('KIRIN_NAVITIA_GTFS_RT_INSTANCE', 'sherbrooke')
+NAVITIA_GTFS_RT_TOKEN = os.getenv('KIRIN_NAVITIA_GTFS_RT_INSTANCE', None)
```
kinnou02: token

kirin/gtfs_rt/model_maker.py
```diff
@@ -26,19 +26,114 @@
...
+    return date.strftime("%Y%m%dT%H%M%S")
+
 
 class KirinModelBuilder(object):
 
     def __init__(self, nav, contributor=None):
         self.navitia = nav
         self.contributor = contributor
+        self.log = logging.getLogger(__name__)
+        self.period_filter_tolerance = datetime.timedelta(hours=3)
```
xlqian: i seems to me that it's a constant, can we make period_filter_tolerance configurable?

kirin/gtfs_rt/model_maker.py
```diff
@@ -26,19 +26,114 @@
...
         """
         parse the protobuf in the rt_update object
         and return a list of trip updates
 
         The TripUpdates are not yet associated with the RealTimeUpdate
         """
-        # TODO!
-        pass
+        self.log.info("proto = {}".format(data))
+        data_time = datetime.datetime.utcfromtimestamp(data.header.timestamp)
```
xlqian: can this line raise an exception if timestamp is not well formatted? (according to the doc, `ValueError` may be raised if the timestamp is out of the range of `(1970, 2038)`, just in case,  someone inputs a bad value...)

antoine-de: I think a ValueError is nice for this, the data will not be integrated and an error will be returned (and logged)

is it ok for you @kinnou02 ?

xlqian: ok!

kirin/core/model.py
```diff
@@ -221,7 +221,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):
     raw_data = db.Column(db.Text, nullable=True)
 
     trip_updates = db.relationship("TripUpdate", secondary=associate_realtimeupdate_tripupdate, cascade='all',
-                                   lazy='joined', backref=backref('real_time_updates', cascade='all'))
+                                   lazy='select', backref=backref('real_time_updates', cascade='all'))
```
kinnou02: This should'nt cause any performance regression

kirin/gtfs_rt/model_maker.py
kinnou02: yeah why not

PR #85, opened 2017-09-11
First implementation of workers

kinnou02:
- [x] update requirements
- [x] add scheduler
- [x] improve log

### DIFF
--- a/kirin/gtfs_rt/tasks.py
+++ b/kirin/gtfs_rt/tasks.py
@@ -48,21 +48,17 @@ class InvalidFeed(Exception):
 
 @celery.task(bind=True)
 def gtfs_poller(self, config):
-    logger = logging.getLogger(__name__)
-    logger.info('polling of %s', config['feed_url'])
-    response = requests.get(config['feed_url'])
-    if response.status_code != 200:
-        raise InvalidFeed()
+    logger =  logging.LoggerAdapter(logging.getLogger(__name__), extra={'contributor': config['contributor']})
+    logger.debug('polling of %s', config['feed_url'])
+    response = requests.get(config['feed_url'], timeout=config.get('timeout', 1))
+    response.raise_for_status()
 
     nav = navitia_wrapper.Navitia(url=config['navitia_url'], token=config['token'])\
                          .instance(config['coverage'])
     nav.timeout = 5
 
     proto = gtfs_realtime_pb2.FeedMessage()
-    try:
-        proto.ParseFromString(response.content)
-    except DecodeError:
-        raise InvalidFeed()
+    proto.ParseFromString(response.content)
     model_maker.handle(proto, navitia_wrapper, config['contributor'])
     logger.debug('gtfsrt polling finished')
 
--- a/kirin/gtfs_rt/tasks.py
+++ b/kirin/gtfs_rt/tasks.py
@@ -1,0 +1,68 @@
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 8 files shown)


kirin/gtfs_rt/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+
+@celery.task(bind=True)
+def gtfs_poller(self, config):
+    logger = logging.getLogger(__name__)
+    logger.info('polling of %s', config['feed_url'])
+    response = requests.get(config['feed_url'])
+    if response.status_code != 200:
+        raise InvalidFeed()
+
+    nav = navitia_wrapper.Navitia(url=config['navitia_url'], token=config['token'])\
```
antoine-de: I think we can move this in the model_maker

kirin/helper.py
```diff
@@ -0,0 +1,49 @@
...
+
+def make_celery(app):
+    celery_app = celery.Celery(app.import_name, broker=app.config['CELERY_BROKER_URL'])
+    celery_app.conf.update(app.config)
+    TaskBase = celery_app.Task
+
+    class ContextTask(TaskBase):
+        abstract = True
+
+        def __init__(self):
```
antoine-de: do we need this ? and if yes, don't we need to call the parent constructor ?

kirin/tasks.py
```diff
@@ -0,0 +1,67 @@
...
+    # Flask SQLAlchemy will automatically create new sessions for you from
+    # a scoped session factory, given that we are maintaining the same app
+    # context, this ensures tasks have a fresh session (e.g. session errors
+    # won't propagate across tasks)
+    model.db.session.remove()
+
+
+
+from kirin.gtfs_rt.tasks import gtfs_poller
+@celery.task(bind=True)
```
antoine-de: by curiosity what does this `bind` do ?

kirin/gtfs_rt/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+from google.protobuf.message import DecodeError
+
+
+class InvalidFeed(Exception):
+    pass
+
+@celery.task(bind=True)
+def gtfs_poller(self, config):
+    logger = logging.getLogger(__name__)
+    logger.info('polling of %s', config['feed_url'])
```
antoine-de: I don't you said there is more work on the logging, but just to log it (:wink:), I think this should be debug

kirin/gtfs_rt/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+        raise InvalidFeed()
+
+    nav = navitia_wrapper.Navitia(url=config['navitia_url'], token=config['token'])\
+                         .instance(config['coverage'])
+    nav.timeout = 5
+
+    proto = gtfs_realtime_pb2.FeedMessage()
+    try:
+        proto.ParseFromString(response.content)
+    except DecodeError:
```
antoine-de: we should log this

kirin/gtfs_rt/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+
+class InvalidFeed(Exception):
+    pass
+
+@celery.task(bind=True)
+def gtfs_poller(self, config):
+    logger = logging.getLogger(__name__)
+    logger.info('polling of %s', config['feed_url'])
+    response = requests.get(config['feed_url'])
+    if response.status_code != 200:
```
antoine-de: we should log at least if there is something usefull in the response

kirin/gtfs_rt/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+        raise InvalidFeed()
```
xlqian: it's more like a `PollFailure` than a `InvalidFeed`, no?

kirin/helper.py
```diff
@@ -0,0 +1,49 @@
...
+            pass
+
+        def __call__(self, *args, **kwargs):
+            with app.app_context():
```
xlqian: why do you need to context the task? because of the app.config?

kinnou02: http://flask.pocoo.org/docs/0.12/patterns/celery/

kirin/helper.py
kinnou02: http://flask.pocoo.org/docs/0.12/patterns/celery/

kirin/tasks.py
kinnou02: this `bind` give us access to the task by the self parameter

kirin/helper.py
antoine-de: I don't see it in the doc :laughing:

kirin/tasks.py
antoine-de: but you don't need self no ?

PR #91, opened 2017-09-22
purge db trip_update

xlqian:
A little problem with the current archi:

For now, we have only one worker listening to one queue to which the scheduler pushes jobs(poller, purge_real_time_update, purge_trip_update). So, a heavy job will block all the other jobs, for instance, a poller job will be blocked by a purge job during 20 min until the purge is finished...

several solutions are possible:

1. Make the only worker concurrent with "-c N" (-P gevent shut down the worker's concurrency)
2. Spawn several workers
3. Make dedicated workers for example, a worker only for poller and a worker only for puge jobs

To test:

1. What will happen when poller and purge are executed at the same time? (reading, writing and deleting at the same time)
2. What will happen when purge and load_realtime are executed at the same time? (reading and deleting at the same time)


kirin/core/model.py
```diff
@@ -188,13 +189,25 @@ def find_by_dated_vj(cls, navitia_trip_id, vj_circulation_date):
...
         query = cls.query.filter(cls.contributor.in_(contributors))
         if start_date:
-            query = query.filter("vehicle_journey_1.circulation_date >= '{start_date}'".
-                                 format(start_date=start_date))
+            query = query.filter(sqlalchemy.text("vehicle_journey_1.circulation_date >= '{start_date}'".
+                                 format(start_date=start_date)))
         if end_date:
-            query = query.filter("vehicle_journey_1.circulation_date <= '{end_date}'".
-                                 format(end_date=end_date))
+            query = query.filter(sqlalchemy.text("vehicle_journey_1.circulation_date <= '{end_date}'".
```
kadhikari: Since circulation_date doesn't have hour value in database, it's better to exclude "=" in the filter as following:
"vehicle_journey_1.circulation_date < '{end_date}'"

The purge after mid-night will not have any history for "NB_DAYS_TO_KEEP_TRIP_UPDATE = 1"

kadhikari:
test this please

kirin/default_settings.py
```diff
@@ -29,7 +29,7 @@
 NAVITIA_GTFS_RT_TOKEN = os.getenv('KIRIN_NAVITIA_GTFS_RT_TOKEN', None)
 GTFS_RT_CONTRIBUTOR = os.getenv('KIRIN_GTFS_RT_CONTRIBUTOR', 'realtime.sherbrooke')
 GTFS_RT_FEED_URL = os.getenv('KIRIN_GTFS_RT_FEED_URL', None)
-
+NB_DAYS_TO_KEEP_TRIP_UPDATE = 2
```
kinnou02: add a `getenv` plz

xlqian:
@kinnou02 it's done now

kinnou02:
by default celery start with a concurency equal to the number of processes on the machine, so adding the `-c` should'nt change anything.
But since we use gevent it only start multiple greenlet, that may explain why it wait for the end of the purge

xlqian:
@kinnou02 I just find that this [line](https://github.com/CanalTP/kirin/blob/master/kirin/__init__.py#L41) shut down worker's concurrency

xlqian:
@kinnou02  sorry, the '-P gevent' and the monkey patch in kirin/__ini__.py shut down worker's concurrency

kirin/default_settings.py
```diff
@@ -129,4 +130,14 @@
         'schedule': timedelta(seconds=60),
         'options': {'expires': 30}
     },
+    'purge_gtfs_trip_update': {
+        'task': 'kirin.tasks.purge_gtfs_trip_update',
+        'schedule': timedelta(days=1),
```
kinnou02: we should schedule it at a fixed time, this way we will know when the purge will run

kirin/command/reset_task_locks.py
```diff
@@ -0,0 +1,41 @@
...
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from kirin import manager, app, redis
+import logging
+
+
+@manager.command
+def reset_task_locks():
```
pbougue: as discussed will not be used so far:
* so we remove that in this PR (maybe keep it aside if we do something later)
* we add a timeout to the locks so that they disappear

PR #95, opened 2017-10-02
fix disconnection of worker (probably) caused by gevent

xlqian:


kinnou02:
Can you update the deployement too plz

Procfile
```diff
@@ -1,5 +1,5 @@
-web: ./manage.py runserver
+web: USE_GEVENT=true ./manage.py runserver
 load_realtime: ./manage.py load_realtime
```
kinnou02: We will need gevent on this one too I think

Procfile
```diff
@@ -1,5 +1,5 @@
-web: ./manage.py runserver
+web: USE_GEVENT=true ./manage.py runserver
```
kinnou02: Also, you should prefix it with `KIRIN_`

kirin/core/handler.py
```diff
@@ -50,7 +50,7 @@ def persist(real_time_update):
 
 def log_stu_modif(trip_update, stu, string_additional_info):
     logger = logging.getLogger(__name__)
-    logger.info("TripUpdate {vj_id} on {date}, StopTimeUpdate {order} modified: {add_info}".format(
+    logger.debug("TripUpdate {vj_id} on {date}, StopTimeUpdate {order} modified: {add_info}".format(
```
pbougue: 👍

kirin/gtfs_rt/model_maker.py
```diff
@@ -86,7 +86,7 @@ def build(self, rt_update, data):
 
         The TripUpdates are not yet associated with the RealTimeUpdate
         """
-        self.log.info("proto = {}".format(data))
+        self.log.debug("proto = {}".format(data))
```
pbougue: 👍

xlqian:
@kinnou02 here it is: https://github.com/CanalTP/fab_kirin/pull/15

Procfile
```diff
@@ -1,5 +1,5 @@
-web: ./manage.py runserver
-load_realtime: ./manage.py load_realtime
-worker: celery worker -P gevent -A kirin.tasks.celery
+web: KIRIN_USE_GEVENT=true ./manage.py runserver
+load_realtime: KIRIN_USE_GEVENT=true ./manage.py load_realtime
+worker: celery worker -A kirin.tasks.celery -c 3
```
pbougue: To be cautious, this should be `-c 1`:
* it's none of this PR's responsibility to deal with that
* bigger problem: we're not sure how kirin will react with 2 poller polling the same thing at the same time (which might happen if poll is slower than beat)

PR #96, opened 2017-10-04
change use gevent setting

xlqian:


kirin/default_settings.py
```diff
@@ -30,7 +30,8 @@
 GTFS_RT_CONTRIBUTOR = os.getenv('KIRIN_GTFS_RT_CONTRIBUTOR', 'realtime.sherbrooke')
 GTFS_RT_FEED_URL = os.getenv('KIRIN_GTFS_RT_FEED_URL', None)
 
-USE_GEVENT = os.getenv('KIRIN_USE_GEVENT', False)
+USE_GEVENT = os.getenv('KIRIN_USE_GEVENT', 'false').lower() == 'true'
```
pbougue: @kinnou02 mentionned that there is something in flask for that
also, you can add log at startup so that we know once the setup of those variables :)

kirin/__init__.py
```diff
@@ -75,6 +73,13 @@
     app.logger.addHandler(handler)
     app.logger.setLevel('INFO')
 
+# We need to log all kinds of patch, all patch must be done as soon as possible
+logger = logging.getLogger(__name__)
+if 'threading' not in sys.modules:
+    logger.info('threading is deleted from import')
+if app.config['USE_GEVENT']:
```
pbougue: I'd rather read all the config dict (evene if a bit dirty).

PR #99, opened 2017-10-10
use ConsumerProducerMixin

xlqian:


kirin/rabbitmq_handler.py
```diff
@@ -41,95 +42,113 @@
...
+    def get_consumers(self, Consumer, channel):
+        return [Consumer(
+            queues=[self.rpc_queue],
+            on_message=self.on_request,
+            prefetch_count=1,
+        )]
+
+    def on_request(self, message):
+
+        retrying.Retrying(stop_max_attempt_number=5,
```
kinnou02: since we already have a retry policy on the publish I don't think we have to retry here too

kirin/rabbitmq_handler.py
```diff
@@ -41,95 +42,113 @@
 from kirin.utils import str_to_date
 from socket import error
 import time
+from kombu.mixins import ConsumerProducerMixin
+
+
+class RTReloader(ConsumerProducerMixin):
```
kinnou02: :+1:

xlqian:
jenkins, test it please!

PR #112, opened 2017-10-20
add more logs when a vj cannot be found in navitia

xlqian:
In many cases, it's impossible to find a VJ supplied by GTFS-RT and at the same time, all of its delay are zero, so it's ok that the vj cannot be found. some logs are added to get alerted in cases where these delays are NOT zero

### DIFF
--- a/kirin/gtfs_rt/model_maker.py
+++ b/kirin/gtfs_rt/model_maker.py
@@ -107,6 +107,17 @@ class KirinModelBuilder(object):
         vjs = self._get_navitia_vjs(input_trip_update.trip, data_time=data_time)
 
         trip_updates = []
+
+        # we may be not able to find the vj in navitia, but if all delay of the trip_update is 0, it's just ok,
+        # otherwise we need to analyze this case
+        if not vjs:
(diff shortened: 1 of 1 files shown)


kirin/gtfs_rt/model_maker.py
```diff
@@ -107,6 +107,17 @@ def _make_trip_updates(self, input_trip_update, data_time):
         vjs = self._get_navitia_vjs(input_trip_update.trip, data_time=data_time)
 
         trip_updates = []
+
+        # we may be not able to find the vj in navitia, but if all delay of the trip_update is 0, it's just ok,
+        # otherwise we need to analyze this case
+        if not vjs:
+            if all((stu.arrival.delay == 0 for stu in input_trip_update.stop_time_update)):
+                self.log.debug('can not find vj %s, but all its stop_time_update delays are zero',
```
pbougue: Hmm maybe a debug in that case, but for the rest (vj not found, or multiple vjs), I think it's dealt with in _get_navitia_vjs, and it's an info, it's not an error from kirin, it's just an inconsistency detection of fully external values.
Maybe we can add some info to the logs there, but I'm not sure there's a point in this PR.

PR #113, opened 2017-10-20
TripUpdate's start_date must be UTC when passing to kraken (2/3)

pbougue:
tests and bugfix

Read all in #115 (last one)

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -27,6 +27,7 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 from datetime import timedelta
+from pytz import utc
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.orm import backref, deferred
 from sqlalchemy.ext.orderinglist import ordering_list
@@ -73,21 +74,23 @@ class VehicleJourney(db.Model):
     """
     id = db.Column(postgresql.UUID, default=gen_uuid, primary_key=True)
     navitia_trip_id = db.Column(db.Text, nullable=False)
+    start_timestamp = db.Column(db.DateTime, nullable=True) # timestamp of VJ's start
     circulation_date = db.Column(db.Date, nullable=False)  # in local timezone
 
-    __table_args__ = (db.UniqueConstraint('navitia_trip_id', 'circulation_date', name='vehicle_journey_navitia_trip_id_circulation_date_idx'),)
+    __table_args__ = (db.UniqueConstraint('navitia_trip_id', 'circulation_date',
+                                          name='vehicle_journey_navitia_trip_id_circulation_date_idx'),)
 
-    def __init__(self, navitia_vj, circulation_date):
+    def __init__(self, navitia_vj, local_circulation_date):
         self.id = gen_uuid()
         if 'trip' in navitia_vj and 'id' in navitia_vj['trip']:
             self.navitia_trip_id = navitia_vj['trip']['id']
-        self.circulation_date = circulation_date
+        self.circulation_date = local_circulation_date
+
+        # dummy start_timestamp
+        self.start_timestamp = utc.localize(datetime.datetime.combine(local_circulation_date,
+                                                                      datetime.time(0, 0)))
         self.navitia_vj = navitia_vj  # Not persisted
 
-    @classmethod
-    def purge(cls, until):
-        for vj in cls.query.filter(cls.circulation_date < until):
-            db.session.delete(vj)
 
 
 class StopTimeUpdate(db.Model, TimestampMixin):
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -88,7 +88,11 @@ class VehicleJourney(db.Model):
     """
     id = db.Column(postgresql.UUID, default=gen_uuid, primary_key=True)
     navitia_trip_id = db.Column(db.Text, nullable=False)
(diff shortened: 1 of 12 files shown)


kirin/core/model.py
```diff
@@ -181,19 +219,21 @@ def __repr__(self):
         return '<TripUpdate %r>' % self.vj_id
 
     @classmethod
-    def find_by_dated_vj(cls, navitia_trip_id, vj_circulation_date):
+    def find_by_dated_vj(cls, navitia_trip_id, start_timestamp):
         return cls.query.join(VehicleJourney).filter(VehicleJourney.navitia_trip_id == navitia_trip_id,
-                                              VehicleJourney.circulation_date == vj_circulation_date).first()
+                                              VehicleJourney.start_timestamp == start_timestamp).first()
```
kinnou02: if for any reason we have one sec of offset we won't find the vj

migrations/versions/2762a0e3f0bf_add_constraints_start_timestamp.py
```diff
@@ -0,0 +1,48 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # needs to be done again in case some were created during previous upgrade
+    op.execute("UPDATE vehicle_journey SET start_timestamp=circulation_date + TIME '00:00:00' "
+               "WHERE start_timestamp IS NULL")
+    # add constraints on start_timestamp
+    op.alter_column('vehicle_journey', 'start_timestamp', nullable=False)
```
kinnou02: so a this point the kirin actually deployed in production stop working.

migrations/versions/2761a0e3f0bf_add_vj_start_timestamp.py
```diff
@@ -0,0 +1,27 @@
...
+revision = '2761a0e3f0bf'
+down_revision = '37444da1fe56'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('vehicle_journey', sa.Column('start_timestamp', sa.DateTime(), nullable=True))
+    op.execute("UPDATE vehicle_journey SET start_timestamp = circulation_date + TIME '00:00:00' "
```
kinnou02: what happen to the record inserted right after this request but before the the update of the application?

kirin/core/model.py
```diff
@@ -67,27 +68,64 @@ class TimestampMixin(object):
...
 
-    def __init__(self, navitia_vj, circulation_date):
         self.id = gen_uuid()
         if 'trip' in navitia_vj and 'id' in navitia_vj['trip']:
             self.navitia_trip_id = navitia_vj['trip']['id']
-        self.circulation_date = circulation_date
+        self.circulation_date = local_circulation_date #only for retrocompatibility
+
+        first_stop_time = navitia_vj.get('stop_times', [{}])[0]
+        start_time = first_stop_time['arrival_time']
```
xlqian: you might want to test first_stop_time before calling it or use `get` instead of `[]`

or 
`start_time = first_stop_time.get('arrival_time') or first_stop_time.get('departure_time')`

kirin/core/model.py
```diff
@@ -67,27 +68,64 @@ class TimestampMixin(object):
...
+        tzinfo = get_timezone(first_stop_time)
+        self.start_timestamp = tzinfo.localize(datetime.datetime.combine(local_circulation_date,
+                                                                         start_time))
         self.navitia_vj = navitia_vj  # Not persisted
 
-    @classmethod
-    def purge(cls, until):
-        for vj in cls.query.filter(cls.circulation_date < until):
-            db.session.delete(vj)
+    def get_start_timestamp(self):
```
xlqian: if `start_timestamp` is always converted to a utc when reading, you can use `@property`

kirin/core/model.py
pbougue: Yep, and we will recreate one: but as unicity constraint is also migrated, it should be fine (so we might need to trace something about that at some point if we get bugs, but don't know what right now)

kirin/core/model.py
```diff
@@ -181,19 +219,21 @@ def __repr__(self):
...
-                                              VehicleJourney.circulation_date == vj_circulation_date).first()
+                                              VehicleJourney.start_timestamp == start_timestamp).first()
 
     @classmethod
     def find_by_contributor_period(cls, contributors, start_date=None, end_date=None):
         query = cls.query.filter(cls.contributor.in_(contributors))
         if start_date:
-            query = query.filter(sqlalchemy.text("vehicle_journey_1.circulation_date >= '{start_date}'".
-                                 format(start_date=start_date)))
+            start_dt = datetime.datetime.combine(start_date, datetime.time(0, 0))
```
xlqian: why?

pbougue: This is used for full-reload: we are now comparing with timestamp (used to be date), so I wanted the same behaviour, and explicitly implemented it.

kirin/core/model.py
pbougue: OK to discuss about that

kirin/core/model.py
pbougue: don't know if it's worth avoiding crash... maybe only crash more explicitly

xlqian: looks good to me

PR #115, opened 2017-10-20
Finish TripUpdate's start_date is UTC (3/3)

pbougue:
Final migration
To be merged after that #114 then #113 are merged.

Read by commit (so you can follow migration, as they are PR): have to be retrocompatible n <-> n-1 when deploying

* Read carefully model.py first.
* New tests are only in gtfs_rt_test.py (others are "only" updated).
* SQL migration may be "too" smart? Hope it's correctly anticipated.

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -93,8 +93,6 @@ class VehicleJourney(db.Model):
     # timestamp of VJ's start
     start_timestamp = db.Column(db.DateTime, nullable=False) # ! USE get_start_timestamp() !
 
-    circulation_date = db.Column(db.Date, nullable=True) #only for retrocompatibility
-
     __table_args__ = (db.UniqueConstraint('navitia_trip_id', 'start_timestamp',
                                           name='vehicle_journey_navitia_trip_id_start_timestamp_idx'),)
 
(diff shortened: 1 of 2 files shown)


kirin/core/model.py
```diff
@@ -67,27 +68,62 @@ class TimestampMixin(object):
...
+        tzinfo = get_timezone(first_stop_time)
+        self.start_timestamp = tzinfo.localize(datetime.datetime.combine(local_circulation_date,
+                                                                         start_time))
         self.navitia_vj = navitia_vj  # Not persisted
 
-    @classmethod
-    def purge(cls, until):
-        for vj in cls.query.filter(cls.circulation_date < until):
-            db.session.delete(vj)
+    def get_start_timestamp(self):
```
xlqian: If i understand well, you can indeed use `@property` for start_timestamp

kirin/core/model.py
```diff
@@ -67,27 +68,62 @@ class TimestampMixin(object):
...
+
+    def get_local_circulation_date(self):
+        from kirin.utils import get_timezone
+
+        if self.navitia_vj is None:
+            return None
+
+        first_stop_time = self.navitia_vj.get('stop_times', [{}])[0]
+        tzinfo = get_timezone(first_stop_time)
+        return self.get_start_timestamp().astimezone(tzinfo).date()
```
xlqian: wait, isn't `self.start_timestamp` already localized? (https://github.com/CanalTP/kirin/pull/115/files#diff-d582ffa4339c008d83001df44608e3b8R108)

pbougue: depends if we get it from db, in which case it's naive datetime (not localized)

kirin/core/model.py
pbougue: 👍 (will do it in the PR 2/3)

xlqian:
Jenkins, test this please!

PR #116, opened 2017-10-23
cache call to navitia with redis

kinnou02:
On my laptop a call without cache take 75sec, with cache it take 15sec.
The main problem is that the datetime of the request are computed from
'now' so even with the rounding of the since and until, the cache will
be invalidated every hour :(


kirin/gtfs_rt/model_maker.py
```diff
@@ -123,8 +123,8 @@ def _make_trip_updates(self, input_trip_update, data_time):
     def _get_navitia_vjs(self, trip, data_time):
         vj_source_code = trip.trip_id
 
-        since = data_time - self.period_filter_tolerance
-        until = data_time + self.period_filter_tolerance
+        since = round_datetime(data_time - self.period_filter_tolerance, -1)
```
xlqian: :+1:

kirin/gtfs_rt/tasks.py
```diff
@@ -67,6 +67,9 @@ def gtfs_poller(self, config):
         nav = navitia_wrapper.Navitia(url=config['navitia_url'], token=config['token'])\
                              .instance(config['coverage'])
         nav.timeout = 5
+        pubdate_cache_timeout = app.config.get('NAVITIA_PUBDATE_CACHE_TIMEOUT', 600)
```
xlqian: `PUBDATE` i'm not sure if it's a typo?

kirin/default_settings.py
```diff
@@ -51,6 +51,11 @@
 
 ENABLE_RABBITMQ = boolean(os.getenv('KIRIN_ENABLE_RABBITMQ', True))
 
+NAVITIA_QUERY_CACHE_TIMEOUT = int(os.getenv('KIRIN_NAVITIA_QUERY_CACHE_TIMEOUT', 60*60*24))  # in seconds
```
xlqian: minor: `NAVITIA_QUERY_CACHE_TIMEOUT = int(os.getenv('KIRIN_NAVITIA_QUERY_CACHE_TIMEOUT', timedelta(days=1).total_seconds()))`

kirin/gtfs_rt/tasks.py
kinnou02: pubdate for publication date, but it's a bit too long for my taste

kirin/gtfs_rt/model_maker.py
pbougue: I'd remove that `-1`

pbougue: minor: Also I wouldn't deal this offset into round function (or maybe a round_up/down kind of parameter...). As is, a `floor` function would be a good name

PR #120, opened 2017-10-24
fix status when some contributor didn't have any realtime_update created

kinnou02:


pbougue:
Jenkins, test this please.

PR #131, opened 2017-11-09
Fix order of stop times from lollipop

kadhikari:
Use of stop_id and order to manage lollipop problem

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -266,7 +266,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
     local_circulation_date = new_trip_update.vj.get_local_circulation_date()
 
     has_changes = False
-    for navitia_stop in navitia_vj.get('stop_times', []):
+    for nav_order, navitia_stop in enumerate(navitia_vj.get('stop_times', [])):
         # TODO handle forbidden pickup/dropoff (in those case set departure/arrival at None)
         nav_departure_time = navitia_stop.get('departure_time')
         nav_arrival_time = navitia_stop.get('arrival_time')
@@ -288,14 +288,14 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
             base_departure = _get_datetime(local_circulation_date, nav_departure_time, timezone)
 
         stop_id = navitia_stop.get('stop_point', {}).get('id')
-        new_st = new_trip_update.find_stop(stop_id, len(res_stoptime_updates))
+        new_st = new_trip_update.find_stop(stop_id, nav_order)
 
         if db_trip_update is not None and new_st is not None:
             """
             First case: we already have recorded the delay and we find update info in the new trip update
             Then      : we should probably update it or not if the input info is exactly the same as the one in db
             """
-            db_st = db_trip_update.find_stop(stop_id, len(res_stoptime_updates))
+            db_st = db_trip_update.find_stop(stop_id, nav_order)
             new_st_update = _make_stop_time_update(base_arrival,
                                                    base_departure,
                                                    last_departure,
@@ -316,7 +316,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
                                             last_departure,
                                             new_st,
                                             navitia_stop['stop_point'],
-                                            order=len(res_stoptime_updates))
+                                            order= nav_order)
             res_st.message = new_st.message
 
         elif db_trip_update is not None and new_st is None:
@@ -328,11 +328,11 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
                         
                         *** Here, we MUST NOT do anything, only update stop time's order ***
             """
-            db_st = db_trip_update.find_stop(stop_id, len(res_stoptime_updates))
+            db_st = db_trip_update.find_stop(stop_id, nav_order)
             res_st = db_st if db_st is not None else StopTimeUpdate(navitia_stop['stop_point'],
                                                                     departure=base_departure,
                                                                     arrival=base_arrival,
-                                                                    order=len(res_stoptime_updates))
+                                                                    order=nav_order)
             new_order = len(res_stoptime_updates)
             # We need to deal with this case more carefully, this won't work if the trip is a lollipop 
             has_changes |= (db_st is None) or db_st.order != new_order
@@ -347,9 +347,9 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
             res_st = StopTimeUpdate(navitia_stop['stop_point'],
                                     departure=base_departure,
                                     arrival=base_arrival,
-                                    order=len(res_stoptime_updates))
+                                    order = nav_order)
 
-        #res_st.order = len(res_stoptime_updates)
+        res_st.order = nav_order
         last_departure = res_st.departure
         res_stoptime_updates.append(res_st)
         last_nav_dep = nav_departure_time
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -288,14 +288,14 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
             base_departure = _get_datetime(local_circulation_date, nav_departure_time, timezone)
 
         stop_id = navitia_stop.get('stop_point', {}).get('id')
-        new_st = new_trip_update.find_stop(stop_id)
+        new_st = new_trip_update.find_stop(stop_id, len(res_stoptime_updates))
 
         if db_trip_update is not None and new_st is not None:
             """
             First case: we already have recorded the delay and we find update info in the new trip update
             Then      : we should probably update it or not if the input info is exactly the same as the one in db
             """
-            db_st = db_trip_update.find_stop(stop_id)
+            db_st = db_trip_update.find_stop(stop_id, len(res_stoptime_updates))
             new_st_update = _make_stop_time_update(base_arrival,
                                                    base_departure,
                                                    last_departure,
@@ -328,7 +328,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
                         
                         *** Here, we MUST NOT do anything, only update stop time's order ***
             """
-            db_st = db_trip_update.find_stop(stop_id)
+            db_st = db_trip_update.find_stop(stop_id, len(res_stoptime_updates))
             res_st = db_st if db_st is not None else StopTimeUpdate(navitia_stop['stop_point'],
                                                                     departure=base_departure,
                                                                     arrival=base_arrival,
@@ -336,7 +336,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
             new_order = len(res_stoptime_updates)
             # We need to deal with this case more carefully, this won't work if the trip is a lollipop 
             has_changes |= (db_st is None) or db_st.order != new_order
-            res_st.order = new_order
+            #res_st.order = new_order
 
         else:
             """
@@ -349,6 +349,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
                                     arrival=base_arrival,
                                     order=len(res_stoptime_updates))
 
+        #res_st.order = len(res_stoptime_updates)
         last_departure = res_st.departure
         res_stoptime_updates.append(res_st)
         last_nav_dep = nav_departure_time
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -133,10 +133,6 @@ def handle(real_time_update, trip_updates, contributor):
     id_timestamp_tuples = [(tu.vj.navitia_trip_id, tu.vj.get_start_timestamp()) for tu in trip_updates]
     old_trip_updates = TripUpdate.find_by_dated_vjs(id_timestamp_tuples)
     for trip_update in trip_updates:
-        #Nothing to do if trip_update state is none and it doesn't have any stop_time_update
-        if (trip_update.status == 'none') and (len(trip_update.stop_time_updates) == 0):
-            continue
-
         # find if there is already a row in db
         old = next((tu for tu in old_trip_updates if tu.vj.navitia_trip_id == trip_update.vj.navitia_trip_id
                     and tu.vj.get_start_timestamp() == trip_update.vj.get_start_timestamp()), None)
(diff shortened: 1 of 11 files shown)


buildbotctp:
Build finished.

buildbotctp:
Build finished.

kirin/core/handler.py
```diff
@@ -347,8 +347,9 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
             res_st = StopTimeUpdate(navitia_stop['stop_point'],
                                     departure=base_departure,
                                     arrival=base_arrival,
-                                    order=len(res_stoptime_updates))
+                                    order = nav_order)
 
+        res_st.order = nav_order
```
xlqian: i don't think this is necessary...

kadhikari: You are right

tests/mock_navitia/vj_bad_order.py
```diff
@@ -0,0 +1,329 @@
...
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+import navitia_response
+
+response = navitia_response.NavitiaResponse()
+
+response.query = 'vehicle_journeys/?filter=vehicle_journey.has_code(source, Code-orders)' \
+                 '&depth=2&since=20120615T060000Z&until=20120615T130000Z'
+#resquest time is UTC -> 10:00 to 17:00 local
```
pbougue: If it's Sherbrooke, `6:00Z` is `2:00` local

tests/mock_navitia/vj_lollipop.py
```diff
@@ -0,0 +1,290 @@
...
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+import navitia_response
+
+response = navitia_response.NavitiaResponse()
+
+response.query = 'vehicle_journeys/?filter=vehicle_journey.has_code(source, Code-lollipop)' \
+                 '&depth=2&since=20120615T060000Z&until=20120615T130000Z'
+#resquest time is UTC -> 22:00 to 05:00 local
```
pbougue: Comment outdated here as well.

tests/integration/gtfs_rt_test.py
```diff
@@ -609,3 +609,406 @@ def test_gtfs_rt_partial_update_diff_feed_2(partial_update_gtfs_rt_data_2,
...
+
+
+'''
+vj.stop_times:  StopR1	StopR2	StopR3	StopR4	StopR5	StopR6
+order:          0       1       2       3       4       5
+
+gtfs-rt.stop:           StopR2  StopR3  Stop-RT-1   StopR4  StopR6
+stop_sequence:          2       3       4           5       6
+
+Stop-Match:     StopR1	StopR2	StopR3	StopR4	StopR5	StopR6
```
pbougue: OK, so this is the result we hope, this can be clarified in comment (same on other tests).
Also, we probably don't want that result exactly :)

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,6 +105,32 @@ def build(self, rt_update, data):
             trip_updates.extend(tu)
         return trip_updates
 
+    '''
+    We verify  strict match list of trip_update stop codes(trip_update_stop_list) with the list of vehiclejourney
+    stop codes(vehicle_journey_stop_list) from the last element towards the left the left. if trip_update_stop_list is
+    a subset of vehicle_journey_stop_list from the last element, the gtfs-rt trip update is valid for treatment.
+    '''
+    def _match_gtfsrt(self, main_list, sublist):
```
xlqian: doc string should be put inside of the function

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,6 +105,32 @@ def build(self, rt_update, data):
             trip_updates.extend(tu)
         return trip_updates
 
+    '''
+    We verify  strict match list of trip_update stop codes(trip_update_stop_list) with the list of vehiclejourney
+    stop codes(vehicle_journey_stop_list) from the last element towards the left the left. if trip_update_stop_list is
```
xlqian: from the last element towards the left the left???

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,6 +105,32 @@ def build(self, rt_update, data):
...
+    a subset of vehicle_journey_stop_list from the last element, the gtfs-rt trip update is valid for treatment.
+    '''
+    def _match_gtfsrt(self, main_list, sublist):
+        if len(sublist) > len(main_list):
+            return False, -1
+        if main_list == sublist:
+            return True, 0
+        if sublist == main_list[len(main_list) - len(sublist):len(main_list)]:
```
xlqian: `main_list[len(main_list) - len(sublist):] == sublist`

xlqian: slicing in python is very poor, keep in mind that a slicing will return a copy of references of the original list... 
```
all((x==y for x,y in itertools.izip(reversed(l1), reversed(l2))))
```
it should run more faster since there are only iterators....

pbougue: > And it's a better test also, we are sure it's a ending sublist (not in the middle)

kirin/gtfs_rt/model_maker.py
```diff
@@ -113,8 +139,28 @@ def _make_trip_updates(self, input_trip_update, data_time):
...
-            for input_st_update in input_trip_update.stop_time_update:
+            #Initialize stops absent in trip_updates but present in vj
+            for order, stop in enumerate(vj_stop_points):
+                if order < first_index:
+                    st_update = self._init_stop_update(stop, order)
+                    if st_update is not None:
+                        trip_update.stop_time_updates.append(st_update)
+
+            #Manage stops present in gtfs_rt and navitia_vj
+            for order, input_st_update in enumerate(input_trip_update.stop_time_update):
```
pbougue: I think it would be better to merge a bit things, because here we get 1 loop, we already did 1 in _match_gtfsrt, and 1 more in _match_gtfsrt, and 1 for init. Those 4 can probably be merged in one (erroring when there's a problem in sublist, and maybe managing each test in a function for clarity).
But we have to keep that 0 delay for those missing first STU (managed by init so far).

kirin/gtfs_rt/model_maker.py
```diff
@@ -223,17 +274,18 @@ def read_delay(st_event):
...
         return st_update
 
-    def _get_navitia_stop_time(self, input_st_update, navitia_vj):
-        # TODO use input_st_update.stop_sequence to get the right stop_time even for loops
-        for s in navitia_vj.get('stop_times', []):
-            if any(c['type'] == self.stop_code_key and c['value'] == input_st_update.stop_id
+    def _get_navitia_stop_time(self, stop_id, stop_sequence, navitia_vj):
+        # To handle a vj with the same stop served multiple times(lollipop) we match the stop_point and order of navitia
+        # with stop_id and stop_sequence of trip_update of gtfs-rt
+        for order, s in enumerate(navitia_vj.get('stop_times', [])):
```
pbougue: You can just access to `navitia_vj.get('stop_times', [])[stop_sequence]` I think (and test it's the correct stop).

kirin/gtfs_rt/model_maker.py
xlqian: I've benched that method and curiously it doesn't do what i think, it runs much slower than the `==`. :( 

but i find an another method :
`all(x == y for x, y in itertools.izip(islice(l1, len(l1) - len(l2)), l2))`, this method is 2x faster than the simple `==` when len(l1) and len(l2) are close (no allocation of the new list) BUT the `==` is much faster when `len(l1)` is much larger than `len(l2)` (the cost of allocation is much smaller )

here is the bench https://ideone.com/17nbAr

xlqian: you can stay with `main_list[len(main_list) - len(sublist):] == sublist` since in practice, sublist is much smaller than main_list and the code stay simple

kirin/core/model.py
```diff
@@ -275,9 +275,16 @@ def remove_by_contributors_and_period(cls, contributors, start_date=None, end_da
 
         db.session.commit()
 
-    def find_stop(self, stop_id):
-        #TODO: we will need to handle vj who deserve the same stop multiple times
-        return next((st for st in self.stop_time_updates if st.stop_id == stop_id), None)
+    def find_stop(self, stop_id, order=None):
+        # To handle a vj with the same stop served multiple times(lollipop) we search first with stop_id and order
+        # For IRE since we dont care of the order search only with stop_id if no element found
+        # Note: Stops with mis-matched order between navitia-II and gtfs-rt are ignored in model_maker
```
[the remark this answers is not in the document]
kadhikari: ok

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,6 +105,32 @@ def build(self, rt_update, data):
...
+        #list of vj.stops
+        vj_stop_codes = []
+        vj_stop_points = []
+        for s in navitia_vj.navitia_vj.get('stop_times', []):
+            for c in s.get('stop_point', {}).get('codes', []):
+                if c['type'] == self.stop_code_key:
+                    vj_stop_codes.append(c['value'])
+                    vj_stop_points.append(s.get('stop_point'))
+                    continue
+        return vj_stop_codes, vj_stop_points
```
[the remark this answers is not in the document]
kadhikari: Since these lists are used in different ways, i left them unchanged

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,63 +106,63 @@ def build(self, rt_update, data):
...
         vj_stop_source_codes = []
         vj_stop_points = []
+        vj_sp_with_code = []
         for s in navitia_vj.get('stop_times', []):
             for c in s.get('stop_point', {}).get('codes', []):
                 if c['type'] == self.stop_code_key:
                     vj_stop_source_codes.append(c['value'])
                     vj_stop_points.append(s.get('stop_point'))
+                    vj_sp_with_code.append((c['value'], s.get('stop_point')))
                     continue
```
xlqian: the `continue` doesn't do what you want! you may want a `break` here

kadhikari:
With a gtfs-rt having 222 trip_updates, kirin **branch version** takes between 5 to 6 seconds where as **this pull request** takes ~4.5 seconds.

kirin/gtfs_rt/model_maker.py
```diff
@@ -114,11 +133,39 @@ def _make_trip_updates(self, input_trip_update, data_time):
...
+                        trip_update.stop_time_updates.append(st_update)
+                vj_stop_order -= 1
+
+            if not is_tu_valid:
+                self.log.error('stop_time_update do not match with stops in navitia for trip : {}'
+                               .format(input_trip_update.trip.trip_id))
+                record_internal_failure('stop_time_update do not match with stops in navitia',
+                                        contributor=self.contributor)
+                del trip_update.stop_time_updates[:]
+                trip_update.status == 'none'
```
pbougue: We probably also want to delete that trip_update, no?
Otherwise, we might end up merging some "none" trip and in that case to me it's more like if we got no info at all about that trip... So `return None` and no add in `build()` if None.

tests/integration/gtfs_rt_test.py
```diff
@@ -112,8 +117,9 @@ def test_gtfs_model_builder(basic_gtfs_rt_data):
     """
     test the model builder with a simple gtfs-rt
 
-    we have realtime data on only 2 stops, so the model builder should only have 2 stops (even if the VJ
+    we have realtime data on only 3 stops, so the model builder should only have 3 stops (even if the VJ
```
pbougue: Outdated comment: The model builder should have 4 stops (and that's what you test)

tests/integration/gtfs_rt_test.py
```diff
@@ -126,17 +132,17 @@ def test_gtfs_model_builder(basic_gtfs_rt_data):
         db.session.commit()
 
         assert len(trip_updates) == 1
-        assert len(trip_updates[0].stop_time_updates) == 2
+        assert len(trip_updates[0].stop_time_updates) == 4
 
-        second_stop = trip_updates[0].stop_time_updates[0]
+        second_stop = trip_updates[0].stop_time_updates[1]
```
pbougue: You may add a test on stu[0] and check that delay is 0

tests/integration/gtfs_rt_test.py
```diff
@@ -152,7 +158,7 @@ def test_gtfs_rt_simple_delay(basic_gtfs_rt_data, mock_rabbitmq):
     """
     test the gtfs-rt post with a simple gtfs-rt
 
-    we have realtime data on only 2 stops, so the model builder should only have 2 stops (even if the VJ
+    we have realtime data on only 3 stops, so the model builder should only have 3 stops (even if the VJ
```
pbougue: Same: outdated comment

tests/integration/gtfs_rt_test.py
```diff
@@ -328,9 +334,9 @@ def test_gtfs_rt_pass_midnight(pass_midnight_gtfs_rt_data, mock_rabbitmq):
     """
     test the gtfs-rt post with a pass-midnight gtfs-rt
 
-    we have realtime data on only the 4 stops
+    we have realtime data on only the 5 stops
```
pbougue: on `all` 5 stops

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,20 +106,64 @@ def build(self, rt_update, data):
...
+
+                if tu_stop is not None:
+                    if self._get_stop_code(vj_stop_point) != tu_stop.stop_id:
+                        is_tu_valid = False
+                        break
+
+                    tu_stop.stop_sequence = vj_stop_order
+                    st_update = self._make_stoptime_update(tu_stop, vj_stop_point)
+                    if st_update is not None:
+                        trip_update.stop_time_updates.append(st_update)
```
pbougue: So the StopTimeUpdate list is reversed here.
It's OK for now as we random access (using index) in merge(), but if someone thinks it is ordered his code would break.
@kinnou02 @xlqian @TeXitoi Do you think we should have an ordered list of STU?

pbougue: Everyone seems to agree that it would be better ordered, so it would probably be the last thing to do on this PR.

kadhikari: Done

PR #138, opened 2017-12-28
Add an attribute validity in real_time_update and use it in status.last_valid_update. 

kadhikari:
https://jira.canaltp.fr/browse/NAVP-723

```json
{
  "last_update": {
    "realtime.sherbrooke": "2018-01-26T15:42:20Z"
  }, "last_update_error": {
    "realtime.sherbrooke": "Http Error"
  }, "last_valid_update": {
    "realtime.sherbrooke": "2018-01-26T15:41:24Z"
  }
}
```

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -305,7 +305,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):
     error = db.Column(db.Text, nullable=True)
     raw_data = deferred(db.Column(db.Text, nullable=True))
     contributor = db.Column(db.Text, nullable=True)
-    validity = db.Column(db.Boolean, nullable=True)
+    validity = db.Column(db.Enum('valid', 'empty', 'invalid', name='rt_validity'), nullable=True)
 
     trip_updates = db.relationship("TripUpdate", secondary=associate_realtimeupdate_tripupdate, cascade='all',
                                    lazy='select', backref=backref('real_time_updates', cascade='all'))
@@ -321,17 +321,17 @@ class RealTimeUpdate(db.Model, TimestampMixin):
         self.error = error
         self.contributor = contributor
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
-        self.validity = True
+        self.validity = 'valid'
 
     @classmethod
-    def get_last_update_by_contributor(cls, validity):
+    def get_last_update_by_contributor(cls, is_valid=False):
         from kirin import app
         result = {}
         contributor = [app.config['CONTRIBUTOR'], app.config['GTFS_RT_CONTRIBUTOR']]
         for c in contributor:
             sql = db.session.query(db.func.max(cls.created_at)).filter(cls.contributor == c)
-            if validity:
-                sql = sql.filter(cls.validity == validity)
+            if is_valid:
+                sql = sql.filter(cls.validity == 'valid')
             row = sql.one()
             if row[0]:
                 result[c] = row[0].strftime('%Y-%m-%dT%H:%M:%SZ')
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -305,7 +305,6 @@ class RealTimeUpdate(db.Model, TimestampMixin):
     error = db.Column(db.Text, nullable=True)
     raw_data = deferred(db.Column(db.Text, nullable=True))
     contributor = db.Column(db.Text, nullable=True)
-    validity = db.Column(db.Enum('valid', 'empty', 'invalid', name='rt_validity'), nullable=True)
 
     trip_updates = db.relationship("TripUpdate", secondary=associate_realtimeupdate_tripupdate, cascade='all',
                                    lazy='select', backref=backref('real_time_updates', cascade='all'))
@@ -313,7 +312,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):
     __table_args__ = (db.Index('realtime_update_created_at', 'created_at'),
                       db.Index('realtime_update_contributor_and_created_at', 'created_at', 'contributor'))
 
-    def __init__(self, raw_data, connector, contributor, status=None, error=None, received_at=None):
+    def __init__(self, raw_data, connector, contributor, status='OK', error=None, received_at=None):
         self.id = gen_uuid()
         self.raw_data = raw_data
         self.connector = connector
@@ -321,17 +320,16 @@ class RealTimeUpdate(db.Model, TimestampMixin):
         self.error = error
         self.contributor = contributor
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
-        self.validity = 'valid'
 
     @classmethod
-    def get_last_update_by_contributor(cls, is_valid=False):
+    def get_last_update_by_contributor(cls, only_valid=False):
         from kirin import app
         result = {}
         contributor = [app.config['CONTRIBUTOR'], app.config['GTFS_RT_CONTRIBUTOR']]
         for c in contributor:
             sql = db.session.query(db.func.max(cls.created_at)).filter(cls.contributor == c)
-            if is_valid:
-                sql = sql.filter(cls.validity == 'valid')
+            if only_valid:
+                sql = sql.filter(cls.status == 'OK')
             row = sql.one()
             if row[0]:
                 result[c] = row[0].strftime('%Y-%m-%dT%H:%M:%SZ')
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -33,6 +33,8 @@ from sqlalchemy.ext.orderinglist import ordering_list
 from flask_sqlalchemy import SQLAlchemy
 import datetime
 import sqlalchemy
+from sqlalchemy import desc
+
 
 db = SQLAlchemy()
 
@@ -327,12 +329,17 @@ class RealTimeUpdate(db.Model, TimestampMixin):
         result = {}
         contributor = [app.config['CONTRIBUTOR'], app.config['GTFS_RT_CONTRIBUTOR']]
         for c in contributor:
-            sql = db.session.query(db.func.max(cls.created_at)).filter(cls.contributor == c)
+            sql = db.session.query(cls.created_at, cls.error).filter(cls.contributor == c)
             if only_valid:
                 sql = sql.filter(cls.status == 'OK')
-            row = sql.one()
-            if row[0]:
-                result[c] = row[0].strftime('%Y-%m-%dT%H:%M:%SZ')
+            sql = sql.order_by(desc(cls.created_at))
+            row = sql.first()
+            if row:
+                error_message = row[1]
+                if error_message:
+                    result[c] = row[0].strftime('%Y-%m-%dT%H:%M:%SZ') + ' Error: {}'.format(error_message)
+                else:
+                    result[c] = row[0].strftime('%Y-%m-%dT%H:%M:%SZ')
         return result
 
     @classmethod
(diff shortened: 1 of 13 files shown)


kirin/core/model.py
```diff
@@ -320,15 +321,18 @@ def __init__(self, raw_data, connector, contributor, status=None, error=None, re
...
-    def get_last_update_by_contributor(cls):
+    def get_last_update_by_contributor(cls, validity):
         from kirin import app
         result = {}
         contributor = [app.config['CONTRIBUTOR'], app.config['GTFS_RT_CONTRIBUTOR']]
         for c in contributor:
-            row = db.session.query(db.func.max(cls.created_at)) \
-                              .filter(cls.contributor == c).one()
+            sql = db.session.query(db.func.max(cls.created_at)).filter(cls.contributor == c)
+            if validity:
```
xlqian: from what i understood, you want the all results without being filtered if validity is not specified(None). if validity is given, you want results get filtered by the value of validity(True/False or an Enum), right?

`if validity is not None:`

kirin/core/model.py
```diff
@@ -320,15 +321,18 @@ def __init__(self, raw_data, connector, contributor, status=None, error=None, re
         self.error = error
         self.contributor = contributor
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
+        self.validity = True
 
     @classmethod
-    def get_last_update_by_contributor(cls):
+    def get_last_update_by_contributor(cls, validity):
```
xlqian: use "validity=None" to avoid ambiguity

kirin/gtfs_rt/model_maker.py
```diff
@@ -105,6 +105,11 @@ def build(self, rt_update, data):
                 continue
             tu = self._make_trip_updates(entity.trip_update, data_time=data_time)
             trip_updates.extend(tu)
+
+        if len(trip_updates) == 0:
```
xlqian: `if trip_updates:`

migrations/versions/3a12525235bf_add_validity.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+# revision identifiers, used by Alembic.
+revision = '3a12525235bf'
+down_revision = '2763a0e3f0bf'
+
+from alembic import op
+import sqlalchemy as sa
+
+rt_validity = sa.Enum('valid', 'empty', 'invalid', name='rt_validity')
```
pbougue: `empty` means `None` or means that the feed is "OK-but-empty"? Probably the second, if it's clear for anyone else that's fine, otherwise `valid_empty` seems better to me.

migrations/versions/3a12525235bf_add_validity.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def upgrade():
+    rt_validity.create(op.get_bind())
+    op.add_column('real_time_update', sa.Column('validity', rt_validity, nullable=True))
+    op.execute("UPDATE real_time_update SET validity = 'valid' WHERE validity IS NULL")
+    op.alter_column('real_time_update', 'validity', nullable=False)
+    op.create_index('ix_rt_validity', 'real_time_update', ['validity'], unique=False)
```
pbougue: usually (from the other migrate files) it would more be `rt_validity_idx`

kirin/core/model.py
```diff
@@ -320,15 +321,18 @@ def __init__(self, raw_data, connector, contributor, status=None, error=None, re
         self.error = error
         self.contributor = contributor
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
+        self.validity = 'valid'
 
     @classmethod
-    def get_last_update_by_contributor(cls):
+    def get_last_update_by_contributor(cls, is_valid=False):
```
pbougue: minor: I'd name that param `only_valid`

kirin/core/model.py
```diff
@@ -320,15 +321,18 @@ def __init__(self, raw_data, connector, contributor, status=None, error=None, re
         self.error = error
         self.contributor = contributor
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
+        self.validity = 'valid'
```
pbougue: So if we can't decode, will we keep that 'valid' value?

kirin/gtfs_rt/model_maker.py
```diff
@@ -100,11 +100,22 @@ def build(self, rt_update, data):
                        .format(data.header.timestamp, data_time))
 
         trip_updates = []
+
+        if len(data.entity) == 0:
+            rt_update.validity = 'empty'
```
pbougue: to me the 2 cases do not deserve different behavior (all are decodable but empty, no?)

kirin/core/model.py
kadhikari: If there is any exception, validity becomes 'invalide'...
Codes to be modified

kirin/gtfs_rt/model_maker.py
kadhikari: Empty: when the flux gtfs-rt has header only but no entity.
Invalid: For entity.trip_update of the flux gtfs-rt if there is not any information to be sent to navitia

migrations/versions/3a12525235bf_add_validity.py
kadhikari: ok

kirin/core/model.py
```diff
@@ -305,33 +305,31 @@ class RealTimeUpdate(db.Model, TimestampMixin):
...
-    validity = db.Column(db.Enum('valid', 'empty', 'invalid', name='rt_validity'), nullable=True)
 
     trip_updates = db.relationship("TripUpdate", secondary=associate_realtimeupdate_tripupdate, cascade='all',
                                    lazy='select', backref=backref('real_time_updates', cascade='all'))
 
     __table_args__ = (db.Index('realtime_update_created_at', 'created_at'),
                       db.Index('realtime_update_contributor_and_created_at', 'created_at', 'contributor'))
 
-    def __init__(self, raw_data, connector, contributor, status=None, error=None, received_at=None):
+    def __init__(self, raw_data, connector, contributor, status='OK', error=None, received_at=None):
```
pbougue: Not sure about this "OK" init, we have to test what happens if it's not decodable

kirin/gtfs_rt/model_maker.py
kadhikari: It's a list and we check if it contains any element.

kirin/core/model.py
kadhikari: If the proto is not decodable, there is an uncatched exception hence flux gtfs-rt is not saved.
Once decodable, all the exceptions are catched and gtfs-rt saved as 'KO' with an error message.

Note: DecodeException is implemented in api post but not in poller(task).

kirin/utils.py
```diff
@@ -126,6 +126,15 @@ def make_kirin_lock_name(*args):
     return '|'.join([app.config['TASK_LOCK_PREFIX']] + [str(a) for a in args])
 
 
+def save_gtfs_rt_with_error(data, connector, contributor, status='OK', error=None):
```
pbougue: Here, status is probably more 'KO' (we can as well force caller to manage it)

kadhikari:
J'ai mis à jour le ticket: https://jira.canaltp.fr/browse/NAVP-723

kadhikari:
jenkins test this please

xlqian:
Jenkins, test this please!

kadhikari:
jenkins test this please

kirin/core/model.py
```diff
@@ -327,12 +329,17 @@ def get_last_update_by_contributor(cls, only_valid=False):
         result = {}
         contributor = [app.config['CONTRIBUTOR'], app.config['GTFS_RT_CONTRIBUTOR']]
         for c in contributor:
-            sql = db.session.query(db.func.max(cls.created_at)).filter(cls.contributor == c)
+            sql = db.session.query(cls.created_at, cls.error).filter(cls.contributor == c)
```
pbougue: So this shouldn't affect status response time?
We should probably keep that max, as we only use the first row (fonly filter changes depending on the KO/OK)

kirin/utils.py
```diff
@@ -135,6 +135,25 @@ def save_gtfs_rt_with_error(data, connector, contributor, status, error=None):
...
+def manage_gtfs_rt_with_error(data, connector, contributor, status, error=None):
+    """
+    If error='Http Error' and last gtfs rt with the same error is inserted before 5 second then
+    insert in the table real_time_update with error message (5 seconds is fixed value)
+
+    For other error messages, insert gtfs-rt in the table real_time_update with status and error.
+
+    parameters: data, connector, contributor, status, error
+    """
+    if error == 'Http Error':
```
pbougue: This mechanism should not be HttpError-specific

kirin/utils.py
```diff
@@ -135,6 +135,25 @@ def save_gtfs_rt_with_error(data, connector, contributor, status, error=None):
...
+        rt_update_error = model.RealTimeUpdate.get_last_error(connector, status, error)
```
pbougue: It would be better to get the last error (no matter which type it is), then just update created_at value if error-type is the same, and create a new one otherwise.
This way it is "OK" but we can still improve a bit by retrieving only the last RTUpdate and test for date and possible error-type (db request would probably be lighter and result would be more exact if one poll works during those 5 seconds).

tests/integration/gtfs_rt_test.py
```diff
@@ -1369,45 +1369,64 @@ def test_save_gtfs_rt_with_error():
...
-        assert RealTimeUpdate.query.first().raw_data == ''
+        assert RealTimeUpdate.query.first().raw_data == 'toto'
         assert RealTimeUpdate.query.first().status == 'KO'
         assert RealTimeUpdate.query.first().error == 'Http Error'
         assert RealTimeUpdate.query.first().created_at == created_at
+        assert RealTimeUpdate.query.first().updated_at > updated_at
+
+        updated_at = RealTimeUpdate.query.first().updated_at
+
+        time.sleep(6)
```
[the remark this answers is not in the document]
pbougue: Hmm, not absolutly, but it insists on no time-dependant behavior... I'm OK to withdraw it.

pbougue:
One problem we can anticipate with that is that we get a KO probe when it's only partial KO (like Navitia not finding all vj)... But it's probably OK for a probe

kirin/utils.py
```diff
@@ -126,6 +126,48 @@ def make_kirin_lock_name(*args):
...
+    model.db.session.commit()
+
+
+def poke_updated_at(rtu):
+    """
+    just update the updated_at of the RealTimeUpdate object provided
+    """
+    if rtu:
+        status = rtu.status
+        rtu.status = 'pending' if status != 'pending' else 'OK' # just to poke updated_at
```
xlqian: can't we update `updated_at` directly? instead of calling commit twice?

kirin/gtfs_rt/gtfs_rt.py
```diff
@@ -69,6 +71,8 @@ def post(self):
         try:
             proto.ParseFromString(raw_proto)
         except DecodeError:
+            #We save the non-decodable flux gtfs-rt
+            manage_db_error(proto, 'gtfs-rt', contributor=self.contributor, status='KO', error='Decode Error')
```
xlqian: are you sure that you want to persist `proto` not the `raw_proto`(i know that there'll be some complications if we do so)? if `DecodeError` occurs, `proto` may be empty, no?

pbougue: It's the same problem we had before : the raw_proto is clearly not a string, and is rejected by sqlalchemy.
We'd have to migrate the field to a binary type

PR #139, opened 2018-02-02
kirin: fix for migration with lots data and live insert

kadhikari:
During Migraion:

- Script updates the tables updating status = 'OK' where it was NULL

- The poller inserts somes lines in the table real_time_update with status = NULL

- At last when the script alters the table with status not NULL, because of some lines inserted by poller
there is an erreur "column "status" contains null values"

Solution: 

- Alter the column status with server_default='OK' in this version.

- Alter the column status with nullable=False in the next migration.


kadhikari:
jenkins test this please

xlqian:
Jenkins, test this please!

PR #141, opened 2018-03-07
kirin: Update realTimeUpdate with status KO for repeated gtfs_rt

kadhikari:
Ticket: https://jira.canaltp.fr/browse/NAVITIAII-2395


kadhikari:
@pbougue i added a test end point (/status) with error.

PR #142, opened 2018-08-17
Doc: Spec of the core model of Kirin

prhod:


prhod:
thanks for the revue. can you check again the part on VJ-timestamp plz ?

PR #152, opened 2018-08-24
Doc: Kirin purpose + detailed setup + API

mbounabi:

PR #153, opened 2018-08-24
Create endpoint for /cots

### LINKED ISSUE #143: Add a new endpoint for COTS
Add a new endpoint `/cots`
This endpoint behaves like the `/ire` endpoint, except the content will be JSON.
The content is saved in the database "as is" in the table `RealTimeUpdate`

pbougue:
fix #143 

only storing raw_data

TODO:

- [x] hand-test that it's working
- [x] small dev documentation once #152 is merged
- [x] add tests


kirin/cots/cots.py
```diff
@@ -0,0 +1,93 @@
...
+#from model_maker import KirinModelBuilder #TODO create and use model_maker.py
+from datetime import datetime
+import logging
+
+
+def get_cots(req):
+    """
+    get COTS stream, for the moment, it's the raw json
+    """
+    if not req.data:
```
kinnou02: you have http://flask.pocoo.org/docs/1.0/api/#flask.Request.get_json that might be better

pbougue: The little gain is we would tell right away about json parse errors.
But we have to store it as a text (serialize?), and in the current system we give a RTU to the model_maker that it fills with corresponding TripUpdate (json parsing again if we don't adapt it).
Those 2 things tell me not to change that.

kirin/cots/cots.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
```
xlqian: 2018 :D

pbougue: OK, if it's only that, I will change it on the next PR, if it's fine for you ;)

kirin/cots/cots.py
```diff
@@ -0,0 +1,93 @@
...
+            record_call('failure', reason=str(e), contributor=self.contributor)
+            raise
+
+        _, log_dict = core.handle(rt_update, trip_updates, current_app.config['COTS_CONTRIBUTOR'])
+        duration = (datetime.utcnow() - start_datetime).total_seconds()
+        log_dict.update({'duration': duration})
+        record_call('Simple feed publication', **log_dict)
+        logging.getLogger(__name__).info('Simple feed publication', extra=log_dict)
+
+        return 'OK', 200
```
mbounabi: Nothing factorised with IRE because IRE will be deprecated (sometime soon?) ?

PR #155, opened 2018-08-28
Doc: initiate cots connector

papailio:


documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
+# COTS Connector
+
+## Overview
+Realtime information for long distance trains of the SNCF network is received in a COTS stream. This document describes how a COTS realtime stream is modeled in Kirin.
+
+## Input data description
+A realtime COTS stream [link to provide] is obtained as a JSON file via a message queue mechanism. Each feed message represents a `notification` object that contains the information about the train (its status, the associated delay, etc.).
+
+The information concerning the displayed messages related to train modifications is referenced in a separate stream. This stream is provided as another JSON file by an external web service called *paramatreLIV* [link to provide]. Each such message representes a `Motif` object.
```
prhod: I'm not sure there is a need to specify a `Motif` object. Either discribe the contained properties (redundant with the link) or make a light description ? 
`The web service returns a text message for all available situations associated with an id referenced in COTS stream.` ?

documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
+# COTS Connector
+
+## Overview
+Realtime information for long distance trains of the SNCF network is received in a COTS stream. This document describes how a COTS realtime stream is modeled in Kirin.
+
+## Input data description
+A realtime COTS stream [link to provide] is obtained as a JSON file via a message queue mechanism. Each feed message represents a `notification` object that contains the information about the train (its status, the associated delay, etc.).
```
prhod: I don't think the `notification object` is a good idea as it's not described (and not in the Kirin model). What about a simple : 
`Each feed message represents an update on the information about a train (its status, the associated delay, causes, etc.).`

documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
...
+
+The information concerning the displayed messages related to train modifications is referenced in a separate stream. This stream is provided as another JSON file by an external web service called *paramatreLIV* [link to provide]. Each such message representes a `Motif` object.
+<br/>-- Validation of parametreLIV ??
+
+## Connector description
```
prhod: What about adding an introduction stating something like : 
`This document doesn't describe all the fields of the Kirin model. Only COTS relevant fields are described bellow. For exemple, the RealTimeUpdate.id field is managed by Kirin and is not detailed in this specs.` ?
This way, you won't need to specify internal fields like `id` or `trip_updates`.

documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
...
+### RealTimeUpdate
+
+Kirin object | COTS object/value | Description
+--- | --- | ---
+id |  | Unique id of the received stream
+received_at | `YYYY-MM-DD` | System date and time at data reception (UTC)
+connector | `cots` | Name of the realtime data source
```
prhod: I think there is no need to add the field description (redundant with model spec), only a special rule when simple mapping is not sufficiant. 
On this field, should we emphasis the `cots` is a fixed value for this data stream ?

documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
...
+error | ??? | Description of the error
+raw_data | `notification` | Content of the received raw data
+contributor | `COTS_CONTRIBUTOR` | Kirin contributor
+trip_updates | [`TripUpdates`] | A list of `TripUpdates`
+
+### TripUpdate
+Kirin object | COTS object/value | Description
+--- | --- | ---
+id |  | Unique id of the `TripUpdate`
+vj_id | *VehicleJourney/id* | Id of the `VehicleJourney` found in Navitia
```
prhod: we need to describe how this is done

documentation/cots_connector.md
```diff
@@ -0,0 +1,35 @@
...
+status | If *nouvelleVersion/statutOperationnel* = "AJOUTEE", then status = `add` <br/>If *nouvelleVersion/statutOperationnel* = "SUPPRIMEE", then status = `delete` <br/>If *nouvelleVersion/statutOperationnel* = "PERTURBEE", then status = `update` <br/>Otherwise, status = `update` | Modification type of the trip
```
prhod: I think you should have: 
- `nouvelleVersion/statutOperationnel` in the `COTS object/value` field
- the complete rule in the `description` field.
Personally, I try  to avoid If/Then in specs to improve readability.

documentation/cots_connector.md
```diff
@@ -0,0 +1,31 @@
...
+--- | --- | ---
+connector |  | Fixed value `cots`
+raw_data | _Complete received feed_ | 
+contributor |  | Fixed value specified in the configuration of Kirin
+trip_updates |  | List of trip updates information, see `TripUpdates` below
+
+### TripUpdate
+Kirin property | COTS object/value | Comment/Mapping rule
+--- | --- | ---
+vj_id | | Refers to the trip found in Navitia that the update applies to, see `VehicleJourney` below
```
prhod: What do you think of : 
Id of the `VehicleJourney` in Navitia updated by this `TripUpdate`. One COTS feed can update more than one `VehicleJourney`, see below for mapping method.

documentation/cots_connector.md
```diff
@@ -0,0 +1,31 @@
...
+status | *nouvelleVersion/statutOperationnel* | If *nouvelleVersion/statutOperationnel* = "AJOUTEE", then status = `add` <br/>If *nouvelleVersion/statutOperationnel* = "SUPPRIMEE", then status = `delete` <br/>If *nouvelleVersion/statutOperationnel* = "PERTURBEE", then status = `update` <br/>Otherwise, status = `update`
```
prhod: can you avoid using if/then/else ? (if not, it's no big deal)
Maybe having one group for "PERTURBEE" and Otherwise ?

documentation/cots_connector.md
```diff
@@ -0,0 +1,31 @@
...
+message | *nouvelleVersion/idMotifInterneReference* | The label of the message referenced by the value of *nouvelleVersion/idMotifInterneReference*
```
prhod: Where to find the label ?

papailio:
Concerning the if/then part, i tried but i cannot see any other way to present such a conditional case..

PR #165, opened 2018-09-04
Doc:  Add info about GTFS-RT poller task 

mbounabi:
@pbougue : the info we discussed about the poller is now in the doc


mbounabi:
@pbougue agreed!

PR #183, opened 2018-09-13
Purge cots

### LINKED ISSUE #149: Managing a purge on old datasets

pbougue:
Fix #149 

TODO ('do_not_merge' is here for that):
- [x] Hand test locally
- [x] merge #175

### DIFF
--- a/kirin/default_settings.py
+++ b/kirin/default_settings.py
@@ -191,5 +191,15 @@ CELERYBEAT_SCHEDULE = {
         'task': 'kirin.tasks.purge_ire_rt_update',
         'schedule': schedules.crontab(hour='3', minute='45'),
         'options': {'expires': 3600}
+    },
+    'purge_cots_trip_update': {
+        'task': 'kirin.tasks.purge_cots_trip_update',
+        'schedule': schedules.crontab(hour='4', minute='0'),
+        'options': {'expires': 3600}
+    },
+    'purge_cots_rt_update': {
+        'task': 'kirin.tasks.purge_cots_rt_update',
+        'schedule': schedules.crontab(hour='4', minute='15'),
+        'options': {'expires': 3600}
     }
 }
--- a/kirin/tasks.py
+++ b/kirin/tasks.py
@@ -162,3 +162,24 @@ def purge_ire_rt_update(self):
     config = {'nb_days_to_keep': 100,
               'connector': 'ire'}
     purge_rt_update.delay(config)
+
+
+@celery.task(bind=True)
+def purge_cots_trip_update(self):
+    """
+    This task will remove ONLY TripUpdate, StopTimeUpdate and VehicleJourney that are created by COTS but the
+    RealTimeUpdate are kept so that we can replay it for debug purpose. RealTimeUpdate will be remove by another task
+    """
+    config = {'contributor': app.config.get('COTS_CONTRIBUTOR'),
+              'nb_days_to_keep': 10}
+    purge_trip_update.delay(config)
+
+
+@celery.task(bind=True)
+def purge_cots_rt_update(self):
+    """
+    This task will remove realtime update
+    """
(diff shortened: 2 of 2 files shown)


pbougue:
Jenkins, test this please

mbounabi:
I know it is inherited from IRE, but why are there are two  separate functions for purging trip_update and rt_update? In which case is it useful to keep one but not the other?

kirin/tasks.py
```diff
@@ -162,3 +162,24 @@ def purge_ire_rt_update(self):
...
+
+
+@celery.task(bind=True)
+def purge_cots_trip_update(self):
+    """
+    This task will remove ONLY TripUpdate, StopTimeUpdate and VehicleJourney that are created by COTS but the
+    RealTimeUpdate are kept so that we can replay it for debug purpose. RealTimeUpdate will be remove by another task
+    """
+    config = {'contributor': app.config.get('COTS_CONTRIBUTOR'),
+              'nb_days_to_keep': 10}
```
kadhikari: minor: why not configurable as for  purge_gtfs_trip_update ?

kirin/tasks.py
```diff
@@ -162,3 +162,24 @@ def purge_ire_rt_update(self):
...
+    purge_trip_update.delay(config)
+
+
+@celery.task(bind=True)
+def purge_cots_rt_update(self):
+    """
+    This task will remove realtime update
+    """
+    config = {'nb_days_to_keep': 100,
```
kadhikari: minor: same

kirin/default_settings.py
```diff
@@ -191,5 +191,15 @@
         'task': 'kirin.tasks.purge_ire_rt_update',
         'schedule': schedules.crontab(hour='3', minute='45'),
         'options': {'expires': 3600}
+    },
+    'purge_cots_trip_update': {
+        'task': 'kirin.tasks.purge_cots_trip_update',
+        'schedule': schedules.crontab(hour='4', minute='0'),
```
nonifier: Question curiosité: pourquoi on utilise des tâches cron, plutôt que faire dépiler les tâches au fur et à mesure ? Si la tâche `purge_cots_trip_update` prend plus de 15min, ça pause pas de problème ?

kirin/tasks.py
pbougue: Would be better yes, but we probably won't have the same tweak need for SNCF as there is only one instance, quite stable.
So we can let it like this and change it if the need arises (or when we refactor the whole config mechanism)

kirin/default_settings.py
pbougue: > Why cron task instead of little purge tasks?
I'd say (@kinnou02 can corect me) that we don't want to have the task queue (and logs) full of those, as it's not the main task of kirin at all. Also it's easy to configure as a cron and it's supposed to be quite lightweight task... But you're right to challenge it because I think the main reason is that it was a real crontask on platform's side before and we just migrated it inside (you know : "it's historical" but still, we improved it a bit).
So it can be discussed and changed in a future PR or in future projects.
> What if it's longer than 15min?
We should be quite safe on the db-consistency's side as there are 90days of diff between the 2 purges. 
For the rest of it, I guess it would depend on the number of workers. With only one worker, as the following task expires after 1h, if this one is 1h15min long, we won't launch the next one... Which could end being a problem on the long term. But 1h15min is safe as the last IRE rt_upadte purge was <7s, last IRE trip_update purge was 2min30s and we expect the same amount of objects.

Those responses are a bit opiniated, feel free to challenge them!

pbougue: @mbounabi I must recede on one point in my previous reply : the objects for trips are not larger than rt_upate, considering the duration of the purge. Even if we purge from 100days ago for RTU (strikes?) and from 10 days ago for Trips.
Still, the rest holds.

PR #188, opened 2018-09-20
Minor fixes

### LINKED ISSUE #144: Check /status on cots data
check the`/status` api is giving back infos on cots feeds and add infos if necessary on Dev.

pbougue:
Fix #144 

* error message correction
* minor refactor

Short, but easier to read by commit

### DIFF
--- a/kirin/abstract_sncf_model_maker.py
+++ b/kirin/abstract_sncf_model_maker.py
@@ -141,6 +141,6 @@ class AbstractSNCFKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
                 vjs[nav_vj['id']] = vj
 
         if not vjs:
-            raise ObjectNotFound('no train found for headsigns {}'.format(headsigns))
+            raise ObjectNotFound('no train found for headsign(s) {}'.format(headsign_str))
 
         return vjs.values()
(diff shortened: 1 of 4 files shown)


kirin/abstract_sncf_model_maker.py
```diff
@@ -31,6 +31,9 @@
 from __future__ import absolute_import, print_function, division
 import logging
 from datetime import timedelta
+
+import jmespath
```
benoit-bst: :+1:

kirin/abstract_sncf_model_maker.py
```diff
@@ -76,14 +79,12 @@ def headsigns(str):
...
-            if c['value'] == nav_external_code and c['type'] == 'CR-CI-CH':
-                nav_stop_times.append(s)
-                break
+    nav_stop_times = jmespath.search(
+        'stop_times[? stop_point.stop_area.codes[? value==`{nav_ext_code}` && type==`CR-CI-CH`]]'.format(
+            nav_ext_code=nav_external_code),
+        nav_vj)
 
+    log_dict = None
     if not nav_stop_times:
```
benoit-bst: I don't know jmespath, but if you are sure that it can return a `none` when is empty, then it's okay for me.
I hope there is a test, otherwise the `codecov` turns into demon :smiling_imp:

pbougue: Ahah, there's no codecov or black in kirin so far :see_no_evil:

But if I replace that search line with `nav_stop_times = None` tests are crashing (although I don't think the 'missing stop point' part is tested, but...)

And yes, it returns none if it does not match : https://jmespath.readthedocs.io/en/stable/specification.html

PR #189, opened 2018-09-27
Cause message handler

### LINKED ISSUE #154: Retrieve "Motif" from external service
Motifs (causes) are gonna be provided by an external service that we have to manage.

pbougue:
Fix #154 
Handles and test cause messages for COTS feed.

- [x] Hand-tested on SNCF's test API (regular case)
- [x] Hand-test expired-token case (401 response code from resource service)
- [x] If possible merge CanalTP/fab_kirin#31 and https://github.com/CanalTP/kirin_deployment_conf/pull/23 before

:warning: WARNING :warning: tests now uses `test_settings.py`

:mag_right: please read by commit


pbougue:
Jenkins, test this please

pbougue:
Jenkins, test this please

kirin/cots/model_maker.py
```diff
@@ -228,6 +240,11 @@ def _make_trip_update(self, vj, json_train):
             nav_stop = nav_st.get('stop_point', {})
             st_update = model.StopTimeUpdate(nav_stop)
             trip_update.stop_time_updates.append(st_update)
+            st_message_id = get_value(pdp, 'idMotifInterneDepartReference', nullable=True)
```
mbounabi: There can be a message_id in 'Depart' and 'Arrivee' but if there's one in Depart, the one in 'Arrivee' doesn't matter?
Is that the expected behavior?

kirin/cots/message_handler.py
```diff
@@ -0,0 +1,167 @@
...
+            reset_timeout=current_app.config['COTS_PAR_IV_CIRCUIT_BREAKER_TIMEOUT_S']
+        )
+
+    def __repr__(self):
+        """
+        Allow this class to be cacheable
+        """
+        return '{}.{}.{}.{}'.format(self.__class__, self.resource_server, self.token_server, self.client_id)
+
+    def _service_caller(self, method, url, headers, data=None, params=None):
```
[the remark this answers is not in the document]
pbougue: I hesitated to let this or not... I have no strong opinion to keep it, so I may remove it if you want.
It's unused for now, params would passed in the URL by requests (the `?k1=v1&k2=v2` part at the end of it).
Could be used later to filter calls (validity period, attributes to return as it seems to be possible from latest ParIV doc we got).

kirin/cots/message_handler.py
```diff
@@ -0,0 +1,167 @@
...
+            return None
+        content = response.json()
+        access_token = content.get('access_token')
+        if not access_token:
+            logging.getLogger(__name__).error('COTS message sub-service, no access_token in response')
+            return None
+        return access_token
+
+    def _call_webservice(self):
+        access_token = self._get_access_token()
```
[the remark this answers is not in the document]
pbougue: You're right for an exception (raised by _service_caller) I will change it.
As for the `token is None` part, I don't know it may be on purpose that there is no token (test service)... I can change it also (current tests use a token, so they shouldn't break).

kirin/cots/model_maker.py
pbougue: Yep, for now we only have one message at a stop_time (so the same for arrival and departure).
https://github.com/CanalTP/kirin/blob/master/documentation/cots_connector.md#stoptimeupdate (message row) specify the behavior. I can add a comment refering to the doc (or just explaining what's done if you prefer).

kirin/test_settings.py
```diff
@@ -0,0 +1,10 @@
+# encoding: utf-8
+
+from __future__ import absolute_import
+
+
+COTS_PAR_IV_API_KEY = 'tchoutchou_api_key'
+COTS_PAR_IV_MOTIF_RESOURCE_SERVER = 'https://messages.service/resource'
+COTS_PAR_IV_TOKEN_SERVER = 'https://messages.service/token'
+COTS_PAR_IV_CLIENT_ID = 'tchoutchou_id'
```
xlqian: :steam_locomotive: :laughing:

tests/readme.md
```diff
@@ -2,7 +2,7 @@
 
 To run the tests, just run py.test in the kirin root dir, using appropriate python virtualenv:
 ```bash
-py.test --doctest-modules
+KIRIN_CONFIG_FILE=test_settings.py py.test --doctest-modules
```
xlqian: I just checkouted your PR, i didn't need to use KIRIN_CONFIG_FILE to have all tests passed...

kirin/cots/model_maker.py
```diff
@@ -228,6 +240,12 @@ def _make_trip_update(self, vj, json_train):
             nav_stop = nav_st.get('stop_point', {})
             st_update = model.StopTimeUpdate(nav_stop)
             trip_update.stop_time_updates.append(st_update)
+            # using the message from departure-time in priority, if absent fallback on arrival-time's message
+            st_message_id = get_value(pdp, 'idMotifInterneDepartReference', nullable=True)
+            if not st_message_id:
+                st_message_id = get_value(pdp, 'idMotifInterneArriveeReference', nullable=True)
+            if st_message_id:
```
xlqian: else?

kirin/core/handler.py
```diff
@@ -251,7 +261,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update):
     res_stoptime_updates = []
 
     res.status = new_trip_update.status
-    if new_trip_update.message:
+    if new_trip_update.message is not None or is_new_complete:
```
xlqian: the logic is not quite trivial for me :/, but ok

tests/readme.md
pbougue: As discussed it's probably just your :eyes: , it should break without test_settings :stuck_out_tongue_closed_eyes:

PR #192, opened 2018-10-08
Cleanup

pbougue:
Lot of meaningless cleanup (headers)

Also removing unused `testing_settings.py` file, as it was only mentionned commented in CI: https://ci.navitia.io/view/kirin/job/kirin_master/configure (@kinnou02 if you have some memory of it, but it looks like old tests that were plugged on dev)


pbougue:
Jenkins, test this please

Issue #182, opened 2018-09-12
COTS: check the back-to-normal behavior


pbougue:
We have to check that the behavior of Kirin+Navitia is OK when we get a disruption, then a new one indicating everything is fine again (we might miss some informations if we don't get a pronosticIV=0 in feed).

In the following examples, first stops are removed.
* Final stops are late by 5min [Application du retard.json.txt](https://github.com/CanalTP/kirin/files/2374918/Application.du.retard.json.txt)
* Final stops are back-to-normal [Annulation en OPE des suppressions des desserte supprimées par PTP .json.txt](https://github.com/CanalTP/kirin/files/2374921/Annulation.en.OPE.des.suppressions.des.desserte.supprimees.par.PTP.json.txt)

PR #194, opened 2018-10-11
Add test for back to normal

pbougue:
Fix https://github.com/CanalTP/kirin/issues/182
(mostly inspired from provided feeds)

* (partial removal + delay) then (one stop removed and delayed stops are OK)
* test also behavior when missing some info on back-to-normal feed (on delayed stops)
* test that causes disappear correctly

- [x] hand-tested with navitia plugged on Artemis' instance

### DIFF
--- a/tests/integration/utils_sncf_test.py
+++ b/tests/integration/utils_sncf_test.py
@@ -98,6 +98,246 @@ def check_db_96231_delayed(contributor=None, motif_externe_is_null=False):
         assert db_trip_delayed.contributor == contributor
 
 
+def check_db_870154_partial_removal(contributor=None):
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) >= 1
+        assert len(TripUpdate.query.all()) == 1
+        assert len(StopTimeUpdate.query.all()) == 12
+        db_trip = TripUpdate.find_by_dated_vj('OCE:SN870154F01001', datetime(2018, 11, 2, 9, 54, tzinfo=utc))
+        assert db_trip
+
+        assert db_trip.vj.navitia_trip_id == 'OCE:SN870154F01001'
+        assert db_trip.vj.get_start_timestamp() == datetime(2018, 11, 2, 9, 54, tzinfo=utc)
+        assert db_trip.vj_id == db_trip.vj.id
+        assert db_trip.status == 'update'
+        # 12 stop times must have been created
+        assert len(db_trip.stop_time_updates) == 12
+
+        # the first 5 stops are removed (Rodez to Viviez-Decazeville, also arrival in Capdenac)
+        first_st = db_trip.stop_time_updates[0]
+        assert first_st.stop_id == 'stop_point:OCE:SP:TrainTER-87613422'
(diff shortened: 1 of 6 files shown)


tests/integration/utils_sncf_test.py
```diff
@@ -98,6 +98,246 @@ def check_db_96231_delayed(contributor=None, motif_externe_is_null=False):
...
+        assert db_trip
+
+        assert db_trip.vj.navitia_trip_id == 'OCE:SN870154F01001'
+        assert db_trip.vj.get_start_timestamp() == datetime(2018, 11, 2, 9, 54, tzinfo=utc)
+        assert db_trip.vj_id == db_trip.vj.id
+        assert db_trip.status == 'update'
+        # 12 stop times must have been created
+        assert len(db_trip.stop_time_updates) == 12
+
+        # the first 5 stops are removed (Rodez to Viviez-Decazeville, also arrival in Capdenac)
```
mbounabi: the first 5 stops are removed but only 4 are checked.
Update during my review: the fifth one is checked in the function 'delay' but the comment is confusing.

tests/integration/utils_sncf_test.py
```diff
@@ -98,6 +98,246 @@ def check_db_96231_delayed(contributor=None, motif_externe_is_null=False):
...
+        assert len(RealTimeUpdate.query.all()) >= 1
+        assert len(TripUpdate.query.all()) == 1
+        assert len(StopTimeUpdate.query.all()) == 12
+        db_trip = TripUpdate.find_by_dated_vj('OCE:SN870154F01001', datetime(2018, 11, 2, 9, 54, tzinfo=utc))
+        assert db_trip
+
+        assert db_trip.vj.navitia_trip_id == 'OCE:SN870154F01001'
+        assert db_trip.vj.get_start_timestamp() == datetime(2018, 11, 2, 9, 54, tzinfo=utc)
+        assert db_trip.vj_id == db_trip.vj.id
+        assert db_trip.status == 'update'
```
mbounabi: Until this line, the code looks exactly the same as the function above. If it's the case, maybe it can be factorized.

tests/integration/utils_sncf_test.py
```diff
@@ -98,6 +98,246 @@ def check_db_96231_delayed(contributor=None, motif_externe_is_null=False):
...
+
+        twelfth_st = db_trip.stop_time_updates[11]
+        assert twelfth_st.stop_id == 'stop_point:OCE:SP:TrainTER-87594002'
+        assert twelfth_st.arrival_status == 'update'
+        assert twelfth_st.arrival == datetime(2018, 11, 2, 12, 25)
+        assert twelfth_st.arrival_delay == timedelta(minutes=10)
+        assert twelfth_st.departure_status == 'none'
+        assert twelfth_st.departure == datetime(2018, 11, 2, 12, 25)
+        assert twelfth_st.departure_delay == timedelta(minutes=10)
+        assert twelfth_st.message == u'Régulation du trafic'
```
mbounabi: The check of all these stops can surely be factorized in a way. TBC

PR #203, opened 2018-10-29
Doc COTS: stop_time creation

papailio:


pbougue:
After discussion, no dev to just stick to spec' is planned.
@papailio I added a do_not_merge so that you change what has to be (feel free to remove the label).

PR #204, opened 2018-11-05
Handle added stop_time

mbounabi:


tests/fixtures/cots_train_870154_partial_normal.json
```diff
@@ -593,6 +593,67 @@
 			"rang": 9,
 			"typeArret": "CH"
 		},
+		{
```
pbougue: That way, it's obvious what we wait from Kirin.
But this case of missing information could (and will) happen and we should check that we at least don't crash badly.

What we do exactly in that case is a gray area of spec', but I definitely think we should have test that it's not crashing (maybe not testing anything else, though).
I chose to keep the last information here (like what was done on IRE) so we kept the last RT-info we had on that stop instead of falling back to base-schedule.

We can however put that in another file if the tests are shared with IRE and separate those cases seems like a good idea. So my idea is to duplicate COTS file and write a new (probably simpler) test, almost the same COTS feeds chain (delay then normal), and at least test that we don't get a 500 error.

kirin/core/handler.py
```diff
@@ -127,6 +128,13 @@ def manage_consistency(trip_update):
     return True
 
 
+def find_st_in_vj(st_id, vj_sts):
+    for vj_st in vj_sts:
+        vj_st_id = vj_st.get('stop_point').get('id')
```
pbougue: Could end in a 500 if no 'stop_point' exists.

kirin/core/handler.py
```diff
@@ -276,8 +284,33 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
     last_departure = None
     local_circulation_date = new_trip_update.vj.get_local_circulation_date()
 
+    def get_next_stop():
+        if is_new_complete:
```
pbougue: Like you said, it does feel dirty.
I imagine more a system doing always the same : iterate smartly (hard part) on both trips, depending on the previous stop, and depending if it's an added stop...
Something like [`new_RT_st is that one being added`; `base_st is None, in between those two st`; `old_RT_st is that one as it was added already`]... But a lot of complicated things in that place I guess.

tests/fixtures/cots_train_870154_partial_normal.json
mbounabi: I don't see any grey area here. The COTS feed is supposed to be complete, so if 11 stops (PdP) are in it, there should be 11 stopTimeUpdates in the new TripUpdate. Not 12. 
@papailio can clarify.
Also, a COTS json isn't shared with IRE. I don't get your last point.

pbougue: True, COTS feed should be complete... But also any stop disappearing from base-schedule should be "SUPPRESSION" in COTS feed (that's part of COTS feed being complete). So if one is missing in the feed, we just have no news about it, hence the gray area. It's possible to consider it removed, but the truth is: we know nothing about it.

As for the shared part, I mentioned the fact that `assert`s (output testing) is shared with IRE most of the time as I remodeled IRE xml inputs in COTS json input for the biggest part to insure it works the same.

pbougue: Anyway (gray area or not) we don't have test on a stop being in base-schedule, but disappearing completely in COTS and I think we should have some.

kirin/cots/model_maker.py
```diff
@@ -337,17 +341,40 @@ def _make_trip_update(self, vj, json_train):
...
-        return get_navitia_stop_time_sncf(cr=get_value(pdp, 'cr'),
-                                          ci=get_value(pdp, 'ci'),
-                                          ch=get_value(pdp, 'ch'),
-                                          nav_vj=nav_vj)
+        nav_st, log_dict = get_navitia_stop_time_sncf(cr=get_value(pdp, 'cr'),
+                                                      ci=get_value(pdp, 'ci'),
+                                                      ch=get_value(pdp, 'ch'),
+                                                      nav_vj=nav_vj)
+        if not nav_st:
+            nav_stop, log_dict = self.request_navitia_stop_point(cr=get_value(pdp, 'cr'),
```
nonifier: Could we extract CR-CI-CH once as a whole code and use it across the differents functions ? That would make things so much simpler.

Something along these lines:

```python
cr_ci_ch = get_cr_ci_ch(pdp)
nav_st, log_dict = get_navitia_stop_time_sncf(cr_ci_ch, nav_vj=...)
if not nav_st:
    nav_stop, log_dict = self.request_navitia_stop_point(cr_ci_ch)
```

mbounabi: As the function is used also for IRE and GTFS-RT, it will have to be called the same amount of times (tried it... not really worth it)

kirin/core/handler.py
```diff
@@ -276,8 +284,33 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
...
+            # Iterate on the new trip update stop_times if they're all present in it
+            for order, st in enumerate(new_trip_update.stop_time_updates):
```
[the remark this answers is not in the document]
mbounabi: get_next_stop() at l.305 (not that long). The merge() function is quite long, but moving get_next_stop() outside will require passing several parameters in and out. 
This will change after the Kirin refacto, you can add every ideas you have about it in https://jira.kisio.org/browse/NAVP-1079

kirin/cots/model_maker.py
```diff
@@ -323,8 +324,11 @@ def _make_trip_update(self, vj, json_train):
 
                 elif cots_stop_time_status == 'CREATION':
                     # new stop_time added
-                    self._record_and_log(logger, 'nouvelleVersion/listePointDeParcours/statutCirculationOPE == '
-                                                 '"{}" is not handled (yet)'.format(cots_stop_time_status))
+                    setattr(st_update, _status_map[arrival_departure_toggle], 'add')
+                    cots_stop_time = get_value(cots_traveler_time,
+                                                      'dateHeure',
```
xlqian: strange indentation

kirin/cots/model_maker.py
```diff
@@ -337,17 +341,40 @@ def _make_trip_update(self, vj, json_train):
...
+    def request_navitia_stop_point(self, cr, ci, ch):
+        result = None
+        log_dict = None
+        external_code = '{}-{}-{}'.format(cr, ci, ch)
+        stop_points = self.navitia.stop_points(q={
+            'filter': 'stop_area.has_code("CR-CI-CH", "{}")'.format(external_code),
+            'count': '1'
+            })
+        if stop_points:
+            result = stop_points[0]
```
xlqian: return stop_points[0], log_dict

kirin/cots/model_maker.py
```diff
@@ -337,17 +341,40 @@ def _make_trip_update(self, vj, json_train):
...
+        else:
+            log_dict = {'log': 'No stop point found', 'stop_point_code': external_code}
```
xlqian: return None, {'log': 'No stop point found', 'stop_point_code': external_code}

kirin/core/handler.py
mbounabi: :heavy_check_mark:

kirin/core/handler.py
```diff
@@ -276,8 +285,34 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
...
+                # Find corresponding stop_time in the theoretical VJ
+                vj_st = find_st_in_vj(st.stop_id, new_trip_update.vj.navitia_vj.get('stop_times', []))
+
+                if vj_st is None and st.departure_status == 'add' or st.arrival_status == 'add':
+                    # It is an added stop_time, create a new stop time
+                    timezone = pytz.timezone(st.navitia_stop.get('stop_area').get('timezone'))
+                    added_st = {
+                        'stop_point': st.navitia_stop,
+                        'departure_time': pytz.utc.localize(parser.parse(st.departure).replace(tzinfo=None), is_dst=None).astimezone(timezone).time(),
```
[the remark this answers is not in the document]
mbounabi: localize() needs a naive datetime (with no tzinfo) or it fails. 
I think there might be a way to improve this part. Maybe in the future refacto...

kirin/cots/model_maker.py
mbounabi: indeed :heavy_check_mark:

kirin/cots/model_maker.py
mbounabi: :heavy_check_mark:

kirin/cots/model_maker.py
mbounabi: :heavy_check_mark:

kirin/core/handler.py
```diff
@@ -127,6 +129,34 @@ def manage_consistency(trip_update):
...
+
+
+def convert_to_local_time(timezone, utc_time):
+    """
+    Return local time according to UTC time and timezone
+    :param timezone: timezone info (type: datetime.tzinfo)
+    :param utc_time: UTC time (type: datetime.datetime)
+    :return: local time (type: datetime.time)
+
+    >>> utc_time = '20181108T093000+0000'
```
pbougue: Can you add a test :
```python
>>> utc_time = '20181108T093000+0100'
>>> timezone = pytz.timezone('Europe/Paris')
>>> convert_to_local_time(timezone, utc_time)
datetime.time(09, 30)
```
Not sure if it works (doubting about `replace(tzinfo=None)`).

kirin/core/handler.py
```diff
@@ -276,8 +309,34 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
...
+            for order, st in enumerate(new_trip_update.stop_time_updates):
+                # Find corresponding stop_time in the theoretical VJ
+                vj_st = find_st_in_vj(st.stop_id, new_trip_update.vj.navitia_vj.get('stop_times', []))
+
+                if vj_st is None and st.departure_status == 'add' or st.arrival_status == 'add':
+                    # It is an added stop_time, create a new stop time
+                    st_timezone = pytz.timezone(st.navitia_stop.get('stop_area').get('timezone'))
+                    added_st = {
+                        'stop_point': st.navitia_stop,
+                        'departure_time': convert_to_local_time(st_timezone, st.departure),
```
pbougue: Ahh, you have to do this... We definitely have to refactor to be full UTC in Kirin (and use UTC output now available in Navitia's /vehicle_journeys output) :smiley:

tests/integration/cots_test.py
```diff
@@ -519,3 +519,21 @@ def test_wrong_planned_stop_time_reference_post():
...
+    cots_add_file = get_fixture_data('cots_train_96231_add.json')
+    res = api_post('/cots', data=cots_add_file)
+    assert res == 'OK'
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 1
+        assert len(TripUpdate.query.all()) == 1
+        assert TripUpdate.query.all()[0].status == 'update'
+        assert len(StopTimeUpdate.query.all()) == 7
+        assert StopTimeUpdate.query.all()[3].arrival_status == 'add'
+        assert StopTimeUpdate.query.all()[3].arrival == datetime(2015, 9, 21, 16, 02)
```
pbougue: minor minor: remove trailing `0` prevents git from this red square.

kirin/core/handler.py
```diff
@@ -127,6 +129,34 @@ def manage_consistency(trip_update):
...
+
+    >>> utc_time = '20181108T093000+0000'
+    >>> timezone = pytz.timezone('Europe/Paris')
+    >>> convert_to_local_time(timezone, utc_time)
+    datetime.time(10, 30)
+    >>> timezone = pytz.timezone('Asia/Tokyo')
+    >>> convert_to_local_time(timezone, utc_time)
+    datetime.time(18, 30)
+    """
+    return pytz.utc.localize(parser.parse(utc_time).replace(tzinfo=None), is_dst=None).astimezone(timezone).time()
```
pbougue: Looks like if you just:
```python
return parser.parse(utc_time).astimezone(timezone).time()
```
It does the job and manages the test I proposed (so any localized datetime).

pbougue: We have no guaranty that COTS will contain only `+0000` datetimes, do we?

mbounabi: That's the thing, we don't care about +xxxx (see my other comment). It is replaced by None. What is considered is the tzinfo passe in parameter

mbounabi: Test it with a real case and suggest an implementation if you found one. Let's not just throw ideas out of nowhere in this thread.

tests/mock_navitia/vj_96231_early.py
```diff
@@ -0,0 +1,41 @@
...
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+import navitia_response
+from vj_96231 import VJ_96231_COMMON_RESPONSE
+
+response = navitia_response.NavitiaResponse()
+
+response.query = 'vehicle_journeys/?depth=2&since=20150921T152000&headsign=96231&show_codes=true&until=20150921T193900'
```
pbougue: I see 2 possible improvements here, by order of preference:
* ignore fully added stops when requesting navitia VJ here: https://github.com/CanalTP/kirin/blob/master/kirin/cots/model_maker.py#L204-L217. This would make the call to navitia more accurate and the need for this file would disappear (making the same call). This would also improve caching as the call wouldn't change depending on the RT feed (more important for GTFS-rt than COTS though).
* get inspiration and improve from what's done (poorly I admit) here : https://github.com/CanalTP/kirin/blob/master/tests/mock_navitia/__init__.py#L65 . This would allow grouping possible queries in the same file.

pbougue: Tracked in refactor ticket

PR #206, opened 2018-11-19
Add company in trip update

kadhikari:
Company_id added in trip_update:

- script of migration
- assign company_id in protobuf if present in trip_update
- some test on company_id

For détails see: https://jira.kisio.org/browse/NAVP-1032


benoit-bst:
Fasty Krishna :muscle: :running_man:

PR #210, opened 2018-11-27
kirin: manage effect in kirin

kadhikari:
This PR contains the following modifications:

- Add attribut effect of type enum in the table trip_update.

- Manage effect in kirin as done in kraken. In the next release, the function in kraken should de deleted.

- Existing tests modified and some new tests added.

- For detail please refer to: https://jira.kisio.org/browse/NAVP-1081

### DIFF
--- a/kirin/cots/model_maker.py
+++ b/kirin/cots/model_maker.py
@@ -301,7 +301,7 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
         trip_update.effect = Effect.MODIFIED_SERVICE.name
 
         # Initialize stop_time staus to nochange
-        st_status = 'nochange'
+        highest_st_status = 'nochange'
         pdps = _retrieve_interesting_pdp(get_value(json_train, 'listePointDeParcours'))
         # manage realtime information stop_time by stop_time
         for pdp in pdps:
@@ -393,10 +393,10 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
                     raise InvalidArguments('invalid value {} for field horaireVoyageur{}/statutCirculationOPE'.
                                            format(cots_stop_time_status, arrival_departure_toggle))
 
-                st_status = _get_higher_status(st_status, _arr_or_dep_status[arrival_departure_toggle])
+                highest_st_status = _get_higher_status(highest_st_status, _arr_or_dep_status[arrival_departure_toggle])
 
         # Calculates effect from stop_time status list (this work is also done in kraken and has to be deleted)
-        trip_update.effect = _get_effect_by_stop_time_status(st_status)
+        trip_update.effect = _get_effect_by_stop_time_status(highest_st_status)
         return trip_update
 
     def _get_navitia_stop_point(self, pdp, nav_vj):
--- a/kirin/cots/model_maker.py
+++ b/kirin/cots/model_maker.py
@@ -61,10 +61,6 @@ class Effect(Enum):
     STOP_MOVED = 8
 
 
-trip_effect = ('NO_SERVICE', 'REDUCED_SERVICE', 'SIGNIFICANT_DELAYS', 'DETOUR', 'ADDITIONAL_SERVICE',
-              'MODIFIED_SERVICE', 'OTHER_EFFECT', 'UNKNOWN_EFFECT', 'STOP_MOVED')
-
-
 status_order = {
     'nochange': 0,
     'add': 1,
@@ -287,14 +283,14 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
             # the whole trip is deleted
             trip_update.status = 'delete'
             trip_update.stop_time_updates = []
-            trip_update.effect = trip_effect[Effect.NO_SERVICE.value]
+            trip_update.effect = Effect.NO_SERVICE.name
             return trip_update
 
         elif trip_status == 'AJOUTEE':
             # the trip is created from scratch
             # not handled yet
             self._record_and_log(logger, 'nouvelleVersion/statutOperationnel == "AJOUTEE" is not handled (yet)')
-            trip_update.effect = trip_effect[Effect.ADDITIONAL_SERVICE.value]
+            trip_update.effect = Effect.ADDITIONAL_SERVICE.name
             return trip_update
 
         # all other status is considered an 'update' of the trip_update and effect is calculated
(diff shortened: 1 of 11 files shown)


kirin/cots/model_maker.py
```diff
@@ -164,6 +193,26 @@ def _retrieve_projected_time(source_ref, list_proj_time):
     return None
 
 
+def _get_higher_status(st1, st2):
+    idx1 = stop_time_status.keys().index(st1)
+    idx2 = stop_time_status.keys().index(st2)
+    return stop_time_status.keys()[max(idx1, idx2)]
```
xlqian: `max([st1, st2], key=lambda st: order[st])`, and you don't need `OrderedDict`

kirin/cots/model_maker.py
```diff
@@ -243,17 +288,25 @@ def _make_trip_update(self, vj, json_train):
...
-        # all other status is considered an 'update' of the trip
+        # all other status is considered an 'update' of the trip_update and effect is calculated
+        # from stop_time status list. This part is also done in kraken and is to be deleted later
+        # Ordered stop_time status= 'nochange', 'add', 'delete', 'update'
+        # 'nochange' or 'update' -> SIGNIFICANT_DELAYS, add -> MODIFIED_SERVICE, delete = DETOUR
         trip_update.status = 'update'
+        trip_update.effect = trip_effect.get('MODIFIED_SERVICE', None)
 
+        # Initialize stop_time staus to nochange
+        st_status = stop_time_status.get('nochange')
```
xlqian: `highest_st_status = 'nochange'`

kadhikari: Done

kirin/cots/model_maker.py
```diff
@@ -164,6 +186,24 @@ def _retrieve_projected_time(source_ref, list_proj_time):
...
+    return max([st1, st2], key=lambda st: status_order[st])
+
+
+def _get_effect_by_stop_time_status(status):
+    """
+    :param status: status value of stop_time
+    :return: the corresponding value for trip_update effect
+    """
+    status_to_effect = {
+        'nochange': 'SIGNIFICANT_DELAYS',
```
mbounabi: I don't really get this mapping...

kadhikari: When you calculate the higher stop_time status among all the status, we have to transform this status to trip_update effect.

kirin/cots/model_maker.py
```diff
@@ -243,17 +283,25 @@ def _make_trip_update(self, vj, json_train):
...
 
-        # all other status is considered an 'update' of the trip
+        # all other status is considered an 'update' of the trip_update and effect is calculated
+        # from stop_time status list. This part is also done in kraken and is to be deleted later
+        # Ordered stop_time status= 'nochange', 'add', 'delete', 'update'
+        # 'nochange' or 'update' -> SIGNIFICANT_DELAYS, add -> MODIFIED_SERVICE, delete = DETOUR
         trip_update.status = 'update'
+        trip_update.effect = Effect.MODIFIED_SERVICE.name
 
+        # Initialize stop_time staus to nochange
```
mbounabi: Minor: typo - sta**t**us

kirin/cots/model_maker.py
mbounabi: OK. But if there's no change in the COTS feed provided, why set the TripUpdate effect to 'SIGNIFICANT_DELAYS' ? 
If there truly are delays in the COTS feed, the trip_update status is set to 'update'

kirin/cots/model_maker.py
```diff
@@ -44,9 +44,31 @@
 from kirin.cots.message_handler import MessageHandler
 from kirin.exceptions import InvalidArguments
 from kirin.utils import record_internal_failure
+from enum import Enum
```
pbougue: (depends on comment below)
As enum is only available in python 3, you need to update requirements.txt with `enum34==<whatever>` (if you got inspiration from Jormun).

kirin/cots/model_maker.py
kadhikari: Do you have any suggestion?

kirin/cots/model_maker.py
```diff
@@ -44,9 +44,31 @@
...
 
 DEFAULT_COMPANY_ID = "1187"
 
+
+class Effect(Enum):
```
pbougue: Can't you use the one from model (so db.Enum)?
At least, this should be managed in GTFS-RT also, if we remove it from Kraken (depends how we manage the transition).
So this should be mutualized, not only in COTS.

kirin/cots/model_maker.py
```diff
@@ -310,17 +360,20 @@ def _make_trip_update(self, vj, json_train):
 
                     setattr(st_update, _status_map[arrival_departure_toggle], 'update')
                     setattr(st_update, _delay_map[arrival_departure_toggle], as_duration(cots_delay))
+                    _arr_or_dep_status[arrival_departure_toggle] = 'update'
```
pbougue: Looks like you don't need to store info nor toggle between arrival or departure here.
You can probably just do:
```python
highest_st_status = _get_higher_status(highest_st_status, 'update')
```
Same below.

kirin/cots/model_maker.py
```diff
@@ -44,9 +44,31 @@
...
+    SIGNIFICANT_DELAYS = 2
+    DETOUR = 3
+    ADDITIONAL_SERVICE = 4
+    MODIFIED_SERVICE = 5
+    OTHER_EFFECT = 6
+    UNKNOWN_EFFECT = 7
+    STOP_MOVED = 8
+
+
+status_order = {
```
pbougue: This matches the order from Kraken, so it's OK for GTFS-RT, but it looks like it's not what is requested in JIRA ticket.
@kinnou02 what's the highest priority is not clear in ticket, but I'd say `ADDITIONAL_SERVICE` is the most important one for COTS.
And we can probably have GTFS-RT order match COTS order as it makes more sense (at least to me) and as it's not a breaking change?

kirin/cots/model_maker.py
pbougue: As far as I can remember, we managed it that way in Kraken, because in GTFS-RT, the guideline is "if there is no significant effect, just don't send any RT info"... Which is missing some possible features like: "back-to-normal" or "everything is OK, I have RT info on that to reassure travelers".
We chose to keep the same enum and fallback on SIGNIFICANT_DELAYS for those reasons.
Not fan of that either, it's a functionnal (interface) debt we have, I'd say.

mbounabi: If not set, UNKNOWN_EFFECT will be set. To be discussed.

kirin/cots/model_maker.py
kadhikari: I tried declare en enum for Effect to be used in db as Enum and also in other modules but didn't work. 
i couldn't use db.Enum in model_maker and doesn't work neither.

kirin/cots/model_maker.py
kadhikari: ADDITIONAL_SERVICE is an effect only for added trip_update not for added stop_time so i don't understand the priority of ADDITIONAL_SERVICE

kirin/cots/model_maker.py
kadhikari: After consulting Alex, i set unchange to UNKNOWN_EFFECT

PR #211, opened 2018-11-27
reject cots if added stop_time is earlier than previous

xlqian:
https://jira.kisio.org/browse/NAVP-1096

When a new stop_point is added by COTS and it's stop_time(arrival/departure) is earlier than the previous departure, the cots  will be rejected. 

While without the fix, the VJ is considered as a pass midnight

At the same time, I did some cleaning on integration test's fixtures...

### DIFF
--- a/kirin/cots/model_maker.py
+++ b/kirin/cots/model_maker.py
@@ -312,7 +312,7 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
         highest_st_status = 'nochange'
         pdps = _retrieve_interesting_pdp(get_value(json_train, 'listePointDeParcours'))
 
-        # this dict is used to memoize the last stop_time in order to check the stop_time consistency
+        # this variable is used to memoize the last stop_time's departure in order to check the stop_time consistency
         # ex. stop_time[i].arrival/departure must be greater than stop_time[i-1].departure
         last_stop_time_depart = None
 
@@ -406,6 +406,7 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
 
                 elif cots_stop_time_status == 'DETOURNEMENT':
                     # new stop_time added also?
+                    # TODO: stop_time's consistency should also be tested once the feature is implemented
                     self._record_and_log(logger, 'nouvelleVersion/listePointDeParcours/statutCirculationOPE == '
                                                  '"{}" is not handled (yet)'.format(cots_stop_time_status))
 
--- a/kirin/cots/model_maker.py
+++ b/kirin/cots/model_maker.py
@@ -33,7 +33,7 @@ from __future__ import absolute_import, print_function, division
 
 import logging
 from datetime import datetime
-from dateutil import parser
+from dateutil import parser, tz
 from flask.globals import current_app
 from kirin.abstract_sncf_model_maker import AbstractSNCFKirinModelBuilder, get_navitia_stop_time_sncf
 # For perf benches:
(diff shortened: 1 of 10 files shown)


kirin/cots/model_maker.py
```diff
@@ -263,6 +263,14 @@ def _record_and_log(self, logger, log_str):
         log_dict.update({'contributor': self.contributor})
         logger.info('metrology', extra=log_dict)
 
+    def _check_stop_time_consistency(self, last_stop_time_depart, projected_stop_time, pdp):
```
mbounabi: is it worth passing the whole list of pdp for a log in the exception?

kirin/cots/model_maker.py
```diff
@@ -263,6 +263,14 @@ def _record_and_log(self, logger, log_str):
...
+        if last_stop_time_depart:
+            for arrival_departure_toggle in ['Arrivee', 'Depart']:
+                if (projected_stop_time[arrival_departure_toggle] and
+                        projected_stop_time[arrival_departure_toggle] < last_stop_time_depart):
+                    raise InvalidArguments('invalid cost: stop_point\'s({}) {} time is less than previous departure time'.format(
```
mbounabi: typo: cost? You had to pay something?

kirin/cots/model_maker.py
```diff
@@ -360,6 +379,7 @@ def _make_trip_update(self, vj, json_train):
                     setattr(st_update, _status_map[arrival_departure_toggle], 'update')
                     setattr(st_update, _delay_map[arrival_departure_toggle], as_duration(cots_delay))
                     highest_st_status = _get_higher_status(highest_st_status, 'update')
+                    projected_stop_time[arrival_departure_toggle] += as_duration(cots_delay)
```
pbougue: It's weird that you have to recompute it, can't you pick what's already processed?
Internally, we use the base_stop_time from navitia + cots_delay, so you don't check exactly the same thing (when bases differ).

kirin/cots/model_maker.py
```diff
@@ -263,6 +263,14 @@ def _record_and_log(self, logger, log_str):
         log_dict.update({'contributor': self.contributor})
         logger.info('metrology', extra=log_dict)
 
+    def _check_stop_time_consistency(self, last_stop_time_depart, projected_stop_time, pdp):
+        if last_stop_time_depart:
+            for arrival_departure_toggle in ['Arrivee', 'Depart']:
+                if (projected_stop_time[arrival_departure_toggle] and
```
pbougue: minor (it may become to permissive): You may add a check on
```python
arrival_departure_toggle in projected_stop_time and
projected_stop_time[arrival_departure_toggle] is not None
```

tests/fixtures/cots_train_96231_add_stop_time_earlier_than_previous.json
```diff
@@ -0,0 +1,571 @@
...
+            "dateHeure": "2018-10-23T14:54:00+0200",
+            "dateHeureEmission": "2018-10-23T09:22:19+0000",
+            "pronosticBrut": 300,
+            "pronosticIV": 300,
+            "source": "IRE-GOP"
+          }
+        ],
+        "listeMotifArrivee": [],
+        "listeMotifDepart": [],
+        "rang": 11,
```
pbougue: The rank is supposed to be ordered: we actually read following the list's order, but we are not supposed to, if we switch to a closer implem of spec, it might change your test.

xlqian: i noticed that too... and i was too lazy to change that :p

kirin/cots/model_maker.py
```diff
@@ -263,6 +263,32 @@ def _record_and_log(self, logger, log_str):
...
+        #
+        # def _last_depart_is_earlier_than_projected():
+        #     if last_stop_time_depart:
+        #         return all((last_stop_time_depart < projected_departure if projected_departure else True,
+        #                    last_stop_time_depart < projected_arrival if projected_arrival else True))
+        #     return True
+        #
+        # if not _project_stop_time_is_consistent() or not _last_depart_is_earlier_than_projected():
+        #     raise InvalidArguments('invalid cots: stop_point\'s({}) time is not consistent'
+        #                                    .format(pdp_code))
```
xlqian: I have two implementation of this function... which one do you prefer?

pbougue: The first one :+1: (commented one seems little less understandable)

kirin/cots/model_maker.py
pbougue: OK, so we check consistency at that point and it's already good to reject it here.
If it's consistent here but inconsistent after internal process, we'll consider other solutions (and maybe remove that one if it's time-consuming). :+1:

PR #214, opened 2018-12-05
Handle detour

nonifier:
This is the first part of the work to add support for detour with Cots.
Only the nominal case is supported here with one trip update that has 2 stop time updates: 1 deletion, 1 addition.

```
A------B----(C)[deleted_for_detour]----E------F
        \                             /
          \---D [added_for_detour]---/
```


buildbotctp:

tests/mock_navitia/__init__.py
```diff
@@ -92,7 +94,12 @@ def mock_navitia_query(self, query, q=None):
             query_str += sep + param_name + '=' + param_value
             sep = '&'
 
-    mock = _mock_navitia_call[query_str]
+    try:
+        mock = _mock_navitia_call[query_str]
+    except:
+        print("cannot find query in the mock: {}".format(query_str))
```
xlqian: print?

tests/fixtures/cots_train_96231_added_for_detour.json
```diff
@@ -0,0 +1,554 @@
...
+                },
+                "listeHoraireProjeteArrivee":[
+                ],
+                "listeHoraireProjeteDepart":[
+                ],
+                "listeMotifArrivee":[
+                ],
+                "listeMotifDepart":[
+                ],
+                "rang":1,
```
xlqian: "range" doesn't increment correctly... (yeah, I know it's not used in code but sooner or later @pbougue will make the comment if you don't correct that  :D )

kirin/cots/model_maker.py
```diff
@@ -412,10 +409,12 @@ def _make_trip_update(self, vj, json_train):
...
-                    # new stop_time added also?
-                    # TODO: stop_time's consistency should also be tested once the feature is implemented
-                    self._record_and_log(logger, 'nouvelleVersion/listePointDeParcours/statutCirculationOPE == '
-                                                 '"{}" is not handled (yet)'.format(cots_stop_time_status))
+                    cots_stop_time = get_value(cots_traveler_time,
+                                               'dateHeure',
+                                               nullable=True)
+                    setattr(st_update, _add_map[arrival_departure_toggle], cots_stop_time)
+                    setattr(st_update, _status_map[arrival_departure_toggle], 'added_for_detour')
+
```
mbounabi: Shouldn't we check for highest_status here?

nonifier: Good point! I've actually done the checking once for every cases. Hopefully, that looks much nicer now.

tests/mock_navitia/st_713666.py
```diff
@@ -0,0 +1,84 @@
...
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+import navitia_response
+
+response = navitia_response.NavitiaResponse()
+
+response.query = 'stop_points/?filter=stop_area.has_code("CR-CI-CH", "0087-713666-BV")&count=1'
```
pbougue: :cry: This will not conflict with #212 but tests will break after merge if I don't rebase :wink: 
I add a TODO to my PR in case it's merged after yours (and if it's the opposite we'll have to change it in yours)

migrations/versions/28958470072c_merge_revision_update.py
```diff
@@ -0,0 +1,22 @@
...
+
+Revision ID: 28958470072c
+Revises: ('174583a01aea', '45f4c90aa775')
+Create Date: 2018-12-05 12:16:18.141329
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '28958470072c'
+down_revision = ('174583a01aea', '45f4c90aa775')
```
pbougue: Why not just change down_revision in the other file?
Would be cleaner and less "concurrent-prone"

kirin/cots/model_maker.py
```diff
@@ -385,41 +362,36 @@ def _make_trip_update(self, vj, json_train):
...
 
                 elif cots_stop_time_status == 'SUPPRESSION':
                     # partial delete
                     setattr(st_update, _status_map[arrival_departure_toggle], 'delete')
-                    highest_st_status = _get_higher_status(highest_st_status, 'delete')
 
                 elif cots_stop_time_status == 'SUPPRESSION_DETOURNEMENT':
                     # stop_time is replaced by another one
                     setattr(st_update, _status_map[arrival_departure_toggle], 'deleted_for_detour')
-                    highest_st_status = _get_higher_status(highest_st_status, 'deleted_for_detour')
```
pbougue: same here, wrong merge?

kirin/core/handler.py
```diff
@@ -231,11 +231,11 @@ def _get_update_info_of_stop_time(base_time, input_status, input_delay):
 
 def _make_stop_time_update(base_arrival, base_departure, last_departure, input_st, stop_point, order):
     dep, dep_status, dep_delay = _get_update_info_of_stop_time(base_departure,
-                                                               input_st.departure_status,
-                                                               input_st.departure_delay)
+                                                   input_st.departure_status,
```
pbougue: weird indent remodeling

kirin/cots/model_maker.py
```diff
@@ -385,42 +362,36 @@ def _make_trip_update(self, vj, json_train):
...
+                    setattr(st_update, _status_map[arrival_departure_toggle], 'added_for_detour')
+                    setattr(st_update, _add_map[arrival_departure_toggle], base_schedule_datetime)
 
                 else:
                     raise InvalidArguments('invalid value {} for field horaireVoyageur{}/statutCirculationOPE'.
                                            format(cots_stop_time_status, arrival_departure_toggle))
 
+
+                arr_dep_status = getattr(st_update, _status_map[arrival_departure_toggle], 'nochange')
+                highest_st_status = _get_higher_status(highest_st_status, arr_dep_status)
```
pbougue: :+1:

tests/fixtures/cots_train_96231_added_for_detour.json
```diff
@@ -0,0 +1,554 @@
...
+                ],
+                "listeHoraireProjeteDepart":[
+                ],
+                "idMotifInterneDepartReference": 3,
+                "idMotifInterneArriveeReference": 3,
+                "listeMotifArrivee":[
+                ],
+                "listeMotifDepart":[
+                ],
+                "rang":3,
```
pbougue: @xlqian The remark I would have done (hypothetically of course :wink: ):
Is it possible to get 2 `"rang":3` in a real case?
Anyway, related to possible wrong consistency-exception mentionned above, we should have different "rang" so have `4` here

tests/fixtures/cots_train_96231_added_for_detour.json
```diff
@@ -0,0 +1,554 @@
...
+                },
+                "horaireProductionDepart":{
+                    "idFuseauHoraire":"Europe/Paris",
+                    "heureIncertaine":false,
+                    "heureLocale":"17:53:00",
+                    "nbJoursPasseMinuit":0
+                },
+                "horaireVoyageurArrivee":{
+                    "dateHeure":"2015-09-21T15:58:00+0000",
+                    "heureLocale":"17:58:00",
```
pbougue: We should test that it's OK with a added stop_time **earlier** than removed one AND placed **after** in list's order and also considering the `rang` field.
So have all `"dateHeure"` fields before 15:44 for that added stop_time.
It shouldn't be a consistency issue but I suspect it may be detected as one.
If you don't want to check yourself, I can !

tests/mock_navitia/st_713666.py
```diff
@@ -0,0 +1,84 @@
...
+      "links": [],
+      "fare_zone": {
+        "name": "0"
+      },
+      "id": "stop_point:OCE:SP:TrainTER-87713666",
+      "coord": {
+        "lat": "47.666666",
+        "lon": "4.666666"
+      },
+      "label": "Malin"
```
pbougue: Better to be "malin" :smile:

migrations/versions/174583a01aea_add_new_modification_types.py
```diff
@@ -0,0 +1,23 @@
...
+
+from alembic import op
+
+def upgrade():
+    op.execute("ALTER type modification_type ADD VALUE 'deleted_for_detour'")
+    op.execute("ALTER type modification_type ADD VALUE 'added_for_detour'")
+
+def downgrade():
+    """
+    Postgresql doesn't support the removal of a specific enum value
```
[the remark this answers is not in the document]
nonifier: This is now updated and tested on my machine ;)

tests/mock_navitia/st_713666.py
nonifier: Welcome to git flow. I've had to merge/rebase myself already. That's part of the game !

tests/fixtures/cots_train_96231_added_for_detour.json
nonifier: It's actually taken from a real COTS coming from Matrix as an example. 
It actually makes sense as we replace a stop_time by another.

tests/fixtures/cots_train_96231_added_for_detour.json
nonifier: That's planned in another PR. We've added a test like this with @xlqian, but it needs significant changes in the code base and that would have made this PR much less digest. 

It will come.

tests/mock_navitia/st_713666.py
nonifier: 666 number of the `malin`

PR #221, opened 2019-01-10
Add delays-after-delete test

benoit-bst:
Test : Feeds Kirin with a `delayed COTS` after a `delete COTS`

JIRA : https://jira.kisio.org/browse/NAVP-1132


tests/integration/cots_test.py
```diff
@@ -613,13 +613,22 @@ def test_cots_added_and_deleted_stop_time():
...
         assert StopTimeUpdate.query.all()[3].created_at == created_at_for_delete
 
+    cots_delayed_file = get_fixture_data('cots_train_96231_delayed.json')
+    res = api_post('/cots', data=cots_delayed_file)
+    assert res == 'OK'
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 5
+        assert len(TripUpdate.query.all()) == 1
+        assert TripUpdate.query.all()[0].company_id == 'company:OCE:SN'
+        assert len(StopTimeUpdate.query.all()) == 6
```
benoit-bst: why is `6` here ?

kadhikari: Because in nouvelleVersion.listePointDeParcours contains 
- 6 elements in cots_train_96231_delayed.json
- 7 elements in cots_train_96231_deleted.json

benoit-bst: :+1:

benoit-bst:
Because in the previous ticket, it was `Add -> Delete > delay` (https://jira.kisio.org/browse/NAVP-1095). 
It lacked the delay...

tests/integration/cots_test.py
```diff
@@ -613,13 +613,22 @@ def test_cots_added_and_deleted_stop_time():
...
         assert TripUpdate.query.all()[0].effect == 'REDUCED_SERVICE'
-        assert TripUpdate.query.all()[0].company_id == 'company:OCE:TH'
+        assert TripUpdate.query.all()[0].company_id == 'company:OCE:SN'
         assert len(StopTimeUpdate.query.all()) == 7
         assert StopTimeUpdate.query.all()[3].arrival_status == 'delete'
         assert StopTimeUpdate.query.all()[3].departure_status == 'delete'
         # It has already been deleted, so it is not allowed to deleted once again.
         assert StopTimeUpdate.query.all()[3].created_at == created_at_for_delete
 
+    cots_delayed_file = get_fixture_data('cots_train_96231_delayed.json')
```
xlqian: According to @papailio , the `SUPPRESSION` block will always be present even though the stop_time is deleted... 
you cannot use `'cots_train_96231_delayed.json'` and you have to create a new json,  which also means you must have 7 stop_times in the db,

xlqian: let's talk about it tomorrow

benoit-bst: hohoho, news are coming

benoit-bst: Finally, we called _Nadine_ to check his information, and the answer his : 
We keep `SUPPRESSION` into the new COTS flow that contains `DELAYS`

In fact, `SUPPRESSION` persist in the COTS flow until his death... and this is the unique information type wich can persist

So I'll update my test according to this rule

tests/integration/cots_test.py
```diff
@@ -613,13 +613,23 @@ def test_cots_added_and_deleted_stop_time():
...
         assert StopTimeUpdate.query.all()[3].created_at == created_at_for_delete
 
+    cots_delayed_file = get_fixture_data('cots_train_96231_deleted_and_delayed.json')
+    res = api_post('/cots', data=cots_delayed_file)
+    assert res == 'OK'
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 5
+        assert len(TripUpdate.query.all()) == 1
+        assert TripUpdate.query.all()[0].company_id == 'company:OCE:SN'
+        assert len(StopTimeUpdate.query.all()) == 6
```
benoit-bst: After change, there is a probleme here, `StopTimeUpdate.query.all()` have to be equal to 7... 

Indeed, as I said, `supression` must be keep during all the `suppression life`, COTS after COTS. This test highlight a **Bug**.
Poke @pbougue

tests/integration/utils_sncf_test.py
```diff
@@ -49,7 +49,7 @@ def check_db_96231_delayed(contributor=None, motif_externe_is_null=False):
         assert db_trip_delayed.vj_id == db_trip_delayed.vj.id
         assert db_trip_delayed.status == 'update'
         # Cots contain delayed stop_times only
-        assert db_trip_delayed.effect == 'SIGNIFICANT_DELAYS'
+        assert db_trip_delayed.effect == expected_effect
```
pbougue: minor: I'd outsource that test (to avoid too much params)

tests/integration/cots_test.py
pbougue: Agreed, since we now know that SUPPRESSION never disappear we have to keep it.
Basically Kirin has no memory at that point, when reading COTS.

It's not an absolute necessity to have some as the COTS feed is supposed to have it.

kirin/core/handler.py
```diff
@@ -316,6 +316,13 @@ def is_new_stop_event_valid(event_name, stop_id, stop_order, nav_stop, db_tu, ne
     if new_stu is None:
         return False
 
+    # Re-sending the same status is valid (COTS always re-send deleted or added stops)
+    if db_tu is not None:
```
nonifier: IMO, checking for the obj to be `is not None` every time is very verbose and cumbersome. 
I would rather have something much simpler like:
```python
if db_stu and new_stu:
    stop_event_status = db_stu.get_stop_event_status(event_name)
    new_stop_event_status = new_stu.get_stop_event_status(event_name)
    if stop_event_status and stop_event_status == new_stop_event_status:
        return True
```

pbougue: OK, will change on new_stu, db_st and db_stu, but on enum I don't want to get a bad implicit cast ;)

PR #222, opened 2019-01-10
fix delete the previously added stop_time

xlqian:
When it comes a deletion of a previously added stop_time at the beginning of a VJ ('add' then 'delete'), Kraken rejects the gtfs-rt, because `arrival` of the first stop_time is 0(timestamp).

In this PR, I just take the arrival/departure which was previously stored in the db.

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -411,9 +411,9 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
                 else:
                     # selection of extra iterable stop-times
                     sp_id = new_stu.stop_id
-                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id.encode('utf-8'), stop_order=order,
+                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id, stop_order=order,
                                                nav_stop=None, db_tu=db_trip_update, new_stu=new_stu) or \
-                            is_new_stop_event_valid(event_name='departure', stop_id=sp_id.encode('utf-8'), stop_order=order,
+                            is_new_stop_event_valid(event_name='departure', stop_id=sp_id, stop_order=order,
                                                     nav_stop=None, db_tu=db_trip_update, new_stu=new_stu):
                         # It is an added stop_time or a modification on a previously added stop_time, create a
                         # new "fake" Navitia stop time (even if it's not in navitia kirin needs to iterate on it)
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -447,8 +447,7 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
 
         # we compute the arrival time and departure time on base schedule and take past mid-night into
         # consideration
-        base_arrival = utc_nav_arrival_time
-        base_departure = utc_nav_departure_time
+        base_arrival = base_departure = None
         stop_id = navitia_stop.get('stop_point', {}).get('id')
         new_st = new_trip_update.find_stop(stop_id, nav_order)
 
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -411,9 +411,9 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
                 else:
                     # selection of extra iterable stop-times
                     sp_id = new_stu.stop_id
-                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id, stop_order=order,
+                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id.encode('utf-8'), stop_order=order,
                                                nav_stop=None, db_tu=db_trip_update, new_stu=new_stu) or \
-                            is_new_stop_event_valid(event_name='departure', stop_id=sp_id, stop_order=order,
+                            is_new_stop_event_valid(event_name='departure', stop_id=sp_id.encode('utf-8'), stop_order=order,
                                                     nav_stop=None, db_tu=db_trip_update, new_stu=new_stu):
                         # It is an added stop_time or a modification on a previously added stop_time, create a
                         # new "fake" Navitia stop time (even if it's not in navitia kirin needs to iterate on it)
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -447,7 +447,8 @@ def merge(navitia_vj, db_trip_update, new_trip_update, is_new_complete=False):
 
         # we compute the arrival time and departure time on base schedule and take past mid-night into
         # consideration
-        base_arrival = base_departure = None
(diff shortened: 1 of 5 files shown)


benoit-bst:
Becareful @xlqian, this is the same code part of my PR (https://github.com/CanalTP/kirin/pull/221)

xlqian:
tested locally with Artemis NG, all tests passed with this PR

kirin/core/handler.py
```diff
@@ -402,18 +402,25 @@ def get_next_stop():
                 else:
                     # selection of extra iterable stop-times
                     sp_id = new_stu.stop_id
-                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id, stop_order=order,
+                    if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id.encode('utf-8'), stop_order=order,
```
pbougue: minor: this encore can be moved to the line on top
Also, it should not be needed anymore with https://github.com/CanalTP/kirin/pull/220/files, no?

kirin/core/handler.py
```diff
@@ -402,18 +402,25 @@ def get_next_stop():
...
                         # It is an added stop_time or a modification on a previously added stop_time, create a
                         # new "fake" Navitia stop time (even if it's not in navitia kirin needs to iterate on it)
-                        added_st = {
-                            'stop_point': new_stu.navitia_stop,
-                            'utc_departure_time': extract_str_utc_time(new_stu.departure),
-                            'utc_arrival_time': extract_str_utc_time(new_stu.arrival),
-                        }
-                        yield order, added_st
+                        stu = None
+                        if db_trip_update:
```
pbougue: Not super fan of doing that here: exploring other places on my side.

kirin/core/handler.py
nonifier: What about the case where **none** of the `arrival` and `departure` return `true` for `is_new_stop_event_valid` ?
You might want to raise an error in the `else` case for this condition.

kirin/core/handler.py
nonifier: I agree, I would (at least) rather see this bloc abstracted out in a function like (that could be moved/re-used later):

```python
[...]
if is_new_stop_event_valid(event_name='arrival', stop_id=sp_id, stop_order=order [...] or \
        is_new_stop_event_valid(event_name='departure [...] :
    # It is an added stop_time or a modification on a previously added stop_time, create a
    # new "fake" Navitia stop time (even if it's not in navitia kirin needs to iterate on it)
    yield order, make_fake_realtime_stop_time(order, new_stu, db_trip_update)
```

kirin/core/handler.py
pbougue: This is not exactly related the purpose of this PR, but in that case Kirin just skips that stop, and `is_new_stop_event_valid` outputs a warning.

tests/integration/cots_test.py
```diff
@@ -626,6 +627,7 @@ def test_cots_added_and_deleted_stop_time():
         assert StopTimeUpdate.query.all()[3].arrival_status == 'delete'
         assert StopTimeUpdate.query.all()[3].departure_status == 'delete'
         # No change is stored in db (nothing sent to navitia) as the state is the same
+        # It has already been deleted, so it is not allowed to deleted once again.
```
pbougue: wrong comment

tests/fixtures/cots_train_96231_add_first_then_delete.json
```diff
@@ -0,0 +1,562 @@
...
+          "heureLocale": "16:30:00",
+          "nbJoursPasseMinuit": 0
+        },
+        "horaireVoyageurDepart": {
+          "dateHeure": "2015-09-21T14:30:00+0000",
+          "heureLocale": "16:30:00",
+          "idFuseauHoraire": "Europe/Paris",
+          "nbJoursPasseMinuit": 0
+        },
+        "horaireProductionArrivee": {
```
pbougue: Shouldn't that arrival disappear also?
Not sure how it's managed by SNCF in "real world".

Also the naming of the file should be more `cots_train_96231_delete_first_previously_added.json` (as it doesn't do both add and delete)

tests/integration/cots_test.py
```diff
@@ -664,6 +665,21 @@ def test_cots_added_stop_time_first_position():
...
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 2
+        assert len(TripUpdate.query.all()) == 1
+        assert TripUpdate.query.all()[0].status == 'update'
+        assert TripUpdate.query.all()[0].effect == 'REDUCED_SERVICE'
+        assert TripUpdate.query.all()[0].company_id == 'company:OCE:TH'
+        assert len(StopTimeUpdate.query.all()) == 7
+        assert StopTimeUpdate.query.all()[0].arrival_status == 'none'
+        assert StopTimeUpdate.query.all()[0].departure_status == 'delete'
+        assert StopTimeUpdate.query.all()[0].departure == datetime(2015, 9, 21, 14, 20)
```
pbougue: Maybe we should also test that we have a correct value at second stop (base-schedule's first stop), especially at arrival.
This is also true for the test above.

kirin/core/handler.py
pbougue: Looking at it closer, it seems to follow the logic that's around here:
yield base-stop, otherwise old-db-stop, otherwise new stop...
And as the base yield is handled just above, I think it's better to let it all here.

I would just change the order of code (`if stu is not None` first) and comment a bit.

And we probably have to make the code more robust in Kraken and elsewhere in Kirin but it seems OK to me.

tests/integration/cots_test.py
```diff
@@ -664,6 +664,26 @@ def test_cots_added_stop_time_first_position():
...
+        # Here is the real departure
+        assert StopTimeUpdate.query.all()[1].arrival_status == 'none'
+        assert StopTimeUpdate.query.all()[1].departure_status == 'none'
+        assert StopTimeUpdate.query.all()[1].departure_status == 'none'
```
pbougue: Extra copy/paste ;)
minor: Can you add the same tests on `StopTimeUpdate.query.all()[1]` just after line 666 (arrival may also be "add") ?

tests/integration/cots_test.py
```diff
@@ -625,7 +626,6 @@ def test_cots_added_and_deleted_stop_time():
         assert len(StopTimeUpdate.query.all()) == 7
         assert StopTimeUpdate.query.all()[3].arrival_status == 'delete'
         assert StopTimeUpdate.query.all()[3].departure_status == 'delete'
-        # No change is stored in db (nothing sent to navitia) as the state is the same
```
pbougue: This comment was actually useful.
(previously there were 2 comments from your rebase, and this one is the good one)

PR #226, opened 2019-01-30
Fix add+delay stop reading

pbougue:
Mostly corrects the way model is filled when reading an ADDED stop (for detour or not).
JIRA: https://jira.kisio.org/browse/NAVP-1153

:mag_right: Please review by commit, should be simpler (read their descriptions).

TODO (`do_not_merge` is for that):
- [x] run locally on artemis
  > :heavy_check_mark: Runs OK after 2 corrections **to merge** when merging this PR:  
  https://github.com/CanalTP/artemis/pull/296  
  https://github.com/CanalTP/artemis_references/pull/158
- [x] more hand-test with data provided in JIRA ticket.  
  > :heavy_check_mark: just the last departure's status is a bit weird (like it was the case already), but it's not a real stop-event, so...
- [x] add tests on simple "add+delay" feed, and on chaining "add, then add+delay" (if not already tested)
  > detected a chaining bug, added correction for that

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -194,7 +194,17 @@ def _get_datetime(utc_circulation_date, utc_time):
     return datetime.datetime.combine(utc_circulation_date, utc_time)
 
 
-def _get_update_info_of_stop_time(base_time, input_time, input_status, input_delay):
+def _get_update_info_of_stop_event(base_time, input_time, input_status, input_delay):
+    """
+    Process information for a given stop event: given information available, compute info to be stored in db.
+    :param base_time: datetime in base_schedule
+    :param input_time: datetime in new feed
+    :param input_status: status in new feed
+    :param input_delay: delay in new feed
+    :return: new_time (base-schedule datetime in most case),
+             status (update, delete, ...)
+             delay (new_time + delay = RT datetime)
+    """
     new_time = None
     status = ModificationType.none.name
     delay = datetime.timedelta(0)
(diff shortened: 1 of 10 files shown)


kirin/core/handler.py
```diff
@@ -230,17 +211,21 @@ def _get_update_info_of_stop_time(base_time, input_status, input_delay):
         status = input_status
     elif input_status in (ModificationType.add.name, ModificationType.added_for_detour.name):
         status = input_status
-        new_time = base_time
+        new_time = input_time.replace(tzinfo=None) if input_time else None
```
pbougue: Working with naive datetime from base (compare, etc.) so here remove UTC TZ.

kirin/core/handler.py
```diff
@@ -214,12 +214,14 @@ def _get_datetime(utc_circulation_date, utc_time):
     return datetime.datetime.combine(utc_circulation_date, utc_time)
 
 
-def _get_update_info_of_stop_time(base_time, input_status, input_delay):
+def _get_update_info_of_stop_time(base_time, input_time, input_status, input_delay):
```
mbounabi: Some comments about what is done in this function can be useful

kirin/cots/model_maker.py
```diff
@@ -344,58 +363,68 @@ def _make_trip_update(self, vj, json_train):
...
+                    setattr(st_update, _stop_event_datetime_map[arrival_departure_toggle],
+                            projected_stop_time[arrival_departure_toggle])
+                    # delay already added to stop_event datetime
+                    setattr(st_update, _delay_map[arrival_departure_toggle], None)
 
                 elif cots_stop_time_status == 'DETOURNEMENT':
-                    setattr(st_update, _status_map[arrival_departure_toggle], 'added_for_detour')
-                    setattr(st_update, _add_map[arrival_departure_toggle], base_schedule_datetime)
+                    # new stop_time added to replace another one
+                    cots_base_datetime = _retrieve_stop_event_datetime(cots_traveler_time)
```
xlqian: can you make a function for the same codes above?

kadhikari: If i am not wrong, the following code is repeated three times:
"cots_base_datetime = _retrieve_stop_event_datetime(cots_traveler_time)	
if cots_base_datetime:
	projected_stop_time[arrival_departure_toggle] = cots_base_datetime
cots_delay = _retrieve_stop_event_delay(pdp, arrival_departure_toggle)
if cots_delay is not None:
	projected_stop_time[arrival_departure_toggle] += cots_delay
"

kadhikari:
Ok for me except the block of code repeated three times.

kirin/cots/model_maker.py
pbougue: The code isn't exactly the same for "update" case...
So I mutualized code for ADD and DETOUR only, and explained a bit more update.

What could be mutualized is how projected_stop_time is updated, but it's kind of a side effect for consistency, so I don't feel like externalizing it.

kirin/core/handler.py
pbougue: Explained a bit more and renamed... Hope it's enough.

PR #227, opened 2019-02-11
Add trip update in bdd and gtfs rt

kadhikari:
This PR manages an added trip with stop_times and the details are as follows:

- Attribut physical_mode_id added in the table trip_update
- call to navitia to get physical_mode_id using "indicateurFer" and "codeMarqueTransporteur" in the flux cots.
- Extention physical_mode_id added VehicleDescriptor
- Some tests added

Concerned tickets:
https://jira.kisio.org/browse/NAVP-1061
https://jira.kisio.org/browse/NAVP-1062
https://jira.kisio.org/browse/NAVP-1063

### DIFF
--- a/kirin/cots/model_maker.py
+++ b/kirin/cots/model_maker.py
@@ -237,13 +237,13 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
             raise InvalidArguments('No object "nouvelleVersion" available in feed provided')
 
         dict_version = get_value(json, 'nouvelleVersion')
-        vjs = self._get_vjs(dict_version)
-
-        trip_updates = [self._make_trip_update(vj, dict_version) for vj in vjs]
+        status_op = get_value(dict_version, 'statutOperationnel', 'PERTURBEE')
+        vjs = self._get_vjs(dict_version, status_op=status_op)
+        trip_updates = [self._make_trip_update(dict_version, vj) for vj in vjs]
 
         return trip_updates
 
-    def _get_vjs(self, json_train):
+    def _get_vjs(self, json_train, status_op='PERTURBEE'):
         train_numbers = get_value(json_train, 'numeroCourse')
         pdps = _retrieve_interesting_pdp(get_value(json_train, 'listePointDeParcours'))
         if not pdps:
@@ -252,14 +252,18 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
 
         # retrieve base-schedule's first departure and last arrival
         def get_first_not_fully_added(list_pdps, hour_obj_name):
-            p = next(p for p in list_pdps if not _is_fully_added_pdp(p))
+            # For an added trip we always take first stop_time.
+            if status_op == 'AJOUTEE':
+                p = next(p for p in list_pdps)
+            else:
+                p = next(p for p in list_pdps if not _is_fully_added_pdp(p))
             str_time = get_value(get_value(p, hour_obj_name), 'dateHeure') if p else None
             return as_utc_dt(str_time) if str_time else None
 
         utc_vj_start = get_first_not_fully_added(pdps, 'horaireVoyageurDepart')
         utc_vj_end = get_first_not_fully_added(reversed(pdps), 'horaireVoyageurArrivee')
 
-        return self._get_navitia_vjs(train_numbers, utc_vj_start, utc_vj_end)
+        return self._get_navitia_vjs(train_numbers, utc_vj_start, utc_vj_end, status_op=status_op)
 
     def _record_and_log(self, logger, log_str):
         log_dict = {'log': log_str}
@@ -282,7 +286,7 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
             raise InvalidArguments('invalid cots: stop_point\'s({}) time is not consistent'
                                    .format(pdp_code))
 
-    def _make_trip_update(self, vj, json_train):
+    def _make_trip_update(self, json_train, vj=None):
         """
         create the new TripUpdate object
         Following the COTS spec: https://github.com/CanalTP/kirin/blob/master/documentation/cots_connector.md
@@ -307,17 +311,16 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
 
         elif trip_status == 'AJOUTEE':
             # the trip is created from scratch
-            # not handled yet
-            self._record_and_log(logger, 'nouvelleVersion/statutOperationnel == "AJOUTEE" is not handled (yet)')
             trip_update.effect = TripEffect.ADDITIONAL_SERVICE.name
-            return trip_update
+            trip_update.status = ModificationType.add.name
 
         # all other status is considered an 'update' of the trip_update and effect is calculated
         # from stop_time status list. This part is also done in kraken and is to be deleted later
         # Ordered stop_time status= 'nochange', 'add', 'delete', 'update'
         # 'nochange' or 'update' -> SIGNIFICANT_DELAYS, add -> MODIFIED_SERVICE, delete = DETOUR
-        trip_update.status = ModificationType.update.name
-        trip_update.effect = TripEffect.MODIFIED_SERVICE.name
+        else:
+            trip_update.status = ModificationType.update.name
+            trip_update.effect = TripEffect.MODIFIED_SERVICE.name
 
         # Initialize stop_time status to nochange
         highest_st_status = ModificationType.none.name
@@ -330,7 +333,7 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
         # manage realtime information stop_time by stop_time
         for pdp in pdps:
             # retrieve navitia's stop_point corresponding to the current COTS pdp
-            nav_stop, log_dict = self._get_navitia_stop_point(pdp, vj.navitia_vj)
+            nav_stop, log_dict = self._get_navitia_stop_point(pdp, vj)
             projected_stop_time = {'Arrivee': None, 'Depart': None}  # used to check consistency
 
             if log_dict:
@@ -424,10 +427,11 @@ class KirinModelBuilder(AbstractSNCFKirinModelBuilder):
             last_stop_time_depart = projected_stop_time['Depart']
 
         # Calculates effect from stop_time status list(this work is also done in kraken and has to be deleted)
-        trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
+        if trip_update.effect != TripEffect.ADDITIONAL_SERVICE.name:
+            trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
         return trip_update
 
-    def _get_navitia_stop_point(self, pdp, nav_vj):
+    def _get_navitia_stop_point(self, pdp, vj=None):
         """
         Get a navitia stop point from the stop_time in a 'Point de Parcours' dict.
         The dict MUST contain cr, ci, ch tags.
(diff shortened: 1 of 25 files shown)


tests/integration/cots_test.py
```diff
@@ -959,3 +959,90 @@ def test_chain_add_no_delay_and_add_with_delay_around():
...
+    assert status[2].arrival == datetime(2012, 11, 20, 14, 00)
+    assert status[2].departure_status == 'add'
+    assert status[2].departure == datetime(2012, 11, 20, 14, 10)
+    assert status[3].arrival_status == 'add'
+    assert status[3].arrival == datetime(2012, 11, 20, 15, 00)
+    assert status[3].departure_status == 'add'
+    assert status[3].departure == datetime(2012, 11, 20, 15, 10)
+    assert status[4].arrival_status == 'add'
+    assert status[4].arrival == datetime(2012, 11, 20, 16, 00)
+    assert status[4].departure_status == 'none'
```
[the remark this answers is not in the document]
kadhikari: Except : "statutOperationnel": "PERTURBEE",

tests/integration/cots_test.py
```diff
@@ -959,3 +959,90 @@ def test_chain_add_no_delay_and_add_with_delay_around():
...
+        check_add_trip_151515()
+
+    cots_add_file = get_fixture_data('cots_train_151515_added_trip_with_delay.json')
+    res = api_post('/cots', data=cots_add_file)
+    assert res == 'OK'
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 2
+        check_add_trip_151515_with_delay()
+
+    cots_add_file = get_fixture_data('cots_train_151515_added_trip_to_normal.json')
```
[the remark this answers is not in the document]
kadhikari: Except "statutOperationnel": "PERTURBEE"

migrations/versions/51d98b3b8e58_add_physical_mode_id.py
```diff
@@ -0,0 +1,22 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '51d98b3b8e58'
+down_revision = '174583a01aea'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
```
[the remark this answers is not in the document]
kadhikari: As you said, it is a simple column and can be deleted while rolling back.

tests/integration/cots_test.py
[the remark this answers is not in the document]
kadhikari: Yes

tests/integration/cots_test.py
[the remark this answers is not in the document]
benoit-bst: hooooo Yes :bulb:, thanks

kirin/cots/model_maker.py
```diff
@@ -205,6 +208,38 @@ def _is_fully_added_pdp(pdp):
...
+    """
+    Verify if trip in flux cots is a newly added or a modification on a newly added trip
+    :param dict_version: Value of attribut nouvelleVersion in Flux cots
+    :return: True or False
+    """
+    is_added_trip = get_value(dict_version, 'statutOperationnel', 'PERTURBEE') == 'AJOUTEE'
+    if is_added_trip:
+        return True
+    if not is_added_trip:
+        utc_vj_start = _get_first_not_fully_added(pdps, 'horaireVoyageurDepart', is_added_trip=True)
```
pbougue: It's misleading: You pass `is_added_trip=True` here, but from above it's not the case, changing signature of _get_first_not_fully_added() would help here.

kirin/cots/model_maker.py
```diff
@@ -205,6 +208,38 @@ def _is_fully_added_pdp(pdp):
     return dep_arr_statuses and all(s == 'CREATION' for s in dep_arr_statuses)
 
 
+def _get_first_not_fully_added(list_pdps, hour_obj_name, is_added_trip=False):
```
pbougue: naming and comment is not corresponding to what the code does
the signature would be clearer to me like so:
`def _get_first_stop(list_pdps, hour_obj_name, skip_fully_added_stops=True)`
Then, not sure it needs any comment.

kirin/cots/model_maker.py
```diff
@@ -205,6 +208,38 @@ def _is_fully_added_pdp(pdp):
...
+    """
+    Retrieve base-schedule's first departure and last arrival
```
pbougue: Comment is misleading: It's not related to base schedule directly (we're _trying_ to deduce it sometimes, but not always).

kirin/cots/model_maker.py
```diff
@@ -205,6 +208,38 @@ def _is_fully_added_pdp(pdp):
...
+    :return: True or False
+    """
+    is_added_trip = get_value(dict_version, 'statutOperationnel', 'PERTURBEE') == 'AJOUTEE'
+    if is_added_trip:
+        return True
+    if not is_added_trip:
+        utc_vj_start = _get_first_not_fully_added(pdps, 'horaireVoyageurDepart', is_added_trip=True)
+        utc_vj_start = utc_vj_start - timedelta(hours=1)
+        train_id = TRAIN_ID_FORMAT.format(train_numbers)
+        db_trip_update = model.TripUpdate.find_by_dated_vj(train_id, utc_vj_start)
```
pbougue: We should probably do the same kind of search done in navitia: train starting between vj_start-1 and vj_end+1.
This would be safer/ more resilient if COTS feeds are not consistent one after another.

Also that 1 should be shared param here like `COTS_SEARCH_MARGIN=timedelta(hours=1)`.

kirin/cots/model_maker.py
```diff
@@ -438,10 +471,14 @@ def _get_navitia_stop_point(self, pdp, nav_vj):
...
-        nav_st, log_dict = get_navitia_stop_time_sncf(cr=get_value(pdp, 'cr'),
-                                                      ci=get_value(pdp, 'ci'),
-                                                      ch=get_value(pdp, 'ch'),
-                                                      nav_vj=nav_vj)
+        if vj and vj.navitia_vj:
+            nav_st, log_dict = get_navitia_stop_time_sncf(cr=get_value(pdp, 'cr'),
+                                                          ci=get_value(pdp, 'ci'),
+                                                          ch=get_value(pdp, 'ch'),
+                                                          nav_vj=vj.navitia_vj)
+        else:
```
pbougue: I don't get when this case can happen?

kirin/cots/model_maker.py
```diff
@@ -424,10 +456,11 @@ def _make_trip_update(self, vj, json_train):
             last_stop_time_depart = projected_stop_time['Depart']
 
         # Calculates effect from stop_time status list(this work is also done in kraken and has to be deleted)
-        trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
+        if trip_update.effect != TripEffect.ADDITIONAL_SERVICE.name:
+            trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
```
pbougue: Not directly related to the goal of this PR, but: Shouldn't we also update trip_update.status ?

kirin/cots/model_maker.py
```diff
@@ -424,10 +456,11 @@ def _make_trip_update(self, vj, json_train):
             last_stop_time_depart = projected_stop_time['Depart']
 
         # Calculates effect from stop_time status list(this work is also done in kraken and has to be deleted)
-        trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
+        if trip_update.effect != TripEffect.ADDITIONAL_SERVICE.name:
```
pbougue: Here, the exact test is probably more `if trip_update.effect == TripEffect.MODIFIED_SERVICE.name` as it's the only case where we compute it from stops.
AJOUTEE or SUPPRIMEE should stay the same (although we early-return for SUPPRIMEE, it would be less bug-prone).

kirin/cots/model_maker.py
```diff
@@ -477,3 +514,20 @@ def _request_navitia_company(self, code):
...
+
+    def _get_navitia_physical_mode(self, indicator=None, cots_mode=None):
+        """
+        Get a navitia physical_mode for the codes present in COTS ("indicateurFer" and "codeMarqueTransporteur")
+        If the physical_mode doesn't exist in navitia, another request is made default physical_mode
+        "physical_mode.id = physical_mode:LongDistanceTrain"
+        """
+        return self._request_navitia_physical_mode(indicator, cots_mode) or self._request_navitia_physical_mode()
+
+    def _request_navitia_physical_mode(self, indicator=None, cots_mode=None):
```
pbougue: minor: Not sure this requires a specific method, you can probably merge those 2 methods.

kirin/cots/model_maker.py
```diff
@@ -305,19 +334,22 @@ def _make_trip_update(self, vj, json_train):
...
+            trip_update.physical_mode_id = self._get_navitia_physical_mode(cots_indicator, cots_mode)
+
 
         # all other status is considered an 'update' of the trip_update and effect is calculated
         # from stop_time status list. This part is also done in kraken and is to be deleted later
         # Ordered stop_time status= 'nochange', 'add', 'delete', 'update'
         # 'nochange' or 'update' -> SIGNIFICANT_DELAYS, add -> MODIFIED_SERVICE, delete = DETOUR
-        trip_update.status = ModificationType.update.name
-        trip_update.effect = TripEffect.MODIFIED_SERVICE.name
+        else:
```
pbougue: :mag_right: well-spotted! :+1:

tests/fixtures/cots_train_151515_added_trip_with_delay.json
```diff
@@ -0,0 +1,327 @@
...
+        "listeMotifDepart": [
+          {
+            "dateHeureDeclaration": "2012-11-20T08:35:06+0000",
+            "idMotifInterne": 78,
+            "source": "IRE-GOP"
+          }
+        ],
+        "rang": 2,
+        "typeArret": "CH",
+        "horaireVoyageurDepart": {
```
pbougue: minor: better to have it above with "horaireVoyageurArrivee"
(same below)

kirin/cots/model_maker.py
pbougue: Answering to myself: no, it's an update and stays that way.

kadhikari: It's already done above after the line 'elif is_added_trip:'

kirin/cots/model_maker.py
kadhikari: agree

tests/integration/populate_pb_test.py
```diff
@@ -416,3 +420,73 @@ def test_populate_pb_deleted_stop_times_status():
...
+        trip_update.effect = 'ADDITIONAL_SERVICE'
+        trip_update.physical_mode_id = 'physical_mode:LongDistanceTrain'
+        real_time_update = RealTimeUpdate(raw_data=None, connector='cots', contributor='realtime.cots')
+        real_time_update.trip_updates.append(trip_update)
+        st = StopTimeUpdate({'id': 'sa:1'}, departure=_dt("8:15"), departure_delay=timedelta(minutes=5),
+                            arrival=None, arr_status='none', dep_status='add')
+        trip_update.stop_time_updates.append(st)
+
+        st = StopTimeUpdate({'id': 'sa:2'},
+                            departure=_dt("8:21"), departure_delay=timedelta(minutes=-40),
```
pbougue: Why did you manage departure here? It's not managed currently in model_maker, no?

Also, I'd add a middle stop, to check that both arrival and departure are "ADDED".

kadhikari: Here since we check only the function 'convert_to_gtfsrt' i think two stops are enough. I added this to test the extension of physical_mode_id.

kirin/cots/model_maker.py
```diff
@@ -424,10 +454,11 @@ def _make_trip_update(self, vj, json_train):
             last_stop_time_depart = projected_stop_time['Depart']
 
         # Calculates effect from stop_time status list(this work is also done in kraken and has to be deleted)
-        trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
+        if trip_update.effect == TripEffect.MODIFIED_SERVICE.name:
+            trip_update.effect = get_effect_by_stop_time_status(highest_st_status)
         return trip_update
 
-    def _get_navitia_stop_point(self, pdp, nav_vj):
+    def _get_navitia_stop_point(self, pdp, vj=None):
```
pbougue: you can remove this change I guess

kirin/cots/model_maker.py
```diff
@@ -330,7 +360,7 @@ def _make_trip_update(self, vj, json_train):
         # manage realtime information stop_time by stop_time
         for pdp in pdps:
             # retrieve navitia's stop_point corresponding to the current COTS pdp
-            nav_stop, log_dict = self._get_navitia_stop_point(pdp, vj.navitia_vj)
+            nav_stop, log_dict = self._get_navitia_stop_point(pdp, vj)
```
pbougue: That change can be removed too, it's "chained" to the previous one.

kirin/cots/model_maker.py
```diff
@@ -441,7 +472,7 @@ def _get_navitia_stop_point(self, pdp, nav_vj):
         nav_st, log_dict = get_navitia_stop_time_sncf(cr=get_value(pdp, 'cr'),
                                                       ci=get_value(pdp, 'ci'),
                                                       ch=get_value(pdp, 'ch'),
-                                                      nav_vj=nav_vj)
+                                                      nav_vj=nav_vj.navitia_vj)
```
pbougue: That change can be removed too, it's "chained" to the previous one.

PR #229, opened 2019-02-20
kirin: Refuse a flux cots with AJOUTEE if already present in database

kadhikari:
Case found while treating https://jira.kisio.org/browse/NAVP-1068


kirin/cots/model_maker.py
```diff
@@ -227,18 +227,22 @@ def _is_added_trip(train_numbers, dict_version, pdps):
...
+                                          skip_fully_added_stops=False)
+    train_id = TRAIN_ID_FORMAT.format(train_numbers)
+    db_trip_update = model.TripUpdate.find_vj_by_period(train_id,
+                                                        start_date=utc_vj_start-SNCF_SEARCH_MARGIN,
+                                                        end_date=utc_vj_end+SNCF_SEARCH_MARGIN)
+    if db_trip_update and db_trip_update.status == ModificationType.delete.name:
+        raise InvalidArguments('Invalid action, trip {} already deleted in database'.format(train_numbers))
+    trip_in_db = db_trip_update and db_trip_update.status == ModificationType.add.name
+    if is_added_trip and trip_in_db:
+        raise InvalidArguments('Invalid action, trip {} already exist in database'.format(train_numbers))
```
pbougue: "trip {} `can not be added multiple times`" seems clearer

tests/integration/cots_test.py
```diff
@@ -1217,3 +1217,53 @@ def test_cots_for_added_trip_chain_type_3():
...
+        assert trips[0].status == 'delete'
+        assert trips[0].effect == 'NO_SERVICE'
+        assert trips[0].company_id == 'company:OCE:SN'
+        stop_times = StopTimeUpdate.query.all()
+        assert len(stop_times) == 0
+
+    cots_delete_file = get_fixture_data('cots_train_151515_deleted_trip_with_delay_and_stop_time_added.json')
+    res, status = api_post('/cots', check=False, data=cots_delete_file)
+    assert status == 400
+    assert 'Invalid action, trip 151515 already deleted in database' in res.get('error')
```
pbougue: It would be nice to also add it back at the end. It should work, allowing SNCF to restart from scratch through a "delete" if needed.

kirin/cots/model_maker.py
```diff
@@ -222,23 +222,29 @@ def _get_first_stop_datetime(list_pdps, hour_obj_name, skip_fully_added_stops=Tr
...
+    utc_vj_start = _get_first_stop_datetime(pdps, 'horaireVoyageurDepart', skip_fully_added_stops=False)
+    utc_vj_end = _get_first_stop_datetime(reversed(pdps), 'horaireVoyageurArrivee',
+                                          skip_fully_added_stops=False)
+    train_id = TRAIN_ID_FORMAT.format(train_numbers)
+    trip_added_in_db = model.TripUpdate.find_vj_by_period(train_id,
+                                                          start_date=utc_vj_start-SNCF_SEARCH_MARGIN,
+                                                          end_date=utc_vj_end+SNCF_SEARCH_MARGIN)
+
+    if trip_added_in_db:
+        if cots_trip_status == ADDED_TRIP and trip_added_in_db.status == ModificationType.add.name:
```
pbougue: I think we should separate responsibilities in 2 functions:
* the 2 `raise` are a consistency check
* the rest is computing if the trip has been/is added

kirin/cots/model_maker.py
```diff
@@ -222,23 +222,29 @@ def _get_first_stop_datetime(list_pdps, hour_obj_name, skip_fully_added_stops=Tr
...
+            raise InvalidArguments('Invalid action, trip {} can not be added multiple times'.format(train_numbers))
+        elif cots_trip_status != ADDED_TRIP and trip_added_in_db.status == ModificationType.delete.name:
+            raise InvalidArguments('Invalid action, trip {} already deleted in database'.format(train_numbers))
+        return True
```
pbougue: Here we probably want:
```python
return (trip_added_in_db.status == ModificationType.add.name)
```
-----
It's even weird that this `return True` doesn't break any test. Maybe we don't have enough chaining tests?
I think some would break if navitia's base-VJ differs from cots feed.
Adding a test that breaks with this `return True` would be nice.

kirin/cots/model_maker.py
```diff
@@ -222,23 +222,29 @@ def _get_first_stop_datetime(list_pdps, hour_obj_name, skip_fully_added_stops=Tr
...
+    elif cots_trip_status == ADDED_TRIP:
```
pbougue: minor: then we can directly
```python
return (cots_trip_status == ADDED_TRIP)
```
#oneliner

kirin/cots/model_maker.py
kadhikari: I don't find it necessary to put in a function as used only once. I added some comments though.

kirin/cots/model_maker.py
kadhikari: We should be able to manage a trip as "new trip" in the following conditions:
- Add a trip absent in database 
- Update a trip recently added (present in database as 'add')
- Add a trip recently deleted (present in database as 'delete')

kirin/abstract_sncf_model_maker.py
```diff
@@ -39,14 +39,17 @@
...
 
 TRAIN_ID_FORMAT = 'OCE:SN:{}'
 SNCF_SEARCH_MARGIN = timedelta(hours=1)
-ADDED_TRIP = 'AJOUTEE'
-DELETED_TRIP = 'SUPPRIMEE'
-MODIFIED_TRIP = 'PERTURBEE'
 
 
+class TripStatus(Enum):
+    AJOUTEE = 1,
```
pbougue: Ah yeah, forgot that it implied :fr: naming...
It's worth adding a comment for :gb: meaning, now.

The alternative (overkill to me) being to have a match function...

kirin/cots/model_maker.py
```diff
@@ -238,14 +238,20 @@ def _is_added_trip(train_numbers, dict_version, pdps):
                                                           end_date=utc_vj_end+SNCF_SEARCH_MARGIN)
 
     if trip_added_in_db:
-        if cots_trip_status == ADDED_TRIP and trip_added_in_db.status == ModificationType.add.name:
+        # Raise exception on not permitted actions
```
pbougue: minor: "on `forbidden`/`inconsistent` actions"

kirin/cots/model_maker.py
```diff
@@ -238,14 +238,20 @@ def _is_added_trip(train_numbers, dict_version, pdps):
...
             raise InvalidArguments('Invalid action, trip {} can not be added multiple times'.format(train_numbers))
-        elif cots_trip_status != ADDED_TRIP and trip_added_in_db.status == ModificationType.delete.name:
+        elif cots_trip_status != TripStatus.AJOUTEE.name and trip_added_in_db.status == ModificationType.delete.name:
             raise InvalidArguments('Invalid action, trip {} already deleted in database'.format(train_numbers))
-        return True
-    elif cots_trip_status == ADDED_TRIP:
-        return True
-    return False
+
+        # Can we handle as an added trip
```
pbougue: Not sure what you mean here, suggesting: "`Should` `be` handle`d` as an added trip"

kirin/cots/model_maker.py
```diff
@@ -222,24 +222,36 @@ def _get_first_stop_datetime(list_pdps, hour_obj_name, skip_fully_added_stops=Tr
...
-        utc_vj_end = _get_first_stop_datetime(reversed(pdps), 'horaireVoyageurArrivee',
-                                              skip_fully_added_stops=False)
-        train_id = TRAIN_ID_FORMAT.format(train_numbers)
-        db_trip_update = model.TripUpdate.find_vj_by_period(train_id,
-                                                            start_date=utc_vj_start-SNCF_SEARCH_MARGIN,
-                                                            end_date=utc_vj_end+SNCF_SEARCH_MARGIN)
-        if db_trip_update and db_trip_update.status == 'add':
-            return True
-    return False
+    cots_trip_status = get_value(dict_version, 'statutOperationnel',  TripStatus.PERTURBEE.name)
```
nonifier: enums FTW ! :+1:

PR #233, opened 2019-04-12
Datetime overflow bugfix

pbougue:
Bug happened on a VJ where the stop-event passing midnight (UTC time) is brought back to previous day by an early passage (negative delay).

Lead to a situation where time+delay < 0, and the resulting datetime was BC (not valid). So it's more an "underflow" :wink: 
> This situation was tested only when base-schedule stop-event was passing midnight

JIRA: https://jira.kisio.org/browse/NAVP-1255


pbougue:
Jenkins, test this please.

PR #235, opened 2019-04-15
Protecting generators a bit more

pbougue:
After https://jira.kisio.org/browse/NAVP-1280

Just being cautious: this case is protected prior to method's call, but as it's done outside of the method...


pbougue:
Jenkins, test this please.

PR #239, opened 2019-05-03
kirin: manage add existing trip in navitia

kadhikari:
Concerned Ticket: https://jira.kisio.org/browse/NAVP-1202

### DIFF
--- a/kirin/abstract_sncf_model_maker.py
+++ b/kirin/abstract_sncf_model_maker.py
@@ -52,9 +52,9 @@ class TripStatus(Enum):
 
 
 class ActionOnTrip(Enum):
-    NOT_ADDED = 1,
-    FIRST_TIME_ADDED = 2,
-    PREVIOUSLY_ADDED = 3
+    NOT_ADDED = 1,          #
+    FIRST_TIME_ADDED = 2,   # Add trip for the first time / delete followed by add
+    PREVIOUSLY_ADDED = 3    # add followed by update
 
 
 def make_navitia_empty_vj(headsign):
--- a/kirin/abstract_sncf_model_maker.py
+++ b/kirin/abstract_sncf_model_maker.py
@@ -51,6 +51,12 @@ class TripStatus(Enum):
     PERTURBEE = 3   # Modified or Impacted
 
 
+class ActionOnTrip(Enum):
+    NOT_ADDED = 1,
+    FIRST_TIME_ADDED = 2,
+    PREVIOUSLY_ADDED = 3
+
+
 def make_navitia_empty_vj(headsign):
     headsign = TRAIN_ID_FORMAT.format(headsign)
     return {"id": headsign, "trip": {"id": headsign}}
@@ -115,7 +121,7 @@ class AbstractSNCFKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
         self.navitia = nav
         self.contributor = contributor
 
-    def _get_navitia_vjs(self, headsign_str, utc_since_dt, utc_until_dt, is_added_trip=False):
+    def _get_navitia_vjs(self, headsign_str, utc_since_dt, utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name):
         """
         Search for navitia's vehicle journeys with given headsigns, in the period provided
         :param utc_since_dt: UTC datetime that starts the search period.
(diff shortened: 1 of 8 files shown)


kirin/abstract_sncf_model_maker.py
```diff
@@ -143,23 +143,24 @@ def _get_navitia_vjs(self, headsign_str, utc_since_dt, utc_until_dt, is_added_tr
...
-                })
 
+            navitia_vjs = self.navitia.vehicle_journeys(q={
+                'headsign': train_number,
+                'since': to_navitia_str(extended_since_dt),
+                'until': to_navitia_str(extended_until_dt),
+                'depth': '2',  # we need this depth to get the stoptime's stop_area
+                'show_codes': 'true'  # we need the stop_points CRCICH codes
+            })
+            if not is_added_trip:
```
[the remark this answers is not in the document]
kadhikari: 1. For an added trip, since the VJ has a validity_pattern with only zeros only, the request of the type 
/vehicle_journeys?headsign=151515&show_codes=true&data_freshness=realtime&since=20121120T100000%2B0000&until=20121120T170000%2B0000& to navitia gives NoSolution.

If this mode of operation on /vehicle_journeys (ptrefs) doesn't change, my correction in this PR should not block the modification of previously added trip.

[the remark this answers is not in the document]
pbougue: To keep trace of it after our call: You're right, it would keep on working.
But it's because we're lucky that /vehicle_journeys endpoint is crappy and doesn't return RT-vj... Which could (should?) legitimately change.
So I see 2 clean ways here:
* let the code as is (almost) and ensure that  /vehicle_journeys returns only base_schedule trips, for example with a `?data_freshness=base_schedule` param. This is not a feature implemented in navitia, so it would require designing/implementing it, or at least testing it in navitia to ensure stability.
* only test navitia's response on the first COTS feed add-trip (looks easier).

kirin/cots/model_maker.py
```diff
@@ -247,11 +248,14 @@ def _is_added_trip(train_numbers, dict_version, pdps):
             raise InvalidArguments('Invalid action, trip {} already deleted in database'.format(train_numbers))
 
         # Should be handled as an added trip
-        return (trip_added_in_db.status == ModificationType.add.name) or \
-               (cots_trip_status == TripStatus.AJOUTEE.name and
-                trip_added_in_db.status == ModificationType.delete.name)
+        if (trip_added_in_db.status == ModificationType.add.name) or \
+                (cots_trip_status == TripStatus.AJOUTEE.name and
```
pbougue: Lines 252/253, not sure what's best: should (AJOUTE + previously deleted) be a `FIRST_TIME_ADDED` ?

Navitia would not provide a deleted trip, unless there is a decorrelation...
The question is also to know what would anyone understand from `FIRST_TIME_ADDED`.
And if we allow this sequence: delete a baseVJ - add a VJ with the same headsign (can be the exact same)... I think allowing or forbidding is OK.

I would vote to keep it `PREVIOUSLY_ADDED` :neutral_face: but not sure.

kirin/abstract_sncf_model_maker.py
```diff
@@ -51,6 +51,12 @@ class TripStatus(Enum):
     PERTURBEE = 3   # Modified or Impacted
 
 
+class ActionOnTrip(Enum):
+    NOT_ADDED = 1,
+    FIRST_TIME_ADDED = 2,
```
pbougue: Depending on the choice made on the other comment:
Please comment here what's the value after a sequence: trip add - delete - then the new add
It might be ambiguous if someone reuses it for something else.

Also, I think we should add a test to have this comment respected in the future.

tests/integration/cots_model_maker_test.py
```diff
@@ -136,3 +138,61 @@ def test_cots_train_trip_removal(mock_navitia_fixture):
...
+        json_data = json.loads(input_update_added_trip)
+        dict_version = model_maker.get_value(json_data, 'nouvelleVersion')
+        train_numbers = model_maker.get_value(dict_version, 'numeroCourse')
+        pdps = model_maker._retrieve_interesting_pdp(model_maker.get_value(dict_version, 'listePointDeParcours'))
+
+        action_on_trip = model_maker._get_action_on_trip(train_numbers, dict_version, pdps)
+        assert action_on_trip == ActionOnTrip.PREVIOUSLY_ADDED.name
+
+        # Clean database for further test
+        tables = [str(table) for table in db.metadata.sorted_tables]
```
pbougue: I don't understand this cleanup though, couldn't you just pass `cots_train_151515_deleted_trip_with_delay_and_stop_time_added` directly?

kadhikari: We can't do that as explained as below:

- When we use api_post, we add a line in the table real_time_update but update lines in other tables if it's a flux-cots on a trip already treated(added, modified..).
- In the test we use only some functions for test purpose and add lines in the table real_time_update as well as as other tables. Instead of having two tests i emptied all the tables.

pbougue: OK, so the previous test might also need a cleanup in case we forbid TripUpdate that are starting "close" instead of the precise date.

PR #241, opened 2019-05-13
Add test on delay on first station pass midnight

kadhikari:
https://jira.kisio.org/browse/NAVP-1133

### DIFF
--- a/tests/integration/cots_test.py
+++ b/tests/integration/cots_test.py
@@ -1362,7 +1362,9 @@ def test_cots_add_trip_existing_in_navitia():
 
 def test_cots_update_trip_with_delay_pass_midnight_on_first_station():
     """
-    A trip with delay on first station having pass midnight towards the next day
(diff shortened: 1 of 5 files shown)


tests/integration/cots_test.py
```diff
@@ -1358,3 +1358,40 @@ def test_cots_add_trip_existing_in_navitia():
         assert len(RealTimeUpdate.query.all()) == 1
         assert len(TripUpdate.query.all()) == 0
         assert len(StopTimeUpdate.query.all()) == 0
+
+
+def test_cots_update_trip_with_delay_pass_midnight_on_first_station():
+    """
+    A trip with delay on first station having pass midnight towards the next day
```
pbougue:
```suggestion
    Disruption on a base-schedule VJ that was past-midnight.
    The disruption changes the first station and delays it so that the realtime VJ is not past-midnight anymore.
    Testing that the realtime VJ is circulating on the right day (so one day after the base-schedule VJ).
```

PR #242, opened 2019-05-16
kirin: manage pass midnight with test

kadhikari:
https://jira.kisio.org/browse/NAVP-1128

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -368,6 +368,13 @@ class TripUpdate(db.Model, TimestampMixin):
         return next((st for st in self.stop_time_updates
                      if st.stop_id == stop_id), None)
 
+    def find_first_departure(self):
+        first = next((st for st in self.stop_time_updates
+                      if st.order == 0), None)
+        if first and first.departure:
+            return first.departure.date()
+        return None
+
 
 class RealTimeUpdate(db.Model, TimestampMixin):
     """
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -368,7 +368,7 @@ class TripUpdate(db.Model, TimestampMixin):
         return next((st for st in self.stop_time_updates
                      if st.stop_id == stop_id), None)
 
-    def find_first_departure(self):
+    def find_first_departure_date(self):
         max_date = datetime.date(3000, 12, 11)
         min_date = max_date
         for st in self.stop_time_updates:
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -369,10 +369,14 @@ class TripUpdate(db.Model, TimestampMixin):
                      if st.stop_id == stop_id), None)
 
     def find_first_departure(self):
-        first = next((st for st in self.stop_time_updates
-                      if st.order == 0), None)
-        if first and first.departure:
-            return first.departure.date()
+        max_date = datetime.date(3000, 12, 11)
+        min_date = max_date
+        for st in self.stop_time_updates:
+            if st.departure:
+                min_date = min(st.departure.date(), min_date)
+
+        if min_date < max_date:
+            return min_date
         return None
 
 
--- a/kirin/core/populate_pb.py
+++ b/kirin/core/populate_pb.py
@@ -138,8 +138,9 @@ def fill_trip_update(pb_trip_update, trip_update):
     vj = trip_update.vj
     if vj:
         pb_trip.trip_id = vj.navitia_trip_id
-        # Here we use the departure datetime of first stop_time
-        circulation_date = trip_update.find_first_departure()
+        # WARNING: here trip.start_date is considered UTC, not local
+        # We use the departure date of first stop_time
+        circulation_date = trip_update.find_first_departure_date()
         if not circulation_date:
             circulation_date = vj.get_utc_circulation_date()
         pb_trip.start_date = date_to_str(circulation_date)
--- a/kirin/core/populate_pb.py
+++ b/kirin/core/populate_pb.py
@@ -69,6 +69,7 @@ def get_st_event(st_status):
         # 'update' or 'none' are modeled as 'SCHEDULED'
         return gtfs_realtime_pb2.TripUpdate.StopTimeUpdate.SCHEDULED
 
+
 def get_trip_event(trip_status):
     trip_events = {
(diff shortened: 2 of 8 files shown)


tests/integration/cots_test.py
```diff
@@ -1397,3 +1397,39 @@ def test_cots_update_trip_with_delay_pass_midnight_on_first_station():
         assert stop_times[3].arrival == datetime(2012, 11, 20, 1, 15)
         assert stop_times[3].departure_status == 'none'
         assert stop_times[3].departure == datetime(2012, 11, 20, 1, 15)
+
+
+def test_cots_update_trip_with_advance_pass_midnight_on_first_station():
```
pbougue: I'd name that test:
```suggestion
def test_cots_add_first_stop_with_advance_pass_midnight():
```

tests/integration/cots_test.py
```diff
@@ -1397,3 +1397,39 @@ def test_cots_update_trip_with_delay_pass_midnight_on_first_station():
...
+    """
+    Vehicle_journey with first stop_time at 14:01 and last stop_time at 21:46 for 20121120
+    Disruption on a base-schedule VJ without pass-midnight.
+    The disruption changes (detour) the first station at 21:30 and advances it so that the realtime VJ is
```
pbougue:
```suggestion
    The disruption adds a previous stop before existing stops at 21:30 the day before so that the realtime VJ is
```

tests/integration/cots_test.py
```diff
@@ -1397,3 +1397,39 @@ def test_cots_update_trip_with_delay_pass_midnight_on_first_station():
...
+    Test that the realtime VJ is circulating on the right day: 20121119 (so one day before the base-schedule VJ).
+    """
+    cots_update_file = get_fixture_data('cots_train_9580_pass_midnight_in_advance_on_first_station.json')
+    res = api_post('/cots', data=cots_update_file)
+    assert res == 'OK'
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 1
+        trips = TripUpdate.query.all()
+        assert len(trips) == 1
+        assert trips[0].status == 'update'
```
pbougue: `trip.status` is deprecated, no? If so, that check can be removed.
At least we should also check `trip.effect`.

kirin/core/model.py
```diff
@@ -368,6 +368,13 @@ def find_stop(self, stop_id, order=None):
         return next((st for st in self.stop_time_updates
                      if st.stop_id == stop_id), None)
 
+    def find_first_departure(self):
+        first = next((st for st in self.stop_time_updates
+                      if st.order == 0), None)
```
pbougue: I think we should take min st on `departure` (if stop order is messed up, it will be handled/rejected in consistency check) but looking for order==0 feels hazardous in case feeds are partial or numbered weirdly like GTFS-RT allows, no?
As we do a min afterward with base-VJ it should be safe if the RT feed is partial.

kirin/core/populate_pb.py
```diff
@@ -136,9 +138,12 @@ def fill_trip_update(pb_trip_update, trip_update):
     vj = trip_update.vj
     if vj:
         pb_trip.trip_id = vj.navitia_trip_id
-        # WARNING: here trip.start_date is considered UTC, not local
-        # (this date differs if vj starts during the period between midnight UTC and local midnight)
-        pb_trip.start_date = date_to_str(vj.get_utc_circulation_date())
+        # Here we use the departure datetime of first stop_time
+        circulation_date = trip_update.find_first_departure()
```
pbougue: Not sure about that at all: To me this field is supposed to be the one of the base-schedule VJ (except if it is a trip-add of course) : https://developers.google.com/transit/gtfs-realtime/reference/#message-tripdescriptor > This identifies "an instance of a **GTFS** trip".

We need to check what Kraken does with that field and if Kraken cancels the base-VJ on the good day. If this field is not used properly in Kraken, we will probably need to fix it there then.

kirin/core/populate_pb.py
```diff
@@ -136,9 +138,12 @@ def fill_trip_update(pb_trip_update, trip_update):
     vj = trip_update.vj
     if vj:
         pb_trip.trip_id = vj.navitia_trip_id
-        # WARNING: here trip.start_date is considered UTC, not local
-        # (this date differs if vj starts during the period between midnight UTC and local midnight)
-        pb_trip.start_date = date_to_str(vj.get_utc_circulation_date())
+        # Here we use the departure datetime of first stop_time
```
pbougue: At least we have to keep that UTC warning.
Also, it's the departure `date` (not datetime).

kirin/core/model.py
```diff
@@ -368,6 +368,17 @@ def find_stop(self, stop_id, order=None):
         return next((st for st in self.stop_time_updates
                      if st.stop_id == stop_id), None)
 
+    def find_first_departure(self):
+        max_date = datetime.date(3000, 12, 11)
+        min_date = max_date
+        for st in self.stop_time_updates:
+            if st.departure:
+                min_date = min(st.departure.date(), min_date)
```
pbougue: minor optim: you can as well call `.date()` at the end
or outside the function (if we ever want to use it to find the datetime, not just the date).

kirin/core/model.py
```diff
@@ -368,6 +368,17 @@ def find_stop(self, stop_id, order=None):
         return next((st for st in self.stop_time_updates
                      if st.stop_id == stop_id), None)
 
+    def find_first_departure(self):
```
pbougue:
```suggestion
    def find_first_departure_date(self):
```
or "[...]_date`time`" depending on the minor comment below.

pbougue:
Jenkins, test this please.

kirin/core/model.py
kadhikari: As discussed before, 
1. when we use this function in handler, stop_time.departure is a datetime in UTC with extension '+00' 
2. where as when in populate_pb stop_time.departure is a datetime in UTC without extension
3. We cannot comare two datetime values with and without extension where as two date values can be compared.
So here to be safe, datatime is converted to date and then compared.

nonifier:
This needs a huge change in Kraken, we want to have a test in Kirin that demonstrates pass-midnight in advance that will not crash but log an message to keep track of what's happening.
https://jira.kisio.org/browse/NAVP-1128

PR #245, opened 2019-06-17
A test of trip update on a missed added trip

kadhikari:
https://jira.kisio.org/browse/NAVP-1185
Now Kirin adds error message on all RTU if they don't send new information to Kraken

### DIFF
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -185,6 +185,17 @@ def handle(real_time_update, trip_updates, contributor, is_new_complete=False):
     data_time = datetime.datetime.utcfromtimestamp(feed.header.timestamp)
     log_dict = {'contributor': contributor, 'timestamp': data_time, 'trip_update_count': len(feed.entity),
                 'size': len(feed_str)}
+    # After merging trip_updates information of connector realtime, navitia and kirin database, if there is no new
+    # information destined to navitia, update real_time_update with status = 'KO' and a proper error message.
+    if not real_time_update.trip_updates and real_time_update.status == 'OK':
+        real_time_update.status = 'KO'
+        real_time_update.error = 'No new information destinated to navitia for this {}'.format(
+            real_time_update.connector)
+        logging.getLogger(__name__).error('No new information destinated to navitia for this {}'.format(
+            real_time_update.connector))
+        model.db.session.add(real_time_update)
+        model.db.session.commit()
+
     return real_time_update, log_dict
 
 
--- a/kirin/core/handler.py
+++ b/kirin/core/handler.py
@@ -191,8 +191,8 @@ def handle(real_time_update, trip_updates, contributor, is_new_complete=False):
         real_time_update.status = 'KO'
         real_time_update.error = 'No new information destinated to navitia for this {}'.format(
             real_time_update.connector)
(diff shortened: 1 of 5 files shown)


kirin/core/handler.py
```diff
@@ -185,6 +185,17 @@ def handle(real_time_update, trip_updates, contributor, is_new_complete=False):
     data_time = datetime.datetime.utcfromtimestamp(feed.header.timestamp)
     log_dict = {'contributor': contributor, 'timestamp': data_time, 'trip_update_count': len(feed.entity),
                 'size': len(feed_str)}
+    # After merging trip_updates information of connector realtime, navitia and kirin database, if there is no new
+    # information destined to navitia, update real_time_update with status = 'KO' and a proper error message.
+    if not real_time_update.trip_updates and real_time_update.status == 'OK':
+        real_time_update.status = 'KO'
+        real_time_update.error = 'No new information destinated to navitia for this {}'.format(
+            real_time_update.connector)
+        logging.getLogger(__name__).error('No new information destinated to navitia for this {}'.format(
```
pbougue: You can add the `RTU.id` to the trace to ease debug, and use real_time_update.error here (or store it in a temp var)
```suggestion:0+1
        logging.getLogger(__name__).error('RealTimeUpdate id={}: {}'.format(real_time_update.id, real_time_update.error))
```

tests/integration/cots_test.py
```diff
@@ -1360,6 +1360,21 @@ def test_cots_add_trip_existing_in_navitia():
         assert len(StopTimeUpdate.query.all()) == 0
 
 
+def test_cots_on_add_trip_without_first_cots():
+    """
+    Test that trip update on a newly added trip without first flux cots is rejected with a message
+    """
+    cots_add_file = get_fixture_data('cots_train_151515_added_trip_with_delay.json')
+    res = api_post('/cots', data=cots_add_file)
+    assert res == 'OK'
```
pbougue: So the response is OK, but the status KO?
To improve that we probably need to raise, not sure.

tests/integration/test_end_point.py
```diff
@@ -108,8 +108,7 @@ def setup_database():
         model.db.session.add(rtu3)
 
         rtu4 = model.RealTimeUpdate(None, connector='gtfs-rt', contributor='realtime.sherbrooke', status='KO',
-                                    error='No new information destinated to navitia for this gtfs-rt with '
-                                          'timestamp: 1520351246')
+                                    error='No new information destinated to navitia for this gtfs-rt')
```
pbougue: This change is necessary, of is it just for consistency of codebase?

kadhikari: The change just for consistency of code.

tests/integration/cots_test.py
kadhikari: As you said we need to raise an exception to have response status 'KO' and since it's used for gtfs-rt i think it's better as it is

tests/integration/test_end_point.py
pbougue: minor minor: you can re-change it if you want now that the log changed a bit.
not mandatory at all :beach_umbrella:

PR #246, opened 2019-06-18
kirin: connector IRE removed

kadhikari:
https://jira.kisio.org/browse/NAVP-1341


tests/integration/test_end_point.py
```diff
@@ -37,12 +37,6 @@
 import pytest
 
 
-def test_end_point():
```
mbounabi: Shouldn't there be a test with 'status' and 'cots' in the response?

kadhikari: it's my mistake, i will add it

kirin/default_settings.py
```diff
@@ -21,7 +21,7 @@
 
 NAVITIA_TOKEN = os.getenv('KIRIN_NAVITIA_TOKEN', None)
 
-CONTRIBUTOR = os.getenv('KIRIN_CONTRIBUTOR', 'realtime.ire')
+CONTRIBUTOR = os.getenv('KIRIN_CONTRIBUTOR', 'realtime.cots')
```
pbougue: COTS_CONTRIBUTOR already manages cots ;)
We must remove that var (from everywhere, including fab & conf).

Please, can you also check (if you didn't already) that there are no other IRE-specific params ?
(for example, I think `KIRIN_NAVITIA_INSTANCE` is still used by COTS, to be confirmed. If not it should disappear.)

kirin/resources.py
```diff
@@ -39,7 +39,7 @@ class Index(Resource):
     def get(self):
         response = {
             'status': {'href': url_for('status', _external=True)},
-            'ire': {'href': url_for('ire', _external=True)},
+            'gtfs_rt': {'href': url_for('gtfs_rt', _external=True)},
```
pbougue: I don't think we want to expose that, it's for debug purpose only as gtfs-rt is polled, not posted.

tests/gtfs_rt_test.py
```diff
@@ -43,7 +43,7 @@ def unix_time(y, m, d, h, min, s):
     return calendar.timegm(d.utctimetuple())
 
 def make_96231_20150728_0():
-    """gtfs rt feed corresponding to train_96231_delayed.xml"""
+    """gtfs rt feed corresponding to cots_train_96231_delayed.json"""
```
mbounabi: No changes needed in the rest of the test? 'train_96231_delayed.xml for IRE was exactly the same as 'cots_train_96231_delayed.json' for COTS?

tests/integration/handle_test.py
```diff
@@ -77,7 +77,7 @@ def setup_database():
                                      {'id': 'sa:3', 'departure': None, 'arrival': _dt("10:05"),
                                       'departure_status': 'none', 'arrival_status': 'none'},
                                  ])
-        rtu = RealTimeUpdate(None, 'ire', contributor='realtime.ire')
+        rtu = RealTimeUpdate(None, 'cots', contributor='realtime.cots')
```
mbounabi: General comment about the file:
Does that mean that previously, there weren't any tests COTS for every test cases in this file ? Or does the creation of the RealTimeUpdate doesn't affect the test?

tests/gtfs_rt_test.py
pbougue: As far as I remember, we designed cots-tests (fixtures) to use the same python test-methods than ire-tests.
So it should be alright yes.

tests/integration/handle_test.py
pbougue: `handle()` method doesn't really use the content of the RTU, just the related trip_updates.
So the contributor doesn't affect those tests.

mbounabi: > BTW, the signature of handle does look a bit redundant (RTU + TU...)
Is it supposed to be changed in the refacto? If not, worth added to the ticket

pbougue: Added to the last refactor ticket : https://jira.kisio.org/browse/NAVP-1353

PR #252, opened 2019-08-09
Use unicode literals everywhere

### LINKED ISSUE #176: Switch code base to unicode_literals (prepare python3)
1. This issue aims to prepare for an upgrade to python3, using
`from __future__ import absolute_import, print_function, unicode_literals, division`.
The switch to unicode in particular needs to be done all-in-one.

2. An other goal of this issue is to solve a warning when launching test:
```
tests/integration/ire_test.py::test_ire_trip_removal_simple_post
  /home/pbougue/.venv/kirin_old/local/lib/python2.7/site-packages/sqlalchemy/sql/type_api.py:279: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
    return x == y
```
On debug mode with a breakpoint, it looks like it's on 'raw_data' field from RealTimeUpdate, a str is compared to a unicode (not equal)
2 places at least are involved in this :
* https://github.com/CanalTP/kirin/blob/master/kirin/abstract_sncf_resource.py#L54 > unicode input is encoded to str 
* https://github.com/CanalTP/kirin/blob/master/kirin/ire/model_maker.py#L115 > raw_data is used (xml parser requires str)

The ORM is also probably involved at some point

Also tracked in JIRA: https://jira.kisio.org/browse/NAVP-1178

pbougue:
JIRA: https://jira.kisio.org/browse/NAVP-1178
Fix #176 

:mag: Please review by commit it should be way easier with messages

`str()` choice can be discussed :speaking_head: https://github.com/CanalTP/kirin/pull/252/commits/1b05fcf004f4a80d42efefeb7d5272cb4a4589d6

Still, this will need to run some time on the testing dev-platform and logs be challenged before publishing a stable release including it.

Underlying types depending on code and python version:

| code | python2 | python2 + future | python3 |
|-|-|-|-|
| `b'toto'` | b | b | b |
| `u'toto'` | u | u | u |
| `'toto'` | b | u | u |
| `str('toto')` | b | b | u |
----

:hammer: hand-tested it (checking that disruptions appear in Kraken and no ERROR log is produced):  
- [x] polling Sherbrooke's GTFS-RT
- [x] POSTing 1 COTS fixture from artemis guichet-unique


kirin/abstract_sncf_model_maker.py
```diff
@@ -75,17 +75,17 @@ def headsigns(str_headsign):
     The parity is the number after the '/', it gives an alternative train number.
 
     >>> headsigns('2038')
-    ['2038']
+    [u'2038']
```
nonifier: What's your rational behind using `u'xxx'`instead of `str('xxx')`

pbougue: With unicode_literals everywhere, the value returned is unicode (and will be that way in python3 of course).
In python *2* `str` is an alias for `bytestring`
In python *3* `str` is an alias for `unicode`
And doctests don't follow unicode_literals import.

That said, on this line if I use `[str('2038')]` then I'll actually test in python 2 that the output is `[b'2038']` which is gonna fail.

However I could do something like this (but don't know if it's worthy):
```python
    >>> def stringify(array):
    ...     return [str(h) for h in array]
    >>> stringify(headsigns('2038'))
    ['2038']
```

nonifier: I see, sounds fair :)

kirin/utils.py
```diff
@@ -128,7 +129,7 @@ def should_retry_exception(exception):
 def make_kirin_lock_name(*args):
     from kirin import app
 
-    return "|".join([app.config["TASK_LOCK_PREFIX"]] + [str(a) for a in args])
+    return "|".join([app.config[str("TASK_LOCK_PREFIX")]] + [str(a) for a in args])
```
kinnou02:
```suggestion
    return "|".join([app.config[str("TASK_LOCK_PREFIX")]] + [six.text_type(a) for a in args])
```
there is probably others `str()` that need to be changed

pbougue:
After review from @kinnou02 added 2 new commits.
The idea is to generate unicode as much as possible to go along with literals (so remove `str()` that were in the code before this PR).

On the last commit, I chose to fix raw_data field in db to binary_type, it's debatable.

:heavy_check_mark: Re-tested locally:
* with COTS and GTFS-RT
* with a broken motif-server to generate errors
* requesting /status also (worked fine).

PR #253, opened 2019-08-13
Add new contributor table to the db model

nonifier:
To enable multiple GTFS-RT feeds, we want to upgrade the db model with a new table that will record the new contributors.

The new table `contributor` is defined as follow:

column | type | option
--------- | ----- | -------------- 
"id" | sa.Text() (eg. realtime.sherbrook) | primary_key
"name"| sa.Text() | 
"coverage" | sa.Text() |
"token" | sa.Text() | nullable
"feed_url" | sa.Text() | nullable
"connector_type" | Enum={cots, gtfs-rt} |

The 2 tables `real_time_update` and `trip_update` now take a new foreign key on `contributor.id`. This foreign key is nullable to accommodate the migration.

https://jira.kisio.org/browse/NAVP-1372

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -312,7 +312,6 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
     message = db.Column(db.Text, nullable=True)
     contributor = db.Column(db.Text, nullable=True)
     db.Index("contributor_idx", contributor)
-    contributor_id = db.Column("contributor_id", postgresql.UUID, db.ForeignKey("contributor.id"))
     stop_time_updates = db.relationship(
         "StopTimeUpdate",
         backref="trip_update",
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -520,14 +520,14 @@ class Contributor(db.Model):  # type: ignore
     """
 
     id = db.Column(db.Text, nullable=False, primary_key=True)
-    coverage = db.Column(db.Text, nullable=False)
-    token = db.Column(db.Text, nullable=True)
+    navitia_coverage = db.Column(db.Text, nullable=False)
+    navitia_token = db.Column(db.Text, nullable=True)
     feed_url = db.Column(db.Text, nullable=True)
     connector_type = db.Column(Db_ConnectorType, nullable=False)
 
-    def __init__(self, id, coverage, connector_type, token=None, feed_url=None):
+    def __init__(self, id, navitia_coverage, connector_type, navitia_token=None, feed_url=None):
         self.id = id
-        self.coverage = coverage
+        self.navitia_coverage = navitia_coverage
         self.connector_type = connector_type
-        self.token = token
+        self.navitia_token = navitia_token
         self.feed_url = feed_url
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -79,6 +79,7 @@ class TimestampMixin(object):
 
 Db_TripEffect = db.Enum(*[e.name for e in TripEffect], name="trip_effect")
 Db_ModificationType = db.Enum(*[t.name for t in ModificationType], name="modification_type")
+Db_connector_type = db.Enum("cots", "gtfs-rt", name="connector_type", metadata=meta)
 
 
 def get_utc_timezoned_timestamp_safe(timestamp):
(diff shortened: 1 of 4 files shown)


kirin/core/model.py
```diff
@@ -78,6 +78,7 @@ class TimestampMixin(object):
 
 Db_TripEffect = db.Enum(*[e.name for e in TripEffect], name="trip_effect")
 Db_ModificationType = db.Enum(*[t.name for t in ModificationType], name="modification_type")
+Db_connector_type = db.Enum("cots", "gtfs-rt", name="connector_type", metadata=meta)
```
mbounabi: Why different casing than the enums above?
```suggestion
Db_ConnectorType = db.Enum("cots", "gtfs-rt", name="connector_type", metadata=meta)
```

migrations/versions/53647ffeb3c6_add_contributor.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+# revision identifiers, used by Alembic.
+revision = "53647ffeb3c6"
+down_revision = "4db0c66c3caf"
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+from kirin.core import model
```
kinnou02: You have to redefine the enum in this file. Migration should be constant so you can't import kirin code directly.

migrations/versions/53647ffeb3c6_add_contributor.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def upgrade():
+    op.create_table(
+        "contributor",
+        sa.Column("id", postgresql.UUID(), nullable=False),
```
kinnou02: `id` could also be the contributor code (`realtime.blabla`)

nonifier: I've removed `name` and change `id`'s type to `Text()`

kirin/core/types.py
```diff
@@ -72,6 +72,16 @@ class TripEffect(Enum):
     STOP_MOVED = 9
 
 
+class ConnectorType(Enum):
+    """
+    Represent the type of data provided by a connector.
```
mbounabi: Minor:
```suggestion
    Represents the type of data provided by a connector.
```

kirin/core/types.py
```diff
@@ -72,6 +72,16 @@ class TripEffect(Enum):
...
+
```
mbounabi: Minorer ;) why the blank line?

kirin/core/model.py
```diff
@@ -311,6 +312,7 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
     message = db.Column(db.Text, nullable=True)
     contributor = db.Column(db.Text, nullable=True)
     db.Index("contributor_idx", contributor)
+    contributor_id = db.Column("contributor_id", postgresql.UUID, db.ForeignKey("contributor.id"))
```
mbounabi: change type of contributor_id, which is "db.Text"
Side quest: add contributor_id to the class "RealTimeUpdate" too

kirin/core/model.py
```diff
@@ -510,3 +512,23 @@ def get_last_rtu(cls, connector, contributor):
         q = cls.query.filter_by(connector=connector, contributor=contributor)
         q = q.order_by(desc(cls.created_at))
         return q.first()
+
+
+class Contributor(db.Model):  # type: ignore
+    """
+    Contributor models a feeder for a specific coverage.
+    It's ID refers to its Kraken's name (eg. 'realtime.bla')
```
mbounabi:
```suggestion
    Its ID refers to its Kraken's name (eg. 'realtime.bla')
```

kirin/core/types.py
nonifier: This could be either read as : `it represents ...` or `to represent ...`
Tthe first form is used in [cpp](https://en.cppreference.com/w/cpp/algorithm) while the second in [python](https://docs.python.org/3/library/functions.html#abs).
The second form seems to be prevalent throughout Kirin. That's why I stick to it.

kirin/core/types.py
nonifier: Because black ! :black_flag:

kirin/core/model.py
nonifier: This no longer exists as there is already a `contributor = db.Column(db.Text, nullable=True)` in `TripUpdate` and `RealTimeUpdate`. 

Once the first step of the migration is done, we'll be able to create a foreign key based on this `contributor`.

kirin/core/model.py
```diff
@@ -510,3 +511,23 @@ def get_last_rtu(cls, connector, contributor):
...
+
+class Contributor(db.Model):  # type: ignore
+    """
+    Contributor models a feeder for a specific coverage.
+    Its ID refers to its Kraken's name (eg. 'realtime.bla')
+    """
+
+    id = db.Column(db.Text, nullable=False, primary_key=True)
+    coverage = db.Column(db.Text, nullable=False)
+    token = db.Column(db.Text, nullable=True)
```
pbougue: minor: I'd name it `navitia_token`

nonifier: :heavy_check_mark:

PR #254, opened 2019-08-14
Full naive datetime

pbougue:
JIRA: https://jira.kisio.org/browse/NAVP-1344

The goal of this PR is to have **naive** UTC datetime (timezone-unaware) everywhere in the code.
That will help to have consistent types everywhere, especially when using db (that doesn't store/provide timezone-aware datetimes). Then we will always manipulate the same kind of datetime, being able to compare them, etc.

So remove the datetime right from the start once they are converted to UTC in connectors.
Other interface that needs timezone in when requesting navitia (so TZ is back for URL generation only).

Nota: datetimes were already UTC.

TODO:
- [x] merge #252 as this PR is on top of it (then rebase this branch for clarity?)
- [x] Maybe hand-test that it's OK with COTS, GTFSRT, status ?

:mag: Please review by commit (with commit messages).

:warning: the last commit is open for removal/modification: only there to secure any following dev.

### DIFF
--- a/kirin/abstract_sncf_model_maker.py
+++ b/kirin/abstract_sncf_model_maker.py
@@ -116,25 +116,27 @@ class AbstractSNCFKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
         self.contributor = contributor
 
     def _get_navitia_vjs(
-        self, headsign_str, utc_since_dt, utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name
+        self, headsign_str, naive_utc_since_dt, naive_utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name
     ):
         """
         Search for navitia's vehicle journeys with given headsigns, in the period provided
-        :param utc_since_dt: UTC datetime that starts the search period.
+        :param headsign_str: the headsigns to search for (can be multiple expressed in one string, like "2512/3")
+        :param naive_utc_since_dt: naive UTC datetime that starts the search period.
             Typically the supposed datetime of first base-schedule stop_time.
-        :param utc_until_dt: UTC datetime that ends the search period.
+        :param naive_utc_until_dt: naive UTC datetime that ends the search period.
             Typically the supposed datetime of last base-schedule stop_time.
+        :param action_on_trip: action to be performed on trip. This param is used to do consistency check
         """
         log = logging.getLogger(__name__)
 
-        if (utc_since_dt is None) or (utc_until_dt is None):
+        if (naive_utc_since_dt is None) or (naive_utc_until_dt is None):
             return []
 
         vjs = {}
         # to get the date of the vj we use the start/end of the vj + some tolerance
         # since the SNCF data and navitia data might not be synchronized
-        extended_since_dt = utc_since_dt - SNCF_SEARCH_MARGIN
-        extended_until_dt = utc_until_dt + SNCF_SEARCH_MARGIN
+        extended_since_dt = naive_utc_since_dt - SNCF_SEARCH_MARGIN
+        extended_until_dt = naive_utc_until_dt + SNCF_SEARCH_MARGIN
 
         # using a set to deduplicate
         # one headsign_str (ex: "96320/1") can lead to multiple headsigns (ex: ["96320", "96321"])
@@ -159,6 +161,7 @@ class AbstractSNCFKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
                 }
             )
 
+            # Consistency check on action applied to trip
             if action_on_trip == ActionOnTrip.NOT_ADDED.name:
                 if not navitia_vjs:
                     logging.getLogger(__name__).info(
@@ -180,7 +183,7 @@ class AbstractSNCFKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
 
                 try:
                     vj = model.VehicleJourney(
-                        nav_vj, extended_since_dt, extended_until_dt, vj_start_dt=utc_since_dt
+                        nav_vj, extended_since_dt, extended_until_dt, naive_vj_start_dt=naive_utc_since_dt
                     )
                     vjs[nav_vj["id"]] = vj
                 except Exception as e:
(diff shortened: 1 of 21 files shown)


kirin/abstract_sncf_model_maker.py
```diff
@@ -123,25 +116,30 @@ def __init__(self, nav, contributor=None):
         self.contributor = contributor
 
     def _get_navitia_vjs(
-        self, headsign_str, utc_since_dt, utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name
+        self, headsign_str, naive_utc_since_dt, naive_utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name
```
kinnou02: why renaming it? it's a lot of change for a small gain, docstring has been updated and check are in place

kirin/abstract_sncf_model_maker.py
```diff
@@ -123,25 +116,30 @@ def __init__(self, nav, contributor=None):
...
             Typically the supposed datetime of last base-schedule stop_time.
+        :param action_on_trip: action to be performed on trip. This param is used to do consistency check
         """
         log = logging.getLogger(__name__)
 
-        if (utc_since_dt is None) or (utc_until_dt is None):
+        if (naive_utc_since_dt is None) or (naive_utc_until_dt is None):
             return []
 
+        assert naive_utc_since_dt.tzinfo is None
```
kinnou02: not a fan of `assert` in production code, you should use our own exceptions

kirin/utils.py
```diff
@@ -83,6 +84,15 @@ def make_navitia_wrapper():
     return navitia_wrapper.Navitia(url=url, token=token).instance(instance)
 
 
+def to_navitia_utc_str(naive_utc_dt):
+    """
+    format a naive UTC datetime to a navitia-readable UTC-aware str
+    (to avoid managing coverage's timezone)
+    """
+    assert naive_utc_dt.tzinfo is None
+    return utc.localize(naive_utc_dt).strftime("%Y%m%dT%H%M%S%z")
```
kinnou02: why? what the gain? Keeping  `Z` as a prefix didn't change anything

kirin/abstract_sncf_model_maker.py
```diff
@@ -123,25 +116,30 @@ def __init__(self, nav, contributor=None):
...
+        self, headsign_str, naive_utc_since_dt, naive_utc_until_dt, action_on_trip=ActionOnTrip.NOT_ADDED.name
     ):
         """
         Search for navitia's vehicle journeys with given headsigns, in the period provided
-        :param utc_since_dt: UTC datetime that starts the search period.
+        :param headsign_str: the headsigns to search for (can be multiple expressed in one string, like "2512/3")
+        :param naive_utc_since_dt: naive UTC datetime that starts the search period.
             Typically the supposed datetime of first base-schedule stop_time.
-        :param utc_until_dt: UTC datetime that ends the search period.
+        :param naive_utc_until_dt: naive UTC datetime that ends the search period.
```
nonifier: I don't really get what a `naive` UTC datetime means.

From what I understand, a `naive` datetime doesn't carry any information about its timezone. Which makes it timezone unaware. 
On the other hand, UTC always has a `+00` timezone (aka. GMT with no daylight saving). 

Which should never make UTC naive by definition ?!

kirin/utils.py
nonifier: Also, if we keep the `z`, it should be a capitalised `Z` as in [ISO-86601](https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC))

kirin/abstract_sncf_model_maker.py
pbougue: You are right about naive datetime in python. So these are mainly used to store "local" datetimes (what local means is always tricky).
And yes, it's a bit quirky : using implicit (naïve) UTC datetimes everywhere.

----

Why that change:
Here, as I just added in the description of the PR: That will help to have consistent types everywhere inside Kirin, especially when using db (that doesn't store/provide timezone-aware datetimes). Then we will always manipulate the same kind of datetimes, being able to compare them, etc.

So after this PR the datetimes manipulated inside Kirin must always be naive (no information about timezone stored inside) and the datetime stored must always be a UTC datetime (only it's implicit as we don't store TZ-info).

:arrow_forward:  I plan to add a section to "Contributing" mentionning that (tracked in another ticket). Helping clarify this even more is welcome (here or in the coming PR).

----


Why that choice:
when we refined the ticket I proposed that choice instead of the alternative (storing only UTC aware datetimes) because having to restore all the TZ-info when requesting db (add tzinfo=utc) would have blotted the code more (sometimes we don't know if we have a datetime from db or not, etc.).

kirin/abstract_sncf_model_maker.py
kadhikari: What about since_dt and until_dt ?

kirin/core/model.py
```diff
@@ -112,7 +97,7 @@ class VehicleJourney(db.Model):  # type: ignore
         navitia_trip_id, start_timestamp, name="vehicle_journey_navitia_trip_id_start_timestamp_idx"
     )
 
-    def __init__(self, navitia_vj, utc_since_dt, utc_until_dt, vj_start_dt=None):
+    def __init__(self, navitia_vj, naive_utc_since_dt, naive_utc_until_dt, naive_vj_start_dt=None):
```
kadhikari: As we are using a naive UTC datatime everywhere  and if well documented, there is no use of words like naive, utc while naming parameters and variables.

kirin/utils.py
nonifier: Ok, I think I miss-read the `%z` as a `z`....  my bad!

Yet, I agree with @kinnou02, you express a UTC from a naive time with `utc.localize(naive_utc_dt)`, I don't really understand the need of a time zone offset here.

kirin/abstract_sncf_model_maker.py
pbougue: As discussed aside, another PR will be done, to shorten most of Kirin's names, removing `utc` and `naive` and documenting that it's the default everywhere.
So here we'll have `since_dt` and `until_dt`.

PR #255, opened 2019-08-27
Contributor endpoint

nonifier:
Add `/contributors` endpoint, CRUD style ! 

The endpoint has 2 ways of accessing a resource:
* `/contributors`
* `/contributors/<contributor_id>`

So far, only `C` and `R` are implemented:
- Create via `POST`:
```http
http --json POST :5000/contributors id=new_contrib navitia_coverage=co connector_type=gtfs-rt
```
```http
http --json POST :5000/contributors/other_contrib_id navitia_coverage=co connector_type=gtfs-rt
```
- Read via `GET`:
```
http GET :5000/contributors
```
```
http GET :5000/contributors/new_contrib
```

**http comes from [httpie](https://github.com/jakubroztocil/httpie)

### DIFF
--- a/kirin/resources/contributors.py
+++ b/kirin/resources/contributors.py
@@ -1,0 +1,53 @@
+# coding=utf-8
+
+# Copyright (c) 2001-2019, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from __future__ import absolute_import, print_function, unicode_literals, division
+from flask_restful import Resource, marshal_with_field, fields
+from kirin.core import model
+
+contributor_fields = {
+    "id": fields.String,
+    "coverage": fields.String,
+    "token": fields.String,
+    "feed_url": fields.String,
+    "connector_type": fields.String,
+}
+
+
+class Contributors(Resource):
+    @marshal_with_field(fields.List(fields.Nested(contributor_fields)))
+    def get(self, id=None):
+        q = model.db.session.query(model.Contributor)
+
+        if id is not None:
+            q = q.filter(model.Contributor.id == id)
+
+        return q.all()
--- a/kirin/resources/contributors.py
+++ b/kirin/resources/contributors.py
@@ -54,13 +54,16 @@ class Contributors(Resource):
         return q.all()
 
     @marshal_with(contributor_fields)
-    def post(self):
+    def post(self, id=None):
         data = flask.request.get_json()
+
+        id = id or data["id"]
         token = data.get("navitia_token", None)
         feed_url = data.get("feed_url", None)
+
         try:
             new_contrib = model.Contributor(
-                data["id"], data["navitia_coverage"], data["connector_type"], token, feed_url
+                id, data["navitia_coverage"], data["connector_type"], token, feed_url
             )
             model.db.session.add(new_contrib)
             model.db.session.commit()
--- a/kirin/resources/contributors.py
+++ b/kirin/resources/contributors.py
@@ -30,7 +30,8 @@
 # www.navitia.io
 
 from __future__ import absolute_import, print_function, unicode_literals, division
-from flask_restful import Resource, marshal_with_field, fields
+import flask
+from flask_restful import Resource, marshal_with_field, marshal_with, fields, abort
 from kirin.core import model
 
 contributor_fields = {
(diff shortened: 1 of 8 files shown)


kirin/resources/contributors.py
```diff
@@ -0,0 +1,72 @@
...
+        return q.all()
+
+    @marshal_with(contributor_fields)
+    def post(self):
+        data = flask.request.get_json()
+        token = data.get("navitia_token", None)
+        feed_url = data.get("feed_url", None)
+        try:
+            new_contrib = model.Contributor(
+                data["id"], data["navitia_coverage"], data["connector_type"], token, feed_url
```
kinnou02: It will be nice to validate the input with json-schema

kadhikari: See https://github.com/CanalTP/navitia/blob/bdae285f36ce9d48f9bd05a75d2b8f410b413629/source/tyr/tyr/resources.py#L148 for example

kirin/resources/index.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import absolute_import, print_function, unicode_literals, division
+from flask_restful import Resource, url_for
+
+
+class Index(Resource):
+    def get(self):
+        response = {
+            "status": {"href": url_for("status", _external=True)},
+            "cots": {"href": url_for("cots", _external=True)},
+            "gtfs_rt": {"href": url_for("gtfs_rt", _external=True)},
```
pbougue: This one is just for debug purpose, never to be used (and will probably move with multi-contributors).
We don't want to expose it.

kirin/resources/index.py
```diff
@@ -0,0 +1,44 @@
...
+
+from __future__ import absolute_import, print_function, unicode_literals, division
+from flask_restful import Resource, url_for
+
+
+class Index(Resource):
+    def get(self):
+        response = {
+            "status": {"href": url_for("status", _external=True)},
+            "cots": {"href": url_for("cots", _external=True)},
```
pbougue: It can stay for now, but it should be exposed only when we have a cots contributor configurated, I guess.

kirin/resources/contributors.py
```diff
@@ -0,0 +1,75 @@
...
+from __future__ import absolute_import, print_function, unicode_literals, division
+import flask
+from flask_restful import Resource, marshal_with_field, marshal_with, fields, abort
+from kirin.core import model
+
+contributor_fields = {
+    "id": fields.String,
+    "navitia_coverage": fields.String,
+    "navitia_token": fields.String,
+    "feed_url": fields.String,
```
pbougue: Will it display `None` fields?

tests/integration/contributors_test.py
```diff
@@ -0,0 +1,182 @@
...
+
+def test_get_contributor_end_point(test_client):
+    assert test_client.get("/contributors").status_code == 200
+
+
+@pytest.fixture
+def with_contributors():
+    db.session.add_all(
+        [
+            model.Contributor("realtime.sherbrook", "ca", "gtfs-rt", "my_token", "http://feed.url"),
```
pbougue: minor: to respect :fleur_de_lis:-:canada: add final sherbrook`e` :wink:

tests/integration/contributors_test.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    db.session.commit()
+
+
+def test_get_contributors(test_client, with_contributors):
+    resp = test_client.get("/contributors")
+    assert resp.status_code == 200
+
+    contribs = json.loads(resp.data)
+    assert len(contribs) == 2
```
pbougue: so the json is
```json
[
  {"id": "rt.paris"}
]
```

I think we should output "contributors" everywhere (post, get, put) like what's done in navitia (in case we add context later), to have
```json
{
  "contributors": [
    {"id": "rt.paris"}
  ]
}
```

tests/integration/contributors_test.py
```diff
@@ -0,0 +1,182 @@
...
+
+def test_post_new_partial_contributor(test_client):
+    new_contrib = {"id": "realtime.tokyo", "navitia_coverage": "jp", "connector_type": "gtfs-rt"}
+    resp = test_client.post("/contributors", json=new_contrib)
+    assert resp.status_code == 201
+
+    contrib = db.session.query(model.Contributor).filter(model.Contributor.id == "realtime.tokyo").first()
+    assert contrib.id == "realtime.tokyo"
+    assert contrib.navitia_coverage == "jp"
+    assert contrib.connector_type == "gtfs-rt"
```
pbougue: minor: more realistic to use "cots" here, then

tests/integration/contributors_test.py
```diff
@@ -0,0 +1,182 @@
...
+        json={"id": "realtime.tokyo", "navitia_coverage": "jp", "connector_type": "THIS-TYPE-DOES-NOT-EXIST"},
+    )
+    assert resp.status_code == 400
+
+
+def test_post_new_valid_contributor_with_unknown_parameter_should_work(test_client):
+    resp = test_client.post(
+        "/contributors",
+        json={
+            "UNKNOWN_PARAM": "gibberish",
```
pbougue: Good to know, but I don't know if it's a stable requirement.
Maybe we should add a comment to say that this is not mandatory in case we ever want to block this?

tests/integration/contributors_test.py
```diff
@@ -0,0 +1,182 @@
...
+    resp = test_client.post("/contributors", json=new_contrib)
+    assert resp.status_code == 201
+
+    contrib = db.session.query(model.Contributor).filter(model.Contributor.id == "realtime.tokyo").first()
+    assert contrib.id == "realtime.tokyo"
+    assert contrib.navitia_coverage == "jp"
+    assert contrib.connector_type == "gtfs-rt"
+    assert contrib.navitia_token == None
+    assert contrib.feed_url == None
+
```
pbougue: Could be nice to add a `.get` test here, especially to check if we have empty fields or not in conf (my opinion is we should see empty navitia_token, but not feed_url)

kirin/resources/contributors.py
nonifier: Unfortunately yes. If the field is `None` it will be serialized with its Json equivalent : `null`.
As in :
```json
[{
        "connector_type": "gtfs-rt",
        "feed_url": null,
        "id": "new_contrib",
        "navitia_coverage": "co",
        "navitia_token": null
}]
``` 

We do have a `display_none` in Jormun, but this is a very custom field attribute that we've added on top of serpy. So I can't even use it here :sob: 

But I don't see it been a problem here, unless I'm missing something

kirin/resources/contributors.py
nonifier: :heavy_check_mark:

kirin/resources/index.py
nonifier: it's been removed !

tests/integration/contributors_test.py
nonifier: good catch ! :heavy_check_mark:

tests/integration/contributors_test.py
nonifier: I've also updated single contributor returned by a post to be consistent with:
```json
{ 
"contributor" : { ... }
}
```

tests/integration/contributors_test.py
nonifier: If we ever want to have this as a constrain, we could use `"additional_properties" : false` in the input json schema that is now being used (see [properties](https://json-schema.org/understanding-json-schema/reference/object.html#properties))

tests/integration/contributors_test.py
nonifier: :heavy_check_mark:

PR #256, opened 2019-08-29
Rename and document naive_utc datetime params

pbougue:
Follow-up after https://github.com/CanalTP/kirin/pull/254#discussion_r318050522

:mag: Review by commit (with message) is advised to separate renames from the rest

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,8 +19,9 @@ Especially as the database does not store/provide timezone-aware datetimes.
 _Where?_  
 Everywhere.  
 So a naive-UTC-datetime is produced right after it's read,
-and timezone info is added as late as possible when needed (ex: when requesting navitia).
-
+and timezone info is added as late as possible when needed (ex: when requesting external service).  
+When absolutely needed, datetimes that are not UTC **must** be timezone-aware to avoid
+being able to mix them with naive-UTC-datetimes (it's also the easiest and safest way to handle timezones).
 
 
 ### Python formatting
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ To generate a new database revision script (after an upgrade of the model.py fil
 ```
 honcho run ./manage.py db migrate
 ```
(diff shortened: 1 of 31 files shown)


CONTRIBUTING.md
```diff
@@ -1,6 +1,29 @@
...
+
+_Why?_  
+Simplest way to always manipulate the same kind of datetimes,
+being able to compare them, no matter the source.  
+Especially as the database does not store/provide timezone-aware datetimes.
+
+_Where?_  
+Everywhere.  
+So a naive-UTC-datetime is produced right after it's read,
+and timezone info is added as late as possible when needed (ex: when requesting navitia).
```
kinnou02: You should add a paragraph that a datetime that isn't at UTC should always be aware, sometime we will need them in the connectors, but we must not be able to mix them.

CONTRIBUTING.md
```diff
@@ -15,11 +39,116 @@ But attention: it requires python 3.6+ to run.
...
 ```
+
+
+## Alembic database revisions
+
+To generate a new database revision script (after an upgrade of the model.py file):
+```
+honcho run ./manage.py db migrate
+```
+This will generate a new migration file, that you can amend to your will.
```
nonifier:
```suggestion
This will generate a new migration file, that you can amend at your will.
```

CONTRIBUTING.md
```diff
@@ -15,11 +39,116 @@ But attention: it requires python 3.6+ to run.
...
+There is only one of these on each platform.
+
+#### Kirin-worker
+
+> Alias 'worker'
+
+Its role is to poll an external location and check if new information was published.
+In that case, the worker processes it, stores the result in db and sends the corresponding info to Kraken.
+
+There can be several of these if the load is important. At least one per polled provider is recommended.
```
nonifier:
```suggestion
There can be several of these if the load is important. At least one per polled provider is mandatory.
```
Can you use Kirin without any worker ?

pbougue: Yes, if you only POST data (for example COTS) you can.

PR #257, opened 2019-08-30
Join contributor to other tables

kadhikari:
This PR joins the table contributor  to real_time_update and trip_update with column contributor_id.
- A column contributor_id is added in real_time_update and trip_update.
- Existing column contributor is no more used (will be filled as before).
- New indexes added on contributor_id.
- Tests modified.
**Note:** Elements (column, indexes and uses in codes if any) related to contributor will have to be deleted after deploying this version.

Concerned Ticket: https://jira.kisio.org/browse/NAVP-1376

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -448,7 +448,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         db.Index("realtime_update_contributor_id_and_created_at", "created_at", "contributor_id"),
     )
 
-    def __init__(self, raw_data, connector, contributor=None, status="OK", error=None, received_at=None):
+    def __init__(self, raw_data, connector, contributor, status="OK", error=None, received_at=None):
         self.id = gen_uuid()
         self.raw_data = raw_data
         self.connector = connector
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -319,7 +319,7 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
         self,
         vj=None,
         status="none",
-        contributor="realtime.cots",
+        contributor=None,
         company_id=None,
         effect=None,
         physical_mode_id=None,
@@ -328,6 +328,7 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
         self.created_at = datetime.datetime.utcnow()
         self.vj = vj
         self.status = status
+        self.contributor = contributor
         self.company_id = company_id
         self.effect = effect
         self.physical_mode_id = physical_mode_id
@@ -447,13 +448,12 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         db.Index("realtime_update_contributor_id_and_created_at", "created_at", "contributor_id"),
     )
 
-    def __init__(
-        self, raw_data, connector, contributor="realtime.cots", status="OK", error=None, received_at=None
-    ):
+    def __init__(self, raw_data, connector, contributor=None, status="OK", error=None, received_at=None):
         self.id = gen_uuid()
         self.raw_data = raw_data
         self.connector = connector
         self.status = status
+        self.contributor = contributor
         self.error = error
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
         self.contributor_id = contributor
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -312,12 +312,14 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
     effect = db.Column(Db_TripEffect, nullable=True)
     physical_mode_id = db.Column(db.Text, nullable=True)
     headsign = db.Column(db.Text, nullable=True)
+    contributor_id = db.Column(db.Text, db.ForeignKey('contributor.id'), nullable=False)
+    db.Index("contributor_id_idx", contributor_id)
 
     def __init__(
         self,
         vj=None,
         status="none",
-        contributor=None,
+        contributor="realtime.cots",
         company_id=None,
         effect=None,
         physical_mode_id=None,
@@ -326,11 +328,11 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
         self.created_at = datetime.datetime.utcnow()
         self.vj = vj
         self.status = status
-        self.contributor = contributor
         self.company_id = company_id
         self.effect = effect
         self.physical_mode_id = physical_mode_id
         self.headsign = headsign
+        self.contributor_id = contributor
 
     def __repr__(self):
         return "<TripUpdate %r>" % self.vj_id
@@ -373,7 +375,7 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
 
     @classmethod
     def find_by_contributor_period(cls, contributors, start_date=None, end_date=None):
-        query = cls.query.filter(cls.contributor.in_(contributors))
+        query = cls.query.filter(cls.contributor_id.in_(contributors))
         if start_date:
             start_dt = datetime.datetime.combine(start_date, datetime.time(0, 0))
             query = query.filter(
@@ -429,6 +431,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
     error = db.Column(db.Text, nullable=True)
     raw_data = deferred(db.Column(db.Text, nullable=True))
     contributor = db.Column(db.Text, nullable=True)
+    contributor_id = db.Column(db.Text, db.ForeignKey('contributor.id'), nullable=False)
 
     trip_updates = db.relationship(
         "TripUpdate",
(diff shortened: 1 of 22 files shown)


kirin/core/model.py
```diff
@@ -316,12 +316,14 @@ class TripUpdate(db.Model, TimestampMixin):  # type: ignore
...
     headsign = db.Column(db.Text, nullable=True)
+    contributor_id = db.Column(db.Text, db.ForeignKey("contributor.id"), nullable=False)
+    db.Index("contributor_id_idx", contributor_id)
 
     def __init__(
         self,
         vj=None,
         status="none",
-        contributor=None,
+        contributor="realtime.cots",
```
kinnou02: This seems a bit risky, the generic model should not define default value that are specific to a contributor that might even not exist

kirin/core/model.py
```diff
@@ -464,10 +470,10 @@ def get_probes_by_contributor(cls):
         from kirin import app
 
         result = {"last_update": {}, "last_valid_update": {}, "last_update_error": {}}
-        contributor = [app.config[str("COTS_CONTRIBUTOR")], app.config[str("GTFS_RT_CONTRIBUTOR")]]
-        for c in contributor:
+        contributors = [app.config[str("COTS_CONTRIBUTOR")], app.config[str("GTFS_RT_CONTRIBUTOR")]]
```
kinnou02: we should use contributors from the database here

kirin/cots/model_maker.py
```diff
@@ -378,8 +378,8 @@ def _make_trip_update(self, json_train, vj, action_on_trip=ActionOnTrip.NOT_ADDE
         create the new TripUpdate object
         Following the COTS spec: https://github.com/CanalTP/kirin/blob/master/documentation/cots_connector.md
         """
-        trip_update = model.TripUpdate(vj=vj)
-        trip_update.contributor = self.contributor
```
kinnou02: You should continue to feed the old column during the transition

tests/integration/contributors_test.py
```diff
@@ -50,6 +50,10 @@ def test_get_contributor_end_point(test_client):
 
 @pytest.fixture
 def with_contributors():
+    # Clean table contributor before adding any elements as we don't clean it in the function clean_db
+    db.session.execute("TRUNCATE table contributor CASCADE;")
```
kinnou02: why not clean it in `clean_db`?

kadhikari: We clean and insert two conributors in the function clean_db

kirin/cots/model_maker.py
kadhikari: Done

kirin/core/model.py
kadhikari: This will be done while treating https://jira.kisio.org/browse/NAVP-1374

kirin/core/model.py
kadhikari: Done

migrations/versions/443587af1780_reference_contributor_in_other_tables.py
```diff
@@ -0,0 +1,76 @@
...
+    # Drop foreignKey constraints
+    op.drop_constraint("fk_trip_update_contributor_id", "trip_update", type_="foreignkey")
+    op.drop_constraint("fk_real_time_update_contributor_id", "real_time_update", type_="foreignkey")
+
+    # Delete contributor_id in real_time_update and trip_update
+    op.drop_column("trip_update", "contributor_id")
+    op.drop_column("real_time_update", "contributor_id")
+
+    # Delete lines from the table contributor
+    op.execute("DELETE FROM contributor;")
```
nonifier: You might want to only delete the two contributors that you've created (ie. `realtime.cots' and `realtime.sherbrooke`), otherwise, deleting the entire table will have side effects.

tests/integration/conftest.py
```diff
@@ -64,10 +64,22 @@ def clean_db():
     before all tests the database is cleared
     """
     with app.app_context():
+        # The table contributor should never be emptied as referenced in real_time_update and trip_update
         tables = [six.text_type(table) for table in db.metadata.sorted_tables]
         db.session.execute("TRUNCATE {} CASCADE;".format(", ".join(tables)))
         db.session.commit()
 
+        # Add two contributors in the table
+        db.session.execute(
```
nonifier: To be more robust against changes in the model, you might want to add the objects through it:
```python
    db.session.add_all(
        [
            model.Contributor("realtime.sherbrooke", "sherbrooke", ConnectorType.gtfs_rt.value, "token_to_be_modified", "feed_url_to_be_modified"),
            model.Contributor("realtime.cots", "sncf", ConnectorType.gtfs_rt.value, "token_to_be_modified", "feed_url_to_be_modified"),
        ]
    )
    db.session.commit()

```

migrations/versions/443587af1780_reference_contributor_in_other_tables.py
kadhikari: Bien vu :smile:

tests/integration/conftest.py
kadhikari: Done

kirin/core/model.py
```diff
@@ -445,16 +449,18 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
     __table_args__ = (
         db.Index("realtime_update_created_at", "created_at"),
         db.Index("realtime_update_contributor_and_created_at", "created_at", "contributor"),
+        db.Index("realtime_update_contributor_id_and_created_at", "created_at", "contributor_id"),
     )
 
-    def __init__(self, raw_data, connector, contributor, status="OK", error=None, received_at=None):
+    def __init__(self, raw_data, connector, contributor=None, status="OK", error=None, received_at=None):
```
kinnou02: `contributor_id` can't be `NULL` in database, why put a invqlid default value?

tests/integration/cots_model_maker_test.py
```diff
@@ -174,7 +174,10 @@ def test_get_action_on_trip_add(mock_navitia_fixture):
         assert action_on_trip == ActionOnTrip.PREVIOUSLY_ADDED.name
 
         # Clean database for further test
-        tables = [six.text_type(table) for table in db.metadata.sorted_tables]
+        # The table contributor should never be emptied as referenced in real_time_update and trip_update
```
pbougue: Can't you just re-call clean_db?

tests/integration/gtfs_rt_test.py
```diff
@@ -153,17 +153,23 @@ def test_gtfs_model_builder(basic_gtfs_rt_data, basic_gtfs_rt_data_without_delay
...
+        trip_updates = gtfs_rt.KirinModelBuilder(dumb_nav_wrapper(), contributor="realtime.sherbrooke").build(
+            rt_update, basic_gtfs_rt_data
+        )
 
         # we associate the trip_update manually for sqlalchemy to make the links
         rt_update.trip_updates = trip_updates
         db.session.add(rt_update)
         db.session.commit()
 
+        assert rt_update.contributor == "realtime.sherbrooke"
```
pbougue: Not sure it's worth adding this test (same below multiple times for real_time_update and trip_update)

tests/integration/conftest.py
```diff
@@ -64,10 +64,22 @@ def clean_db():
...
+            "INSERT INTO contributor VALUES('realtime.cots','sncf',"
```
pbougue: It's a lot of changes in test files, but it would be better to avoid in tests those values that are in default config.
That way we will see if we get some hardcoded ones from real code (bad) or the ones from tests (good)
```suggestion
            "INSERT INTO contributor VALUES('rt.tchoutchou','sncf',"
```

tests/integration/conftest.py
```diff
@@ -64,10 +64,22 @@ def clean_db():
...
+            "'token_to_be_modified','feed_url_to_be_modified','cots');"
+        )
+        db.session.execute(
+            "INSERT INTO contributor VALUES('realtime.sherbrooke','sherbrooke',"
```
pbougue: Same
```suggestion
            "INSERT INTO contributor VALUES('rt.vroumvroum','sherbrooke',"
```

kirin/core/model.py
kadhikari: Done

migrations/versions/443587af1780_reference_contributor_in_other_tables.py
```diff
@@ -0,0 +1,76 @@
...
+import sqlalchemy as sa
+
+
+def upgrade():
+    # Insert information for a contributor 'realtime.cots' and 'realtime.sherbrooke'
+    # if present in the table trip_update and absent in contributor
+    op.execute(
+        "INSERT INTO contributor SELECT 'realtime.cots','sncf',"
+        "'token_to_be_modified','feed_url_to_be_modified','cots'"
+        " WHERE 'realtime.cots' in (SELECT DISTINCT contributor FROM real_time_update)"
```
pbougue: We have to do the same for TripUpdate, no?
They are not supposed to be different, but...
Maybe we shouldn't have any contributor/contributor_id in TripUpdate and get the one from associated RTU, to avoid redundancy??

migrations/versions/443587af1780_reference_contributor_in_other_tables.py
pbougue: minor:
Less side effect with your suggestion, but there might still be some unavoidable ones (if cots and sherbrooke where there already, not added by upgrade).
Maybe the safest is to let it be handled by the drop_table in previous migrations?

tests/integration/utils_sncf_test.py
```diff
@@ -103,7 +104,8 @@ def check_db_96231_delayed(motif_externe_is_null=False):
         else:
             assert second_st.message == "Affluence exceptionnelle de voyageurs"
 
-        assert db_trip_delayed.contributor == "realtime.cots"
+        assert db_trip_delayed.contributor is not None
```
pbougue: Shouldn't you just do
```python
assert db_trip_delayed.contributor_id == contributor
assert db_trip_delayed.contributor == db_trip_delayed.contributor_id
```
everywhere in that file?

So here add contributor to `check_db_96231_delayed()`'s params,
and below change a bit the tests.

PR #258, opened 2019-09-02
Contributor endpoint

nonifier:
Part II of endpoint `/contributors`

* PUT `/contributors/<id>` with a JSON data that matches the contributor's model (same as for a POST). _The only field that cannot be changed is the object's `id`_ 
 ```bash
http --json PUT :5000/contributors/<id> navitia_coverage=idf
```
 ```bash
http --json PUT :5000/contributors/<id> navitia_coverage=idf feed_url="http://feed.url" navitia_token="XXXXXX" connector_type={cots|gtfs-rt}
```
* DELETE `/contributors/<id>`
```bash
http DELETE :5000/contributors/<id> 
```

### DIFF
--- a/kirin/resources/contributors.py
+++ b/kirin/resources/contributors.py
@@ -34,6 +34,7 @@ import flask
 import jsonschema
 from flask_restful import Resource, marshal_with_field, marshal_with, fields, abort
 from kirin.core import model
+from kirin.core.types import ConnectorType
 from kirin.exceptions import ObjectNotFound
 
 contributor_fields = {
@@ -57,7 +58,7 @@ class Contributors(Resource):
             "navitia_coverage": {"type": "string"},
             "navitia_token": {"type": "string"},
             "feed_url": {"type": "string", "format": "uri"},
-            "connector_type": {"type": "string", "enum": ["cots", "gtfs-rt"]},
+            "connector_type": {"type": "string", "enum": ConnectorType.values()},
         },
         "required": ["navitia_coverage", "connector_type"],
     }
@@ -68,7 +69,7 @@ class Contributors(Resource):
             "navitia_coverage": {"type": "string"},
             "navitia_token": {"type": "string"},
             "feed_url": {"type": "string", "format": "uri"},
-            "connector_type": {"type": "string", "enum": ["cots", "gtfs-rt"]},
+            "connector_type": {"type": "string", "enum": ConnectorType.values()},
         },
     }
 
--- a/kirin/resources/contributors.py
+++ b/kirin/resources/contributors.py
@@ -52,27 +52,21 @@ contributor_nested_fields = {"contributor": fields.Nested(contributor_fields)}
 
 class Contributors(Resource):
 
+    schema_properties = {
+        "id": {"type": "string"},
+        "navitia_coverage": {"type": "string"},
+        "navitia_token": {"type": "string"},
+        "feed_url": {"type": "string", "format": "uri"},
+        "connector_type": {"type": "string", "enum": ConnectorType.values()},
+    }
+
     post_data_schema = {
         "type": "object",
-        "properties": {
-            "id": {"type": "string"},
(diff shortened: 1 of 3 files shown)


kirin/resources/contributors.py
```diff
@@ -62,12 +62,24 @@ class Contributors(Resource):
         "required": ["navitia_coverage", "connector_type"],
     }
 
+    put_data_schema = {
+        "type": "object",
+        "properties": {
+            "navitia_coverage": {"type": "string"},
+            "navitia_token": {"type": "string"},
+            "feed_url": {"type": "string", "format": "uri"},
+            "connector_type": {"type": "string", "enum": ["cots", "gtfs-rt"]},
```
kinnou02: should be extracted from one of our enums.
some field should be required too.

kirin/resources/contributors.py
```diff
@@ -99,3 +111,51 @@ def post(self, id=None):
...
         except Exception as e:
             abort(400, message="Error while creating contributor - {}".format(e))
+
+    def delete(self, id=None):
+
+        if id is None:
+            abort(400, message="Contributor's id is missing")
+
+        try:
+            contributor = model.db.session.query(model.Contributor).filter(model.Contributor.id == id)
```
kinnou02: there is some shortcut for that kind of request, something like:
```
contributor = model.Contributor.query.get(id)
```
there is also a `get_or_404()`

kirin/resources/contributors.py
```diff
@@ -99,3 +111,51 @@ def post(self, id=None):
...
+
+            if contributor.count() < 1:
+                raise ObjectNotFound("Contributor '{}' could not be found".format(id))
+
+            contributor.delete()
```
kinnou02: we probably want to disable it, not delete it from the database

kirin/resources/contributors.py
nonifier: Fields are now taken from the Enum.
 
Also, which field should be be required? I already enforce the `id` to be passed in the endpoint (like in: `PUT /contributors/<id> --data {...}`). 

Any of the field could be used, as you can update partial object. Like if you want to only update a contributor's token with: `PUT /contributors/my-contrib --data {'navitia_token': 'abcdefgh'}`

kirin/resources/contributors.py
nonifier: I'll do another PR on top of this and Krishna's changes to address this. It would be much better to test this with the foreign keys in the model.

tests/integration/contributors_test.py
```diff
@@ -203,3 +205,107 @@ def test_post_new_valid_contributor_with_unknown_parameter_should_work(test_clie
...
+
+
+def test_put_unknown_contributor(test_client, with_contributors):
+    resp = test_client.put("/contributors/SaintMeuMeu", json={"navitia_coverage": "qb"})
+    assert resp.status_code == 404
+
+
+def test_put_contributor_with_malformed_data(test_client, with_contributors):
+    resp = test_client.put("/contributors/realtime.paris", json={"feed_url": 42})
+    assert resp.status_code == 400
```
pbougue: Can you add a test to check that it's possible to put a get:
* POST (or just add in db) a new contributor
* GET the contrib as a json
* PUT the json back
* Check that nothing changed

kirin/resources/contributors.py
pbougue: To me PUT in a complete update.
PATCH would be a partial one, but it's too complicated to handle for the gain (tartare chose to drop it).
GET + PUT is a good enough way for partial update (that's why I asked to add a test on that).

So we may require `navitia_coverage` and `connector_type` that are mandatory in db (later we can require `feed_url` for a gtfs-rt connector).

tests/integration/contributors_test.py
nonifier: :heavy_check_mark: Added !

PR #260, opened 2019-09-05
Have pybreaker around all COTS message-subservice manager

pbougue:
:mag: Review by commit.

As this "motif" service can return `503` HTTP errors on long periods,
no raise is done in GET method, so it was not handled by pybreaker,
as exception was raised outside of it, by `_service_caller()` method.

Also, pybreaker had the lifetime of a request processing, not the one of Kirin.

Looks like there is no easy way to test that (if it is worthy, as message-handling is tested already).

- [x] Hand-test are OK :heavy_check_mark: 

Related from afar to https://jira.kisio.org/browse/NAVITIAII-2800

### DIFF
--- a/kirin/__init__.py
+++ b/kirin/__init__.py
@@ -31,6 +31,9 @@
 
 from __future__ import absolute_import, print_function, unicode_literals, division
 import os
+
+import pybreaker
+
 from kirin import exceptions
 from kirin.rabbitmq_handler import RabbitMQHandler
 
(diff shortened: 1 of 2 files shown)


kirin/cots/message_handler.py
```diff
@@ -99,12 +99,9 @@ def _service_caller(self, method, url, headers, data=None):
...
                     )
                 raise ObjectNotFound(
-                    "non 200 response on COTS cause message " "sub-service {} {}".format(method, url)
+                    "non 200 response on COTS cause message sub-service {} {}".format(method, url)
                 )
             return response
-        except pybreaker.CircuitBreakerError as e:
-            logging.getLogger(__name__).error("COTS cause message sub-service dead (error: {})".format(e))
-            raise SubServiceError("COTS cause message sub-service circuit breaker open")
         except requests.Timeout as t:
```
kinnou02: timeout exception should be handled by the breaker too, in fact a lot of the exception of this method should trigger the breaker

pbougue: Yep, that's why I moved it around `_call_webservice_safer` to wrap all the raises from this MessageHandler.

kirin/__init__.py
```diff
@@ -90,6 +93,11 @@
     password=app.config[str("REDIS_PASSWORD")],
 )
 
+cots_message_breaker = pybreaker.CircuitBreaker(
```
kinnou02: Why did you put it there? it will pollute the global namespace :(

PR #263, opened 2019-09-10
Handle contributors from db

mbounabi:
https://jira.kisio.org/browse/NAVP-1374

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -462,15 +462,22 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         result = {"last_update": {}, "last_valid_update": {}, "last_update_error": {}}
         # TODO :
         #  remove config from file
+        # Get contributors from both file and db (config file has priority)
+        gtfsrt_contributors = []
+        contributor_legacy = None
         if "GTFS_RT_CONTRIBUTOR" in app.config and app.config[str("GTFS_RT_CONTRIBUTOR")]:
-            gtfsrt_contributors = [app.config[str("GTFS_RT_CONTRIBUTOR")]]
-        else:
-            gtfsrt_contributors = [
+            contributor_legacy = app.config[str("GTFS_RT_CONTRIBUTOR")]
+            gtfsrt_contributors.append(contributor_legacy)
+
+        gtfsrt_contributors.extend(
+            [
                 x.id
                 for x in db.session.query(Contributor)
                 .filter(Contributor.connector_type == ConnectorType.gtfs_rt.value)
                 .all()
+                if x.id != contributor_legacy
             ]
+        )
 
         if "COTS_CONTRIBUTOR" in app.config and app.config[str("COTS_CONTRIBUTOR")]:
             cots_contributors = [app.config[str("COTS_CONTRIBUTOR")]]
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -546,4 +546,4 @@ class Contributor(db.Model):  # type: ignore
 
     @classmethod
     def find_by_connector_type(cls, type):
-        return cls.query.filter_by(connector_type=type).all()
+        return cls.query.filter_by(connector_type=type, is_active=True).all()
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -28,6 +28,7 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
 from __future__ import absolute_import, print_function, unicode_literals, division
 from datetime import timedelta
 from sqlalchemy.dialects import postgresql
@@ -459,11 +460,13 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         from kirin import app
 
         result = {"last_update": {}, "last_valid_update": {}, "last_update_error": {}}
-
-        # TODO:
-        #  read configurations from base ONLY if there is no configuration
-        #  available in config file (config file will prevail for transition).
-        contributors = [app.config[str("COTS_CONTRIBUTOR")], app.config[str("GTFS_RT_CONTRIBUTOR")]]
+        # TODO :
+        #  remove config from file
+        if "GTFS_RT_CONTRIBUTOR" in app.config and app.config[str("GTFS_RT_CONTRIBUTOR")]:
+            gtfs_contributors = [app.config[str("GTFS_RT_CONTRIBUTOR")]]
+        else:
+            gtfs_contributors = [x.id for x in db.session.query(Contributor).all()]
+        contributors = [app.config[str("COTS_CONTRIBUTOR")]] + gtfs_contributors
         for c in contributors:
             sql = db.session.query(cls.created_at, cls.status, cls.updated_at, cls.error)
             sql = sql.filter(cls.contributor_id == c)
@@ -522,3 +525,7 @@ class Contributor(db.Model):  # type: ignore
         self.connector_type = connector_type
         self.navitia_token = navitia_token
         self.feed_url = feed_url
+
+    @classmethod
+    def find_by_id(cls, id):
+        return cls.query.filter_by(id=id).one()
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -472,9 +472,7 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         gtfsrt_contributors.extend(
             [
                 x.id
-                for x in db.session.query(Contributor)
-                .filter(Contributor.connector_type == ConnectorType.gtfs_rt.value)
-                .all()
+                for x in Contributor.find_by_connector_type(ConnectorType.gtfs_rt.value)
                 if x.id != contributor_legacy
             ]
         )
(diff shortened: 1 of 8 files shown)


kirin/core/model.py
```diff
@@ -459,11 +460,13 @@ def get_probes_by_contributor(cls):
...
-        #  read configurations from base ONLY if there is no configuration
-        #  available in config file (config file will prevail for transition).
-        contributors = [app.config[str("COTS_CONTRIBUTOR")], app.config[str("GTFS_RT_CONTRIBUTOR")]]
+        # TODO :
+        #  remove config from file
+        if "GTFS_RT_CONTRIBUTOR" in app.config and app.config[str("GTFS_RT_CONTRIBUTOR")]:
+            gtfs_contributors = [app.config[str("GTFS_RT_CONTRIBUTOR")]]
+        else:
+            gtfs_contributors = [x.id for x in db.session.query(Contributor).all()]
+        contributors = [app.config[str("COTS_CONTRIBUTOR")]] + gtfs_contributors
```
kadhikari: configuration for "COTS_CONTRIBUTOR" used  from database as well as from app.config if COTS_CONTRIBUTOR is present in config file non?

kirin/tasks.py
```diff
@@ -113,17 +113,28 @@ def purge_rt_update(self, config):
...
+        contributors = model.Contributor.query.all()
+        for contributor in contributors:
+            config = {
+                "contributor": contributor.id,
+                "navitia_url": app.config.get(str("NAVITIA_URL")),
+                "token": contributor.navitia_token,
+                "coverage": contributor.navitia_coverage,
+                "feed_url": contributor.feed_url,
+            }
+            gtfs_poller.delay(config)
```
kadhikari: In the previous version (presence of config file) we create a poller only for "gtfs-rt", but your correction will create one poller per contributor present in database including "realtime.cots"

kirin/utils.py
```diff
@@ -78,12 +79,17 @@ def make_navitia_wrapper():
...
+    #  remove config from file
     url = current_app.config[str("NAVITIA_URL")]
-    token = current_app.config.get(str("NAVITIA_TOKEN"))
-    instance = current_app.config[str("NAVITIA_INSTANCE")]
+
+    if "NAVITIA_INSTANCE" in current_app.config and current_app.config[str("NAVITIA_INSTANCE")]:
+        instance = current_app.config[str("NAVITIA_INSTANCE")]
+        token = current_app.config[str("NAVITIA_TOKEN")]
+    else:
+        contributor = model.Contributor.find_by_id(current_app.config[str("COTS_CONTRIBUTOR")])
```
kadhikari: Do we have "COTS_CONTRIBUTOR" for configuration from database ?

kirin/tasks.py
```diff
@@ -113,17 +114,28 @@ def purge_rt_update(self, config):
...
+            "contributor": app.config.get(str("GTFS_RT_CONTRIBUTOR")),
+            "navitia_url": app.config.get(str("NAVITIA_URL")),
+            "token": app.config.get(str("NAVITIA_GTFS_RT_TOKEN")),
+            "coverage": app.config.get(str("NAVITIA_GTFS_RT_INSTANCE")),
+            "feed_url": app.config.get(str("GTFS_RT_FEED_URL")),
+        }
+        gtfs_poller.delay(config)
+    else:
+        contributors = model.Contributor.find_by_connector_type(ConnectorType.gtfs_rt.value)
+        for contributor in contributors:
```
kinnou02: we should exclude the contributor that match `app.config[str("GTFS_RT_CONTRIBUTOR")]`

kirin/utils.py
```diff
@@ -78,12 +80,23 @@ def make_navitia_wrapper():
...
+    #  remove config from file
     url = current_app.config[str("NAVITIA_URL")]
-    token = current_app.config.get(str("NAVITIA_TOKEN"))
-    instance = current_app.config[str("NAVITIA_INSTANCE")]
+
+    if "NAVITIA_INSTANCE" in current_app.config and current_app.config[str("NAVITIA_INSTANCE")]:
+        instance = current_app.config[str("NAVITIA_INSTANCE")]
+        token = current_app.config[str("NAVITIA_TOKEN")]
+    else:
+        contributor = model.Contributor.find_by_connector_type(ConnectorType.cots.value)
```
kinnou02: if this is only used by cots it should be moved in the cots namespace.
And this shouldn't query the database directly.

kirin/tasks.py
mbounabi: The config from db and file are exclusive. So it won't be necessary

kirin/utils.py
mbounabi: OK to move it, even if this wasn't introduced in this PR.
However, what do you mean by " this shouldn't query the database directly" ?

kirin/tasks.py
mbounabi: After discussion, both file and db are read, with the config from file prioritary.

kirin/tasks.py
```diff
@@ -113,17 +114,32 @@ def purge_rt_update(self, config):
...
-        "coverage": app.config.get(str("NAVITIA_GTFS_RT_INSTANCE")),
-        "feed_url": app.config.get(str("GTFS_RT_FEED_URL")),
-    }
-    gtfs_poller.delay(config)
+    # TODO :
+    #  remove config from file
+    contributor_legacy = None
+    if "GTFS_RT_CONTRIBUTOR" in app.config and app.config[str("GTFS_RT_CONTRIBUTOR")]:
+        contributor_legacy = app.config.get(str("GTFS_RT_CONTRIBUTOR"))
+        config = {
```
pbougue: It would be nice to have a shared list of contributor's config (or a shared way of building it) that we reuse everywhere.
That way we would separate the way the list is built (which is weird for now), from the way we use it (which is quite clean and is not gonna change much).

My idea is moving this part of the code (+cots config reading) in utils, then probably make a global variable out of that, to be reused everywhere you need in this PR (using just what you need through filter methods).
In utils, you would enforce the "only one COTS" rule and so on.

Here you would just iterate blindly over all the gtfsrt contributors, no matter if they are from config file or db.

kirin/core/model.py
```diff
@@ -459,11 +460,36 @@ def get_probes_by_contributor(cls):
...
+            contributor_legacy = app.config[str("GTFS_RT_CONTRIBUTOR")]
+            gtfsrt_contributors.append(contributor_legacy)
+
+        gtfsrt_contributors.extend(
+            [
+                x.id
+                for x in db.session.query(Contributor)
+                .filter(Contributor.connector_type == ConnectorType.gtfs_rt.value)
+                .all()
+                if x.id != contributor_legacy
```
pbougue: Same here, if you move out the contributor's config list build, you avoid those specific considerations in this code (they would be mutualized in utils).
And also you would make no test/filter about the type of contributor here.

kirin/core/model.py
```diff
@@ -459,11 +460,36 @@ def get_probes_by_contributor(cls):
...
-        #  read configurations from base ONLY if there is no configuration
-        #  available in config file (config file will prevail for transition).
-        contributors = [app.config[str("COTS_CONTRIBUTOR")], app.config[str("GTFS_RT_CONTRIBUTOR")]]
+        if "COTS_CONTRIBUTOR" in app.config and app.config[str("COTS_CONTRIBUTOR")]:
+            cots_contributors = [app.config[str("COTS_CONTRIBUTOR")]]
+        else:
+            cots_contributors = [
+                x.id
+                for x in db.session.query(Contributor)
+                .filter(Contributor.connector_type == ConnectorType.cots.value)
```
pbougue: If you keep that code, you can probably use the find_by_connector_type() method, no?

mbounabi:
General note: I think COTS shouldn't have been handled in this PR, as it is out of the JIRA ticket scope (only about GTFS-RT). To be discussed if it's too late and we handle everything here or create an other PR for COTS config from db...

kirin/tasks.py
mbounabi: "global variable" never sounds good to me, but I agree with the idea.

mbounabi: @kinnou02 what do you think about this idea?

pbougue: No global variable, as it would anyway need to be updated regularly.
So query db anytime we need it.

kirin/gtfs_rt/gtfs_rt.py
```diff
@@ -39,6 +39,35 @@
...
+            navitia_token=current_app.config.get(str("NAVITIA_GTFS_RT_TOKEN")),
+            feed_url=current_app.config.get(str("GTFS_RT_FEED_URL")),
+        )
+        gtfsrt_contributors.append(contributor_legacy)
+
+    gtfsrt_contributors.extend(
+        [
+            c
+            for c in model.Contributor.find_by_connector_type(ConnectorType.gtfs_rt.value)
+            if c.id != contributor_legacy.id
```
pbougue: If there is no contributor_legacy in file you will crash here.
Same in model.py

mbounabi: It won't crash in model. It perfectly works to compare an id with None.

pbougue: True, my bad :pray:

kadhikari:
**I realized some tests on this PR and the details are as followings:**
- 2 contributors gtfs-rt in bdd with only one active : OK
- 2  contributors gtfs-rt in bdd with both active : OK
- 2 contributors gtfs-rt with one in bdd (active) and another in the file : OK
- 1 contributor cots config / 2 contributors gtfs-rt (1 bdd and another config) : OK
- 1 contributor cots and 2 contributors gtfs-rt (all in bdd and active): OK
- 2 contributors gtfs-rt in bdd and active and one of two contributors exists also in config file: Informations of this contributor from bdd is omitted.

**Note: Presence of unique cots contributor in config file = "NAVITIA_INSTANCE" where as for gtfs-rt contribuor it's "GTFS_RT_CONTRIBUTOR".**

PR #60, opened 2015-12-21
save invalid xml into db

xlqian:
save invalid xml into db
http://jira.canaltp.fr/browse/NAVITIAII-1975

PR #266, opened 2019-09-19
Track handler exceptions

pbougue:
Move handler call inside try

As it can also raise (for example rabbit connection error) and
it's better to keep track of this in status.

Don't know how to easily add test on that


pbougue:
@kadhikari : Initial work was on IRE, https://github.com/CanalTP/kirin/pull/60/files
This was just intended to save wrong input, then we never though of including possible raise of handler, not even when we decided to use those KO status for the /status endpoint.
With time other exception were managed as they appeared, I guess.

PR #267, opened 2019-09-19
Allow reprocessing after particular errors even if ETag is unchanged

pbougue:
Previously if there was some issue in processing (for example connection loss to Navitia) we didn't retry to process the same input, as its ETag was unchanged.
Now we forget about previous' ETag if there was an error that is not related to the file itself.

:mag: Main code is in utils.py. Please read by commit with message.

TODO:
- [x] merge #266 as this branch is including it
- [x] merge #268 (and rebase) as Redis is needed to have tests running now
- [x] hand-test it:
  * With a good gtfsrt.pb : second poll avoided :heavy_check_mark: 
  * With a bad gtfsrt.pb : second poll avoided :heavy_check_mark: 
  * With a good gtfsrt.pb but no navitia linked to kirin : second poll redone, poll OK when navitia up again :heavy_check_mark: 
  * Always repolled when ETag changes :heavy_check_mark: 
- [x] maybe add unit tests if possible: Not added
  Added test just on ETag stored in redis, without the whole process (mostly basic low-level error management functions).
  :warning: as there is no poll test, it's not easy to add. It may be a separate task (here or/and in Artemis): tracked in https://jira.kisio.org/browse/NAVP-1412

Related to https://jira.kisio.org/browse/NAVITIAII-2800

### DIFF
--- a/kirin/abstract_sncf_resource.py
+++ b/kirin/abstract_sncf_resource.py
@@ -33,7 +33,7 @@ from flask_restful import Resource
 
 import logging
 from datetime import datetime
-from kirin.utils import make_rt_update, record_call, set_rtu_status_ko
+from kirin.utils import make_rt_update, record_call, set_rtu_status_ko, allow_reprocess_same_data
 from kirin.exceptions import KirinException
 from kirin import core
 from kirin.core import model
@@ -47,10 +47,21 @@ class AbstractSNCFResource(Resource):
         self.builder = builder
 
     def process_post(self, input_raw, contributor_type, is_new_complete=False):
-
-        # create a raw rt_update obj, save the raw_input into the db
-        rt_update = make_rt_update(input_raw, contributor_type, contributor=self.contributor)
         start_datetime = datetime.utcnow()
+
+        try:
+            # create a raw rt_update obj, save the raw_input into the db
+            rt_update = make_rt_update(input_raw, contributor_type, contributor=self.contributor)
+        except Exception as e:
+            # as rt_update is probably not built, make sure reprocess is allowed
+            allow_reprocess_same_data(self.contributor)
+            # regular exception handling
+            set_rtu_status_ko(rt_update, e.message, is_reprocess_same_data_allowed=True)
+            model.db.session.add(rt_update)
+            model.db.session.commit()
+            record_call("failure", reason=six.text_type(e), contributor=self.contributor)
+            raise
+
         try:
             # assuming UTF-8 encoding for all input
             rt_update.raw_data = rt_update.raw_data.encode("utf-8")
--- a/kirin/abstract_sncf_resource.py
+++ b/kirin/abstract_sncf_resource.py
@@ -33,7 +33,7 @@ from flask_restful import Resource
 
 import logging
 from datetime import datetime
-from kirin.utils import make_rt_update, record_call
+from kirin.utils import make_rt_update, record_call, set_rtu_status_ko
 from kirin.exceptions import KirinException
 from kirin import core
 from kirin.core import model
@@ -60,15 +60,13 @@ class AbstractSNCFResource(Resource):
             _, log_dict = core.handle(rt_update, trip_updates, self.contributor, is_new_complete=is_new_complete)
             record_call("OK", contributor=self.contributor)
         except KirinException as e:
-            rt_update.status = "KO"
-            rt_update.error = e.data["error"]
+            set_rtu_status_ko(rt_update, e.data["error"], is_reprocess_same_data_allowed=True)
             model.db.session.add(rt_update)
             model.db.session.commit()
             record_call("failure", reason=six.text_type(e), contributor=self.contributor)
             raise
         except Exception as e:
-            rt_update.status = "KO"
-            rt_update.error = e.message
+            set_rtu_status_ko(rt_update, e.message, is_reprocess_same_data_allowed=True)
             model.db.session.add(rt_update)
             model.db.session.commit()
             record_call("failure", reason=six.text_type(e), contributor=self.contributor)
--- a/kirin/abstract_sncf_resource.py
+++ b/kirin/abstract_sncf_resource.py
@@ -34,7 +34,7 @@ from flask_restful import Resource
 import logging
 from datetime import datetime
 from kirin.utils import make_rt_update, record_call, set_rtu_status_ko, allow_reprocess_same_data
-from kirin.exceptions import KirinException
+from kirin.exceptions import KirinException, InvalidArguments
 from kirin import core
 from kirin.core import model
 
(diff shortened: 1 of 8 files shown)


kirin/utils.py
```diff
@@ -135,11 +136,48 @@ def make_kirin_lock_name(*args):
...
+def build_redis_etag_key(contributor):
+    # type: (unicode) -> unicode
+    return "|".join([contributor, "polling_HEAD"])
+
+
+def allow_reprocess_same_data(contributor):
+    # type: (unicode) -> None
+    from kirin import redis
+
+    redis.delete(build_redis_etag_key(contributor))  # wipe previous' ETag memory
```
pbougue: :dart: Main change is here (in utils).
Other files are "just" using it.

kirin/abstract_sncf_resource.py
```diff
@@ -47,10 +47,21 @@ def __init__(self, nav_wrapper, timeout, contributor, builder):
...
         start_datetime = datetime.utcnow()
+
+        try:
+            # create a raw rt_update obj, save the raw_input into the db
+            rt_update = make_rt_update(input_raw, contributor_type, contributor=self.contributor)
+        except Exception as e:
+            # as rt_update is probably not built, make sure reprocess is allowed
+            allow_reprocess_same_data(self.contributor)
+            # regular exception handling
+            set_rtu_status_ko(rt_update, e.message, is_reprocess_same_data_allowed=True)
```
kadhikari: Here, we should pass is_reprocess_same_data_allowed=False as have already used "allow_reprocess_same_data(self.contributor)" above.
What happens if object rt_update is "non"

In fact function "make_rt_update" is simple and less susceptible to have an exception?

kirin/abstract_sncf_resource.py
```diff
@@ -60,15 +71,13 @@ def process_post(self, input_raw, contributor_type, is_new_complete=False):
             _, log_dict = core.handle(rt_update, trip_updates, self.contributor, is_new_complete=is_new_complete)
             record_call("OK", contributor=self.contributor)
         except KirinException as e:
-            rt_update.status = "KO"
-            rt_update.error = e.data["error"]
+            set_rtu_status_ko(rt_update, e.data["error"], is_reprocess_same_data_allowed=True)
             model.db.session.add(rt_update)
```
kadhikari: Here you want to allow_reprocess_same_data ?

Even if it might have done in the function handle() ?

kirin/abstract_sncf_resource.py
```diff
@@ -60,15 +71,13 @@ def process_post(self, input_raw, contributor_type, is_new_complete=False):
...
             model.db.session.commit()
             record_call("failure", reason=six.text_type(e), contributor=self.contributor)
             raise
         except Exception as e:
-            rt_update.status = "KO"
-            rt_update.error = e.message
+            set_rtu_status_ko(rt_update, e.message, is_reprocess_same_data_allowed=True)
```
kadhikari: Here also ?

kirin/gtfs_rt/model_maker.py
```diff
@@ -36,32 +36,49 @@
...
     start_datetime = datetime.datetime.utcnow()
+
+    try:
+        data = six.binary_type(proto)  # temp, for the moment, we save the protobuf as text
+        rt_update = make_rt_update(data, "gtfs-rt", contributor=contributor)
+    except Exception as e:
+        # as rt_update is probably not built, make sure reprocess is allowed
+        allow_reprocess_same_data(contributor)
+        # regular exception handling
+        set_rtu_status_ko(rt_update, e.message, is_reprocess_same_data_allowed=True)
```
kadhikari: Same question as for cots: allow_reprocess_same_data is already done so why not is_reprocess_same_data_allowed=False ?

kirin/gtfs_rt/model_maker.py
```diff
@@ -106,9 +123,9 @@ def build(self, rt_update, data):
             trip_updates.extend(tu)
 
         if not trip_updates:
-            rt_update.status = "KO"
-            rt_update.error = "No information for this gtfs-rt with timestamp: {}".format(data.header.timestamp)
-            self.log.error("No information for this gtfs-rt with timestamp: {}".format(data.header.timestamp))
+            msg = "No information for this gtfs-rt with timestamp: {}".format(data.header.timestamp)
+            set_rtu_status_ko(rt_update, msg, is_reprocess_same_data_allowed=False)
+            self.log.error(msg)
```
kadhikari: :+1:

kirin/gtfs_rt/model_maker.py
```diff
@@ -36,32 +36,49 @@
...
+        raise
+
     try:
         trip_updates = KirinModelBuilder(navitia_wrapper, contributor).build(rt_update, data=proto)
         _, log_dict = core.handle(rt_update, trip_updates, contributor)
         record_call("OK", contributor=contributor)
     except KirinException as e:
-        rt_update.status = "KO"
-        rt_update.error = e.data["error"]
+        allow_reprocess = not isinstance(e, InvalidArguments)  # reprocess is useless if input is invalid
```
kadhikari: No other errors to test?

pbougue: I checked other KirinException, and it seems that no other one is susceptible to block reprocess :disappointed_relieved:

kirin/abstract_sncf_resource.py
pbougue: It is less susceptible, but I had the case while hand-testing it: no contributor in db, so raised an error from foreign_key :smile: 
So in that case `rt_update` was `None`. Therefore `rt_update.contributor` is not available inside `set_rtu_status_ko()`.

So I will just change to check if it's None, then use one way or the other depending on it :+1: 
(same for other occurence).

kirin/abstract_sncf_resource.py
pbougue: Yes, the default would be to retry in my opinion.
Poke @kinnou02 on that point, I can change it if it seems hazardous.

kirin/abstract_sncf_resource.py
pbougue: True, I have to check the kind of Exception, like in gtfsrt. Will change that.

----

That said, yes raise in `handle()` is in fact what motivated that work.
The case was a raise because rabbitmq was unavailable. So we have to try again next time.

And the other errors related to the feed functionally are supposed to be managed inside handle() and only traced in status and error attributes of RTU.

kirin/abstract_sncf_resource.py
kinnou02: it's safer to reprocess the same feed multiple time than missing some.

pbougue:
@kadhikari added some kind of tests, hard to do better without polling.
Let me know if you had more on your mind :wink:

PR #268, opened 2019-09-20
Add and use Redis docker in tests

pbougue:
Add test using Redis (tasks) and provide a docker for them.

Those tests are failing if no Redis exists on host.
Mostly reverting commit https://github.com/CanalTP/kirin/pull/244/commits/ea279e8c31869ab912b7729700db5d064608de63 for the test part. So now testing the same thing, but one level above :tada: 

More generally, this is going to allow test of code parts using Redis.

:mag: Review by commit, as they are independent.

JIRA: https://jira.kisio.org/browse/NAVP-1336

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,8 +31,7 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 from contextlib import closing
 
-import kirin
-from kirin import app, db
+from kirin import app, db, redis_client
 import pytest
 
 from tests.docker_wrapper import postgres_docker, redis_docker
@@ -70,5 +69,5 @@ def redis_docker_fixture():
 
 @pytest.fixture(scope="session", autouse=True)
 def init_redis_db(redis_docker_fixture):
-    # re-init the redis host by overwriting Redis client (same conf than the one tested in wrapper)
-    kirin.redis = redis_docker_fixture.get_redis_client()
+    # Switch global redis-client's connection to use the Redis server from docker (instead of the conf)
+    redis_client.connection_pool = redis_docker_fixture.get_redis_connection_pool()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,26 +34,24 @@ from contextlib import closing
 from kirin import app, db
 import pytest
 
-from tests.docker_wrapper import PostgresDocker
+from tests.docker_wrapper import postgres_docker
 
 
 @pytest.yield_fixture(scope="session", autouse=True)
-def docker():
+def pg_docker_fixture():
     """
-    a docker providing a database is started once for all tests
+    a docker providing a PostgreSQL database is started once for all tests
     """
-    with closing(PostgresDocker()) as database:
-        yield database
+    with closing(postgres_docker()) as pg_db:
+        yield pg_db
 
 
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -31,30 +31,44 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 from contextlib import closing
 
+import kirin
 from kirin import app, db
```
mbounabi:
```suggestion
from kirin import app, db, redis
```
To avoid the previous line because only kirin.redis is used

tests/conftest.py
```diff
@@ -31,30 +31,44 @@
...
+    """
+    a docker providing a redis is started once for all tests
+    """
+    with closing(redis_docker()) as redis_db:
+        yield redis_db
+
+
+@pytest.fixture(scope="session", autouse=True)
+def init_redis_db(redis_docker_fixture):
+    # re-init the redis host by overwriting Redis client (same conf than the one tested in wrapper)
```
mbounabi: Not really sure what it's supposed to do. Can you please explain?

tests/docker_wrapper.py
```diff
@@ -31,68 +31,80 @@
...
     launch a temporary docker with a postgresql db
     """
 
-    def __init__(self, user=USER, pwd=PWD, dbname=DBNAME):
-        log = logging.getLogger(__name__)
+    def __init__(
+        self,
+        image_name,  # type: unicode
+        container_name=None,  # type: Optional[unicode]
+        dockerfile_obj=None,  # type: Optional[IO]
```
mbounabi: What does this parameter represents exactly?

tests/docker_wrapper.py
```diff
@@ -101,33 +113,168 @@ def __init__(self, user=USER, pwd=PWD, dbname=DBNAME):
...
+        The connection string for the database.
+        A string containing all the essentials data for a connection to a database.
+
+        :return: the connection string
+        """
+        return "postgresql://{u}:{pwd}@{h}/{dbname}".format(
+            h=self.host, u=self.user, dbname=self.dbname, pwd=self.password
+        )
+
+    def old_school_cnx_string(self):
```
mbounabi: unused function?

tests/docker_wrapper.py
pbougue: It's basically what you get from a file.open()
Which is what we get with the BytesIO(), that creates a "fake" docker_file.

tests/conftest.py
pbougue: Initially, Kirin creates a redis client with what's provided in config.

Here the session's `redis_docker_fixture` creates a docker for the whole tests, waits until it is up.
Then we change the redis client so that it connects to the redis server inside docker that was just created.

tests/conftest.py
```diff
@@ -31,30 +31,43 @@
...
     db.init_app(app)
+
+
+@pytest.yield_fixture(scope="session", autouse=True)
+def redis_docker_fixture():
+    """
+    a docker providing a redis is started once for all tests
+    """
+    with closing(redis_docker()) as redis_db:
+        yield redis_db
```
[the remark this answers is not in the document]
pbougue: It's what `closing()` does on `__exit__`

PR #269, opened 2019-09-26
Test added on purge_contributor

kadhikari:


kadhikari:
> 
> 
> The purge isn't specific to gtfs-rt, so it shouldn't be in gtfs tests, it should be in contributors_tests.py

I moved as requested

tests/integration/contributors_test.py
```diff
@@ -409,3 +411,68 @@ def test_delete_contributor(test_client, with_custom_contributors):
...
+    test_contributor_count(4)
+    with app.app_context():
+        contributors = Contributor.query.all()
+        assert contributors[0].id == "rt.tchoutchou"
+        assert contributors[1].id == "rt.vroumvroum"
+        for c in contributors:
+            assert c.is_active is True
+
+    # Post a simple gtfs-rt for contributor "rt.vroumvroum" in configuration file.
+    resp = test_client.post("/gtfs_rt", data=basic_gtfs_rt_data.SerializeToString())
```
mbounabi: Note: the tests won't work after https://github.com/CanalTP/kirin/pull/270

tests/integration/contributors_test.py
```diff
@@ -409,3 +411,68 @@ def test_delete_contributor(test_client, with_custom_contributors):
...
+    def has_rt_data(contributor_id):
+        with app.app_context():
+            return len(RealTimeUpdate.query.filter_by(contributor_id=contributor_id).all()) > 0
+
+    # We have 4 contributors and all are active:
+    test_contributor_count(4)
+    with app.app_context():
+        contributors = Contributor.query.all()
+        assert contributors[0].id == "rt.tchoutchou"
+        assert contributors[1].id == "rt.vroumvroum"
```
mbounabi: Minor : risky if an other contributor is added for an other test.
```suggestion
        contributors_id = [c.id for c in contributors]
        assert "rt.vroumvroum" in contributors_id
        assert "rt.tchoutchou" in contributors_id
```

PR #272, opened 2019-10-09
Improve newrelic monitoring a bit

pbougue:
:mag: See commit messages

JIRA: https://jira.kisio.org/browse/NAVP-1421

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -40,7 +40,6 @@ import sqlalchemy
 from sqlalchemy import desc
 from kirin.core.types import ModificationType, TripEffect, ConnectorType
 from kirin.exceptions import ObjectNotFound, InternalException
-from kirin.new_relic import record_custom_parameter
 
 db = SQLAlchemy()
 
@@ -452,7 +451,6 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
         self.error = error
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
         self.contributor_id = contributor
-        record_custom_parameter("real_time_update_id", self.id)
 
     @classmethod
     def get_probes_by_contributor(cls):
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -40,6 +40,7 @@ import sqlalchemy
 from sqlalchemy import desc
 from kirin.core.types import ModificationType, TripEffect, ConnectorType
 from kirin.exceptions import ObjectNotFound, InternalException
+from kirin.new_relic import record_custom_parameter
 
 db = SQLAlchemy()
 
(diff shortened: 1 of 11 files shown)


kirin/core/model.py
```diff
@@ -451,6 +452,7 @@ def __init__(self, raw_data, connector, contributor, status="OK", error=None, re
         self.error = error
         self.received_at = received_at if received_at else datetime.datetime.utcnow()
         self.contributor_id = contributor
+        record_custom_parameter("real_time_update_id", self.id)
```
mbounabi: Suggestion: Wouldn't it be interesting to record when a TripUpdate is created?
A RealTimeUpdate is created every time a feed is received, whether it is valid and processed or not.
A TripUpdate is createdonly when there's something to do with the RealTimeUpdate.

kirin/exceptions.py
```diff
@@ -42,6 +42,17 @@ def __init__(self, detailed_message=None):
...
+    def __str__(self):
+        code = self.code if self.code is not None else "???"
+        res = "{c} {n}:".format(c=code, n=self.name)
+        if self.description is not None:
+            res += " {} -".format(self.description)
+        if self.message is not None:
+            res += " {} -".format(self.message)
+        if self.data and "error" in self.data:
+            res += " {}".format(self.data["error"])
+        return res
```
mbounabi: Outside the scope of this PR but for information:
How does the new_relic function `record_exception` catches our KirinException to print it as defined here?

kirin/core/model.py
kadhikari: For me there will be lots of TripUpdate and will be difficult to use in analysis. 
What about record_custom_parameter("contributor", contributor) also ?

kadhikari: Sorry, parameter "contributor" already exists in record_call(...).

kinnou02: From a design point of view I find it quite ugly to have metrics inside the models.

kirin/exceptions.py
pbougue: Currently there are few places:
* Web (APM): https://github.com/CanalTP/kirin/blob/6804e9a6f2dbab3723de8048b4f536810ccf943c/kirin/api.py#L56-L73
  Here we tie flask's exceptions processing (record_exception handles it)
* Hand-made insights (via record_call): for example https://github.com/CanalTP/kirin/blob/6804e9a6f2dbab3723de8048b4f536810ccf943c/kirin/abstract_sncf_resource.py#L79
* Automatized for celery tasks (APM)... Maybe @kinnou02 can explain more about that but here is some doc for this: https://docs.newrelic.com/docs/agents/python-agent/back-end-services/python-agent-celery

kirin/core/model.py
pbougue: :+1: I moved it outside, as we have a central utils method for this.

PR #273, opened 2019-10-11
fix UnboundLocalError

mbounabi:
```model_maker.py", line 60, in handle
    if rt_update is None:
UnboundLocalError: local variable 'rt_update' referenced before assignment```
Now fixed


mbounabi:
The concept of requesting changes is to inform that something will not work as is. If you think that a fix can be applied somewhere unrelated to this PR, you can suggest it in a comment, not "request changes"!
Will this PR break something if merged? No, it will improve logs. So no need for "Changes requested"

PR #277, opened 2019-10-17
config: params modified to use configuration from database

kadhikari:
https://jira.kisio.org/browse/NAVP-1439


mbounabi:
@pbougue isn't this supposed to remove only default values? If the env variables are still set for any platform, it will work

kirin/default_settings.py
```diff
@@ -28,7 +28,7 @@
 
 # COTS configuration
 # * external configuration
-COTS_CONTRIBUTOR = os.getenv("KIRIN_COTS_CONTRIBUTOR", "realtime.cots")
+COTS_CONTRIBUTOR = os.getenv("KIRIN_COTS_CONTRIBUTOR", None)
```
pbougue: This is not defined in fab_kirin so it will break

kirin/default_settings.py
```diff
@@ -19,7 +19,7 @@
 
 NAVITIA_TIMEOUT = int(os.getenv("KIRIN_NAVITIA_TIMEOUT", 5))
 
-NAVITIA_INSTANCE = os.getenv("KIRIN_NAVITIA_INSTANCE", "sncf")
+NAVITIA_INSTANCE = os.getenv("KIRIN_NAVITIA_INSTANCE", None)
```
pbougue: If we change kirin_deployment_conf: this one is defined twice in common.py, we should remove one.

kirin/default_settings.py
```diff
@@ -55,9 +55,9 @@
 
 
 # TODO : Remove when conf from db is ready
-NAVITIA_GTFS_RT_INSTANCE = os.getenv("KIRIN_NAVITIA_GTFS_RT_INSTANCE", "sherbrooke")
+NAVITIA_GTFS_RT_INSTANCE = os.getenv("KIRIN_NAVITIA_GTFS_RT_INSTANCE", None)
 NAVITIA_GTFS_RT_TOKEN = os.getenv("KIRIN_NAVITIA_GTFS_RT_TOKEN", None)
-GTFS_RT_CONTRIBUTOR = os.getenv("KIRIN_GTFS_RT_CONTRIBUTOR", "realtime.sherbrooke")
+GTFS_RT_CONTRIBUTOR = os.getenv("KIRIN_GTFS_RT_CONTRIBUTOR", None)
```
pbougue: This one is defined only in every config file in kirin_deployment_conf, better to have it in common.py

kadhikari:
If we have configuration in kirin_deployment_conf for all those parameters with None in kirin, this PR should work

kirin/default_settings.py
kadhikari: We better deactivate in each platform during migration "env.gtfs_rt_contributor = None"

kirin/default_settings.py
kadhikari: Done in fab_kirin: https://github.com/CanalTP/kirin_deployment_conf/pull/39/files

kirin/default_settings.py
kadhikari: Added in fab_kirin and kirin_deployment_conf

PR #285, opened 2019-11-27
kirin: endpoint health added with tests

kadhikari:
- This is first version of endpoint /health with tests
- There will be some modification in /status to display details of error on database and/or navitia connectivity.
- Concerns: https://jira.kisio.org/browse/NAVITIAII-2960

### DIFF
--- a/kirin/resources/health.py
+++ b/kirin/resources/health.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 
-# Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
+# Copyright (c) 2019, Canal TP and/or its affiliates. All rights reserved.
 #
 # This file is part of Navitia,
 #     the software to build cool stuff with public transport.
@@ -31,17 +31,16 @@
 
 from __future__ import absolute_import, print_function, unicode_literals, division
 from flask_restful import Resource, abort
-import kirin
-from kirin.utils import connect_to_navitia, connect_to_database
+from kirin.utils import can_connect_to_navitia, can_connect_to_database
 
 
 class Health(Resource):
     def get(self):
         # Verify connection to navitia
-        if not connect_to_navitia():
+        if not can_connect_to_navitia():
             abort(503, message="KO")
         # Verify connection to database
-        if not connect_to_database():
+        if not can_connect_to_database():
             abort(503, message="KO")
 
         return {"message": "OK"}, 200
--- a/kirin/resources/health.py
+++ b/kirin/resources/health.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 
-# Copyright (c) 2001-2019, Canal TP and/or its affiliates. All rights reserved.
+# Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
 #
 # This file is part of Navitia,
 #     the software to build cool stuff with public transport.
@@ -30,7 +30,18 @@
 # www.navitia.io
 
 from __future__ import absolute_import, print_function, unicode_literals, division
-from .index import Index
-from .status import Status
-from .contributors import Contributors
-from .health import Health
+from flask_restful import Resource, abort
+import kirin
+from kirin.utils import connect_to_navitia, connect_to_database
+
+
+class Health(Resource):
+    def get(self):
+        # Verify connection to navitia
+        if not connect_to_navitia():
+            abort(503, message="KO")
+        # Verify connection to database
+        if not connect_to_database():
+            abort(503, message="KO")
+
+        return {"message": "OK"}, 200
--- a/kirin/resources/status.py
+++ b/kirin/resources/status.py
@@ -51,13 +51,7 @@ class Status(Resource):
         res["db_version"] = get_database_version()
         res["navitia_url"] = current_app.config[str("NAVITIA_URL")]
         res["rabbitmq_info"] = kirin.rabbitmq_handler.info()
-        if can_connect_to_navitia():
-            res["navitia_connection"] = "OK"
-        else:
-            res["navitia_connection"] = "KO"
-        if can_connect_to_database():
-            res["db_connection"] = "OK"
-        else:
-            res["db_connection"] = "KO"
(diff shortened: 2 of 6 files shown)


kirin/resources/health.py
```diff
@@ -0,0 +1,47 @@
+# coding=utf-8
+
+# Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
```
mbounabi: 2019 :wink:

kirin/utils.py
```diff
@@ -207,6 +208,24 @@ def manage_db_no_new(connector, contributor):
     poke_updated_at(last)
 
 
+def connect_to_navitia():
+    try:
+        response = requests.head(current_app.config[str("NAVITIA_URL")])
```
mbounabi: Why `head` ?

kadhikari: Just to check if navitia exist or  accessible. We used head because we don't need token for this.

kadhikari:
> So the modifications (add of redis and maybe little adds on db/navitia I guess) on /status are going to be in a different PR ?

Yes in the same PR

kadhikari:
> So the modifications (add of redis and maybe little adds on db/navitia I guess) on /status are going to be in a different PR ?
- > In the same PR
> Did you test manually/locally that it works too ?
- > Yes

kirin/resources/health.py
```diff
@@ -0,0 +1,47 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
```
woshilapin: Should this be changed to Matrix now?

kirin/resources/health.py
```diff
@@ -0,0 +1,47 @@
...
+from flask_restful import Resource, abort
+import kirin
+from kirin.utils import connect_to_navitia, connect_to_database
+
+
+class Health(Resource):
+    def get(self):
+        # Verify connection to navitia
+        if not connect_to_navitia():
+            abort(503, message="KO")
```
woshilapin: Is the `message` forwarded in anyway to the user? If yes, then maybe `KO` could be `Navitia KO` and `Database KO` (or even more detailed error messages)?

tests/integration/test_end_point.py
```diff
@@ -101,6 +103,34 @@ def test_status_from_db(setup_database):
     assert "2015-11-04T07:52:00Z" in resp["last_update"][GTFS_CONTRIBUTOR]
 
 
+def test_health(setup_database):
```
woshilapin: How about splitting this test into 3 test functions so instead of having a comment explaining the purpose of the test, the name of the function explains it? For example :
- `test_health_ok`
- `test_health_navitia_ko`
- `test_health_database_ko`

kirin/resources/health.py
nonifier: The end point `/health` is meant to be used by the Load Balancer. For us (humans), those messages `Navitia KO` and `Database KO` will be available through `/status`.

kirin/utils.py
```diff
@@ -207,6 +208,24 @@ def manage_db_no_new(connector, contributor):
...
+        return response.status_code == 200
```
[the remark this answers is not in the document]
kadhikari: It is done by abort(503, ..) in log and newrelic.

tests/check_utils.py
```diff
@@ -36,15 +36,18 @@
 import navitia_wrapper
 
 
-def api_get(url, display=False, *args, **kwargs):
+def api_get(url, display=False, check=True, *args, **kwargs):
```
nonifier: :+1:

kirin/resources/status.py
```diff
@@ -34,15 +34,30 @@
...
         res["navitia_url"] = current_app.config[str("NAVITIA_URL")]
         res["rabbitmq_info"] = kirin.rabbitmq_handler.info()
+        if can_connect_to_navitia():
+            res["navitia_connection"] = "OK"
+        else:
+            res["navitia_connection"] = "KO"
+        if can_connect_to_database():
+            res["db_connection"] = "OK"
+        else:
+            res["db_connection"] = "KO"
```
nonifier: minor
In a more "Pythonic" way you could :
```python
res["navitia_connection"] = "OK" if can_connect_to_navitia() else "KO"
res["db_connection"] = "OK" if can_connect_to_database() else "KO"
```

kirin/resources/health.py
kadhikari: I have no idea about this, we copy and paste from other existing file.

kirin/utils.py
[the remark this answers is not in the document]
mbounabi: 'abort' is a flask_restful function and it logs something in newrelic?

kirin/resources/health.py
pbougue: The idea is more to change it all at some point, but it's not tracked or planned yet.

kirin/utils.py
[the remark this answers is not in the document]
pbougue: Don't know about the readability and the message itself, just checked that it triggers `log_exception()` : https://github.com/CanalTP/kirin/blob/8f8100e58302f249d64d979d9cc1277543e89d5f/kirin/api.py#L56-L70

Maybe it's not perfect, but it's a trace (in logs and in newrelic), and didn't want to push much as there is a deadline for this work :disappointed:
It may require some conception (how should it be tracked versus other exceptions, etc ? ).

@kadhikari if you have a sample of the log that should be pretty much the same in newrelic ?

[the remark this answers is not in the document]
pbougue: Subject tracked into https://jira.kisio.org/browse/NAVP-1476

PR #287, opened 2019-12-05
Improve APM newrelic and log

pbougue:
JIRA: https://jira.kisio.org/browse/NAVP-1421

This is a partial work on improving logs/newrelic.
:mag: review by commit advised
* Remove some useless logs and some duplicates
* Downgrade some errors to warnings (bad data input)
* Centralize (hopefully clarify) the general approach of handling logs/newrelic

### DIFF
--- a/kirin/abstract_sncf_resource.py
+++ b/kirin/abstract_sncf_resource.py
@@ -34,9 +34,9 @@ from flask_restful import Resource
 import logging
 from datetime import datetime
 
-from kirin.new_relic import record_custom_parameter, must_log_apm, is_invalid_input_exception
+from kirin.new_relic import record_custom_parameter, is_invalid_input_exception
 from kirin.utils import make_rt_update, record_call, set_rtu_status_ko, allow_reprocess_same_data
-from kirin.exceptions import KirinException, InvalidArguments
+from kirin.exceptions import KirinException
 from kirin import core
 from kirin.core import model
 
@@ -85,7 +85,7 @@ class AbstractSNCFResource(Resource):
 
             log_dict.update({"exc_summary": six.text_type(e), "reason": e})
 
-            record_custom_parameter("reason", e)  # use __str__() to have complete details
+            record_custom_parameter("reason", e)  # using __str__() here to have complete details
             raise  # filters later for APM
 
         finally:
--- a/kirin/abstract_sncf_resource.py
+++ b/kirin/abstract_sncf_resource.py
@@ -75,9 +75,7 @@ class AbstractSNCFResource(Resource):
                 allow_reprocess = False  # reprocess is useless if input is invalid
 
             if rt_update is not None:
-                error = e.message
-                if isinstance(e, KirinException) and "error" in e.data:
-                    error = e.data["error"]
+                error = e.data["error"] if (isinstance(e, KirinException) and "error" in e.data) else e.message
                 set_rtu_status_ko(rt_update, error, is_reprocess_same_data_allowed=allow_reprocess)
                 model.db.session.add(rt_update)
                 model.db.session.commit()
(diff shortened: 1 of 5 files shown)


kirin/gtfs_rt/model_maker.py
```diff
@@ -53,44 +54,49 @@
...
+        status = "failure"
+        allow_reprocess = True
+        if is_invalid_input_exception(e):
+            status = "warning"  # Kirin did his job correctly if the input is invalid and rejected
+            allow_reprocess = False  # reprocess is useless if input is invalid
+
+        if rt_update is not None:
+            error = e.message
+            if isinstance(e, KirinException) and "error" in e.data:
+                error = e.data["error"]
```
woshilapin: See similar comment above.

kirin/abstract_sncf_resource.py
```diff
@@ -49,45 +51,53 @@ def __init__(self, nav_wrapper, timeout, contributor, builder):
...
+            status = "failure"
+            allow_reprocess = True
+            if is_invalid_input_exception(e):
+                status = "warning"  # Kirin did his job correctly if the input is invalid and rejected
+                allow_reprocess = False  # reprocess is useless if input is invalid
+
+            if rt_update is not None:
+                error = e.message
+                if isinstance(e, KirinException) and "error" in e.data:
+                    error = e.data["error"]
```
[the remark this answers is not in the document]
pbougue: Nice, it's better indeed: changed.
I thought of doing the same on status and allow-reprocess, but comments are quite important, and "default" value is nice there.

kirin/new_relic.py
```diff
@@ -50,6 +54,34 @@ def init(config_file):
         logger.warning("newrelic hasn't been initialized")
 
 
+def must_never_log(exception):
+    # Not tracking requests to wrong endpoints
+    if isinstance(exception, NotFound):
+        return True
+    # Not tracking requests with wrong HTTP verbs
```
mbounabi: Minor: Verbs? Why not 'methods' ?

kirin/abstract_sncf_resource.py
```diff
@@ -49,45 +51,51 @@ def __init__(self, nav_wrapper, timeout, contributor, builder):
...
+                # rt_update is not built, make sure reprocess is allowed
+                allow_reprocess_same_data(self.contributor)
+
+            log_dict.update({"exc_summary": six.text_type(e), "reason": e})
 
-        duration = (datetime.utcnow() - start_datetime).total_seconds()
-        log_dict.update({"duration": duration})
-        record_call("Simple feed publication", **log_dict)
-        logging.getLogger(__name__).info("Simple feed publication", extra=log_dict)
+            record_custom_parameter("reason", e)  # use __str__() to have complete details
```
mbounabi: Confusing comment. Does it mean that it doesn't record everything as is and if I want more details, I need to change the code?

kirin/new_relic.py
```diff
@@ -50,6 +54,34 @@ def init(config_file):
...
+
+
+def is_invalid_input_exception(exception):
+    if isinstance(exception, InvalidArguments):
+        return True
+    # Any other exception is not related to an invalid input
+    return False
+
+
+def must_log_apm(exception):
```
mbounabi: This function will let us know if we have to record an event if it isn't among the exception we aren't interested in so I'll suggest
```suggestion
def allow_apm_logging(exception):
```

kirin/abstract_sncf_resource.py
pbougue: OK I understand the confusion.
This simple `e` is using __str__, and the result is not the same as six.text_type for KirinExceptions that overloads the method (using https://github.com/CanalTP/kirin/blob/5fa93fc2b001f0ed7d7469d028f1c12521b2f9a0/kirin/exceptions.py#L45-L54).

Suggesting:
```suggestion
            record_custom_parameter("reason", e)  # using __str__() here to have complete details
```

PR #288, opened 2019-12-13
Add a general param for poll interval

pbougue:
Also default it to 10s (1s before)

JIRA: https://jira.kisio.org/browse/NDTMA-76 (see comments)

### DIFF
--- a/kirin/default_settings.py
+++ b/kirin/default_settings.py
@@ -160,11 +160,13 @@ TASK_LOCK_PREFIX = "kirin.lock"
 TASK_STOP_MAX_DELAY = int(os.getenv("KIRIN_TASK_STOP_MAX_DELAY", timedelta(seconds=10).total_seconds()))
 TASK_WAIT_FIXED = int(os.getenv("KIRIN_TASK_WAIT_FIXED", timedelta(seconds=2).total_seconds()))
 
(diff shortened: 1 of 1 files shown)


kirin/default_settings.py
```diff
@@ -160,11 +160,13 @@
 TASK_STOP_MAX_DELAY = int(os.getenv("KIRIN_TASK_STOP_MAX_DELAY", timedelta(seconds=10).total_seconds()))
 TASK_WAIT_FIXED = int(os.getenv("KIRIN_TASK_WAIT_FIXED", timedelta(seconds=2).total_seconds()))
 
+# Must be >= 1. Defines the minimal interval (seconds) between 2 tries for polling (can be longer if a task is running).
+POLLER_MIN_INTERVAL = int(os.getenv("KIRIN_POLLER_MIN_INTERVAL", timedelta(seconds=10).total_seconds()))
```
datanel: I don't understand the use of `timedelta`
Why don't you just use `POLLER_MIN_INTERVAL = int(os.getenv("KIRIN_POLLER_MIN_INTERVAL", '10'))`?

pbougue: It would work too, yes.
It makes thing more explicit (kind of self-documented). And in case it comes to minutes, it's easier to read.
Also, we did this on the rest of the conf, so following the move was fine.

PR #289, opened 2019-12-16
fixed Dockerfile

datanel:
the docker image of Kirin is now based on Debian.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,32 +6,28 @@ COPY . .
 
 #we remove protobuf from the dependency since it's already installed with c++ extension built in
 
-RUN apk --update --no-cache add \
+RUN apt-get update \
+    && apt-get install -y \
         g++ \
-        build-base \
         python-dev \
-        libstdc++ \
-        git \
-        postgresql-dev \
+        python-setuptools \
         postgresql-client \
-        libpq && \
+        git && \
     sed -i -e '/protobuf/d' requirements.txt && \
     pip install --no-cache-dir -r requirements.txt && \
     pip install --no-cache-dir gunicorn && \
     pip install --no-cache-dir newrelic && \
     python setup.py build_version && \
+    git submodule update --init && \
     python setup.py build_pbf && \
-    apk del \
+    apt-get purge -y \
         g++ \
-        build-base \
         python-dev \
-        musl \
-        postgresql-dev \
-        git
+        git && \
+    apt-get autoremove -y
 
 
 ENV PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=cpp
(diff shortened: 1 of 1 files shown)


datanel:
I removed some packages, tell me if it's  correct.

buildbotctp:

Dockerfile
```diff
@@ -6,32 +6,28 @@ COPY . .
...
-RUN apk --update --no-cache add \
+RUN apt-get update \
+    && apt-get install -qq -y \
         g++ \
-        build-base \
         python-dev \
-        libstdc++ \
-        git \
-        postgresql-dev \
-        postgresql-client \
```
woshilapin: If I understand, these 2 dependencies are already installed through `navitia/python`, right? Even if this is the case, I would keep them here because:
- it's a way to document the dependencies of Kirin
- it removes (partly) a dependency on `navitia/python` (for example, if Navitia decides to change the DB engine)
- if the dependency is already installed, it won't be installed twice

datanel: > these 2 dependencies are already installed through navitia/python
No

I don't know those 2 packages are useful. The image builds without and the webservice runs well.
It may be necessary for the workers or to be able to connect to the database from inside a container, I don't know.

woshilapin: Well, these 2 packages are already in `navitia/python`. The parent Docker image actually installs `postgresql-server-dev-all` which contains a bunch of dependencies and among these dependencies are `libql-dev` and `postgresql-client-common` which might be the Debian equivalent names for the 2 dependencies above you're removing.

Did you try to actually remove these packages (`apt-get purge libql-dev postgresql-client-common`) and see if Kirin still works?

woshilapin: So, someone very wise told me I was looking into the wrong `Dockerfile` for `navitia/python` 😛 So my comment is not accurate. Please don't take it into account. My guess is that even if Kirin depends on a connection on PG, everything is installed through `pip` (not `apt`). So please, remove them if they're not needed.

pbougue: Looks like it is used in a corner here : https://github.com/CanalTP/navitia-docker-compose/blob/master/kirin/docker-compose_kirin.yml#L19
So for Artemis.
This is probably a deprecated way of managing things, though (using kirin configurator there also).

mbounabi: The package `postgresql-client` is needed for Kirin to work correctly

mbounabi:
> We need to test it with ArtemisNG at least, I'll try to check that.

Tested. With the package `postgresql-client` in this Dockerfile, all GuichetUnique tested passed

Dockerfile
datanel: ok done

PR #291, opened 2019-12-20
Doc: adding specs for kirin to nav interface

papailio:
Also handles specificities mentioned in https://jira.kisio.org/browse/ND-568 (so https://jira.kisio.org/browse/ND-538)

### DIFF
--- a/documentation/kirin2nav.md
+++ b/documentation/kirin2nav.md
@@ -13,7 +13,6 @@ In the following, the optional fields that are not provided in the output feed a
 Field | Type | Description
 --- | --- | ---
 gtfs_realtime_version | String | Fixed value `1`.
-incrementality | Enum | Fixed value `FULL_DATASET`.
 timestamp | unit64 | The UTC POSIX time when the feed destined to Navitia is created in Kirin.
 
 ### FeedEntity
@@ -29,7 +28,7 @@ trip | TripDescriptor | The `TripDescriptor` that identifies the vehicle journey
 vehicle | VehicleDescriptor | The `VehicleDescriptor` that provides additional information about the vehicle journey.
 stop_time_update | List of StopTimeUpdates | The `StopTimeUpdates` of the vehicle journey.
 trip_message | String | Extended field. The information for the vehicle journey to be displayed (if specified).
-effect | Enum | Extended field. The effect of the affected vehicle journey to be displayed. It follows the [effect](https://gtfs.org/reference/realtime/v2/#enum-effect) of an Alert message as it is specified in GTFS-RT. The possible values are `NO_SERVICE`, `REDUCED_SERVICE`, `SIGNIFICANT_DELAYS`, `DETOUR`, `ADDITIONAL_SERVICE`, `MODIFIED_SERVICE`, `OTHER_EFFECT`, `UNKNOWN_EFFECT`, `STOP_MOVED`.
+effect | Enum | Extended field. The effect of the affected vehicle journey to be displayed. It follows the [effect](https://gtfs.org/reference/realtime/v2/#enum-effect) of an Alert message as it is specified in GTFS-RT. The possible values are `NO_SERVICE`, `REDUCED_SERVICE`, `SIGNIFICANT_DELAYS`, `DETOUR`, `ADDITIONAL_SERVICE`, `MODIFIED_SERVICE`, `UNKNOWN_EFFECT`.
 headsign | String | Extended field. The trip headsign of the vehicle journey as it is found in Navitia.
 
 ### TripDescriptor
@@ -37,7 +36,7 @@ Field | Type | Description
 --- | --- | ---
 trip_id | String | Navitia's id of the vehicle journey being updated.
 start_date | String | The start date in UTC of the static schedule vehicle journey to be updated as it is found in Navitia.
-schedule_relationship | Enum | The relation between the vehicle journey being updated and the static schedule. The possible values are `SCHEDULED`, `ADDED`, `UNSCHEDULED`, `CANCELED`. This field is deprecated in favor of the more detailed `TripUpdate.effect`.
+schedule_relationship | Enum | The relation between the vehicle journey being updated and the static schedule. The possible values are `SCHEDULED`, `ADDED`, `CANCELED`. This field is deprecated in favor of the more detailed `TripUpdate.effect`.
 contributor | String | Extended field. The identifier of the realtime contributor.
 company_id | String | Extended field. The transport operator of the vehicle journey as it is found in Navitia.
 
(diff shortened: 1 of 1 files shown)


documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+The realtime feed is defined with relation to the existing static schedule in navitia.
+Each feed provides realtime information for the complete vehicle journey and
+overwrites all previous realtime information.
+In the following, the optional fields that are not provided in the output feed are not specified.
+
+### FeedHeader
+Field | Type | Description
+--- | --- | ---
+gtfs_realtime_version | String | Fixed value `1`.
+incrementality | Enum | Fixed value `FULL_DATASET`.
```
kinnou02: We use `DIFFERENTIAL` at least for realtime update, `FULL_DATASET` should be use on kraken reloading but I don't think it's the case.

documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+timestamp | unit64 | The UTC time when the feed is created given in POSIX time.
+
+### FeedEntity
+Field | Type | Description
+--- | --- | ---
+id | String | The id of the navitia vehicle journey being updated.
```
pbougue: This is a generated UUID (stable along different updates of the same Navitia dated-VJ) but this is not the id of the VJ in Navitia.
It's kind of the id of the disruption (one per dated-VJ).

documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+
+### VehicleDescriptor
+Field | Type | Description
+--- | --- | ---
+physical_mode_id | String | Extended field. The physical mode of the vehicle journey as it is found in navitia.
+
+### StopTimeUpdate
+Field | Type | Description
+--- | --- | ---
+stop_sequence | String | The order of the given stop for the updated vehicle journey in navitia.
```
pbougue: Looks like it is not filled currently for the Navitia's output.
Instead, the list's order is trusted.

documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+departure | StopTimeEvent | The timing information at the departure of this stop_point.
+schedule_relationship | Enum | The possible values are `SCHEDULED`, `SKIPPED`, `NO_DATA`, `UNSCHEDULED`.
+stoptime_message | String | Extended field. The text to be displayed for the given stop_point.
+
+### StopTimeEvent
+Field | Type | Description
+--- | --- | ---
+time | unit64 | Updated time in POSIX time for the StopTime.
+delay | String | Delay in seconds of the update at the StopTime. The default value is `0`.
+stop_time_event_relationship | Enum | Extended field. The type of the update of the StopTime in relation with the static schedule. The possible values are `SCHEDULED`, `SKIPPED`, `NO_DATA`, `UNSCHEDULED`.
```
pbougue: This field is deprecated in favor of `stop_time_event_status`

documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+physical_mode_id | String | Extended field. The physical mode of the vehicle journey as it is found in navitia.
+
+### StopTimeUpdate
+Field | Type | Description
+--- | --- | ---
+stop_sequence | String | The order of the given stop for the updated vehicle journey in navitia.
+stop_id | String | The identifier of the stop_point as it is found in navitia.
+arrival | StopTimeEvent | The timing information at the arrival of the given stop_point.
+departure | StopTimeEvent | The timing information at the departure of this stop_point.
+schedule_relationship | Enum | The possible values are `SCHEDULED`, `SKIPPED`, `NO_DATA`, `UNSCHEDULED`.
```
pbougue: Looks like it is not filled currently for the Navitia's output.
Instead, it is later deduced from StopTimeEvents inside Kraken.

documentation/kirin2nav.md
```diff
@@ -0,0 +1,66 @@
...
+stoptime_message | String | Extended field. The text to be displayed for the given stop_point.
+
+### StopTimeEvent
+Field | Type | Description
+--- | --- | ---
+time | unit64 | Updated time in POSIX time for the StopTime.
+delay | String | Delay in seconds of the update at the StopTime. The default value is `0`.
```
pbougue: We should probably add that the value of the delay is not trusted, in favor of `time` (as it is redundant and may be inconsistent considering base-schedule time).
Currently in kraken, `delay` is only used to compute StopTime status (DELAYED vs. UNCHANGED, when stop_time_event_status==SCHEDULED).

papailio:
All comments are taken into account in the last commit, except the one regarding `incrementality`

documentation/kirin2nav.md
pbougue: As discussed, I confirm that this is not read by Kraken.
So we may also skip it from the documentation and explain how the reading is done.
The field is required, so it's filled, but there is no meaning to that currently.

I vote for a removal of this `incrementality` description and aside (later?) a description of how it's read, as specified in https://jira.kisio.org/browse/ND-568.

The other question is: understanding what `FULL_DATASET` means and what `DIFFERENTIAL` means.
* Looks like `FULL_DATASET` means that it's an "erase and replace" (so remove all known RT information from that source before processing incoming one).
* And probably `DIFFERENTIAL` would be closer to what we always do (ongoing): updating VJ as the information comes and keep the rest as-is. But there are still ongoing discussions about that.

documentation/kirin2nav.md
```diff
@@ -0,0 +1,64 @@
...
+trip_update | TripUpdate | The `TripUpdate` message with the realtime data for the vehicle journey.
+
+### TripUpdate
+Field | Type | Description
+--- | --- | ---
+trip | TripDescriptor | The `TripDescriptor` that identifies the vehicle journey.
+vehicle | VehicleDescriptor | The `VehicleDescriptor` that provides additional information about the vehicle journey.
+stop_time_update | List of StopTimeUpdates | The `StopTimeUpdates` of the vehicle journey.
+trip_message | String | Extended field. The information for the vehicle journey to be displayed (if specified).
+effect | Enum | Extended field. The effect of the affected vehicle journey to be displayed. It follows the [effect](https://gtfs.org/reference/realtime/v2/#enum-effect) of an Alert message as it is specified in GTFS-RT. The possible values are `NO_SERVICE`, `REDUCED_SERVICE`, `SIGNIFICANT_DELAYS`, `DETOUR`, `ADDITIONAL_SERVICE`, `MODIFIED_SERVICE`, `OTHER_EFFECT`, `UNKNOWN_EFFECT`, `STOP_MOVED`.
```
mbounabi: Effects `OTHER_EFFECT` and STOP_MOVED` aren't used by Navitia. Is it worth mentionning them?

documentation/kirin2nav.md
mbounabi: I also think it should be removed if not used

documentation/kirin2nav.md
pbougue: :heavy_check_mark: Removed unused enum values everywhere in doc.
Although the serialization is ready for much of them.

documentation/kirin2nav.md
```diff
@@ -0,0 +1,69 @@
...
+departure | StopTimeEvent | The timing information at the departure of this stop_point.
+stoptime_message | String | Extended field. The text to be displayed for the given stop_point.
+
+### StopTimeEvent
+Field | Type | Description
+--- | --- | ---
+time | unit64 | Updated time in UTC POSIX time for the StopTime.
+delay | String | Delay in seconds of the update at the StopTime. The default value is `0`. This field is only used to determine the StopTime status and `time` always takes precedence over `delay`.
+stop_time_event_relationship | Enum | Extended field. The type of the update of the StopTime in relation with the static schedule. The possible values are `SCHEDULED`, `SKIPPED`. This field is deprecated in favor of `stop_time_event_status`.
+stop_time_event_status | Enum | Extended field. The possible values are `SCHEDULED`, `DELETED`, `ADDED`, `DELETED_FOR_DETOUR`, `ADDED_FOR_DETOUR`.
```
mbounabi: Add `DELAYED` to this enum. To be discussed between Core and Core-Tools team because it should be handled by Kraken if modified:
https://github.com/CanalTP/chaos-proto/blob/dd1bfccccd0dada24596c31eb29e6b239929303b/kirin.proto#L21

This will allow to determine the delay with only the 'status' and remove the 'delay' field.

pbougue: Tracked into https://jira.kisio.org/browse/ND-631 to be discussed between teams.

PR #293, opened 2020-01-07
Add a 'docker-compose.yml' for development

woshilapin:
This is to simplify the life of the developer who is working on his own machine and don't want to install all the dependencies. Note that this is only a bootstrap `docker-compose.yml` that can probably be improved a lot so I'm creating only a Draft PR for now, let's discuss it and improve it before merging it.

Ideas for improvements so far:
- a script to initialize the DB (instead of using only the environment variable in `docker-compose.yml`); this would allow to have multiple tables like `kirin` and `chaos_testing`
- add entries in `docker-compose.yml` to also start up the kirin instances
- define a volume to persist the DB?

I can try to work these improvements but I first want to be sure some of you are fine with the idea of this PR before investing more time on it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "3.7"
+services:
+    db:
+        image: postgres:12
+        restart: always
+        environment:
+            POSTGRES_USER: "navitia"
+            POSTGRES_PASSWORD: "navitia"
+            POSTGRES_DB: "kirin"
+        ports:
+            - "5432:5432"
+    redis:
+        image: redis:5
+        restart: always
+        ports:
+            - "6379:6379"
+    mq:
+        image: rabbitmq:3
+        restart: always
+        ports:
+            - "5672:5672"


buildbotctp:

woshilapin:
Does anyone has some time to just give some feedback about the direction of this proposition?

woshilapin:
Some notes about next steps:
- make the port configurable (so in the case the local host already has this port occupy, it's possible to change the port of the docker-compose containers); use `.env` for default values
- use [`kirin-configurator`](https://github.com/CanalTP/navitia-docker-compose/blob/master/kirin/Dockerfile-kirin-configurator) to load/migrate the DB
- do not care of `chaos_testing`, it's not really useful for local development anyway
- maybe move back [`docker-compose_kirin.yml`](https://github.com/CanalTP/fab_kirin/blob/master/fabfile/templates/docker-compose_kirin.yml) into `CanalTP/kirin` repository so dev and prod deployment uses the same configurations

pbougue:
After this ideas, some work has been done (that may still be perfected when using it).
Still this is a good step in my opinion, and comprise ideas listed here.

Central work is CanalTP/fab_kirin#49
Induced CanalTP/fab_kirin#48, #315, CanalTP/kirin_deployment_conf#60

So I now close this PR

PR #294, opened 2020-01-08
GTFS-RT endpoints: Separate Index management from POST management

pbougue:
This will avoid crash when requesting GET on `/gtfs_rt/id.contributor` (param  `id` was provided to `get()`, leading to a python crash)

Also `/gtfs_rt` now returns "classic" index:
```
{
  "realtime.toto": {
    "href": "http://localhost:54746/gtfs_rt/realtime.toto"
  },
  "realtime.tata": {
    "href": "http://localhost:54746/gtfs_rt/realtime.tata"
  }
}
```

### DIFF
--- a/tests/integration/gtfs_rt_test.py
+++ b/tests/integration/gtfs_rt_test.py
@@ -124,14 +124,13 @@ def basic_gtfs_rt_data_without_delays():
 
 def test_get_gtfs_rt_contributors():
     """
-    Get GTFS-RT contributors
+    Get GTFS-RT contributors index
     """
     resp = api_get("/gtfs_rt")
-    assert "gtfs-rt" in resp
-    assert len(resp["gtfs-rt"]) == 2
-    for contributor in resp["gtfs-rt"]:
-        assert contributor["connector_type"] == "gtfs-rt"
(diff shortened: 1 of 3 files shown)


mbounabi:
id is not provided? It's the primary key of a contributor and can't be null:
https://github.com/CanalTP/kirin/blob/9ad1198a8c30cba95abcec7c13f763bbd72fb110/kirin/core/model.py#L526
What's the use case where you had a problem with a missing id?

Also, what does the request "/contributors/<contributor_id>" return now? Is it still possible?

pbougue:
@mbounabi : sorry my first comment was wrong (updated it).
The issue was requesting a get on gtfs_rt/toto (so a get with id). The `id` was too much.

tests/integration/gtfs_rt_test.py
```diff
@@ -124,14 +124,13 @@ def basic_gtfs_rt_data_without_delays():
 
 def test_get_gtfs_rt_contributors():
```
mbounabi: A test can also be added for GET `/gtfs_rt/<id>` with an expected failure and a proper message

PR #295, opened 2020-01-08
Cleanup after kirin2nav doc

pbougue:
TODO:
- [x] Validate #291 (deprecation, targeted behavior)

Content:
* Add little comments and explicit/correct code (after crawling for review on #291)
* Update setup part of readme (after kirin install-workshop)
* Minor cleanup


kirin/core/handler.py
```diff
@@ -199,7 +200,7 @@ def handle(real_time_update, trip_updates, contributor, is_new_complete=False):
 
     persist(real_time_update)
 
-    feed = convert_to_gtfsrt(real_time_update.trip_updates)
+    feed = convert_to_gtfsrt(real_time_update.trip_updates, gtfs_realtime_pb2.FeedHeader.DIFFERENTIAL)
```
datanel: the default value for the second parameter is already `gtfs_realtime_pb2.FeedHeader.DIFFERENTIAL`
Do you want to just explicit the value?

pbougue: Yep, that was the point. And to pin that value in case of change in default one.

PR #297, opened 2020-01-17
COTS: Ignore back-to-base-schedule stop-removal for trip status

pbougue:
JIRA: https://jira.kisio.org/browse/ND-554
(please use the xls file provided there if you can for review)

:mag: Review by commit (with message) is advised

Functional changes:
* For trip-status processing: ignore back-to-normal stop-removal (not considered as a reduced-service): https://github.com/CanalTP/kirin/commit/1c5872570fc044efc6c318b2807f0af67e2ae812, https://github.com/CanalTP/kirin/commit/8db9e6130aea91ed76df381dae0715803ce2cfa1 (refactor)
:warning: The way "back-to-normal stop-removal" is determined is a bit shaky in case stop-area appears multiple times in a trip.
* minor various code update and comments: https://github.com/CanalTP/kirin/commit/4c0cadd8558834d1793a28f407bdbcdb38e487dd

Tests changes:
* Add/update various sequence tests for ignoring back-to-normal stop-removal: https://github.com/CanalTP/kirin/commit/1c5872570fc044efc6c318b2807f0af67e2ae812
* Add sequence tests mixing both: https://github.com/CanalTP/kirin/commit/af6bbce318d6193336d4a38285cdde555c19e627
* Add tests for functional cases manageable by Kirin, but maybe not the way they will be actually received in COTS feeds (see table in JIRA): https://github.com/CanalTP/kirin/commit/0251b126e17122a8c06732f858d1266c3b59867c , https://github.com/CanalTP/kirin/commit/0ef8eee6647222f9f779d77847e97ed32211efa4

TODO (hence `do_not_merge`):
- [x] merge #299 (first subset of this PR)
- [ ] Confirm that trip-status trick is really needed with real-life use-case.


kirin/cots/model_maker.py
```diff
@@ -518,10 +520,24 @@ def _make_trip_update(self, json_train, vj, action_on_trip=ActionOnTrip.NOT_ADDE
...
+                    # TODO: be more robust to cases with loops, or where stop order changes in RT.
+                    # Probably do a synchronized pass over realtime and base-schedule VJs
+                    vj_st, _ = get_navitia_stop_time_sncf(
+                        cr=get_value(pdp, "cr"),
+                        ci=get_value(pdp, "ci"),
+                        ch=get_value(pdp, "ch"),
+                        nav_vj=vj.navitia_vj,
+                    )
+                    if not vj_st:
+                        stop_event_status = ModificationType.none.name
```
nonifier: Maybe I misunderstand the whole thing, but it sounds weird not to be able to find the correct VJ in a case of a delete. 
Are you sure you want to silently ignore this case ?

nonifier: Also, this function is already 150 lines long. I would be much nicer split this logic up in a separate function.

pbougue: True for the function, I will try to do something about it.

The issue is not finding the correct VJ, it's finding the correct **stop-time** (especially in cases where one stop-area appears multiple times in the same trip)

mbounabi: For the first comment @nonifier, we discussed this case with the SNCF team, it's the use they want. If a stop is deleted and not in theoritical vj, we consider that nothing happens as it is an added stop that was deleted.

pbougue:
Closing it, as this is not needed anymore.

PR #301, opened 2020-02-11
Purge trip_update from deactivated contributors too

pbougue:
Related to https://jira.kisio.org/browse/ND-556

This will allow purge of db rows linked to deactivated contributors (so later complete removal of deactivated contributors).

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -539,7 +539,9 @@ class Contributor(db.Model):  # type: ignore
         self.is_active = is_active
 
     @classmethod
-    def find_by_connector_type(cls, type):
+    def find_by_connector_type(cls, type, include_deactivated=False):
+        if include_deactivated:
+            return cls.query.filter_by(connector_type=type).all()
         return cls.query.filter_by(connector_type=type, is_active=True).all()
 
     @classmethod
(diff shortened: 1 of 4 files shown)


kirin/core/model.py
```diff
@@ -539,7 +539,9 @@ def __init__(self, id, navitia_coverage, connector_type, navitia_token=None, fee
         self.is_active = is_active
 
     @classmethod
-    def find_by_connector_type(cls, type):
+    def find_by_connector_type(cls, type, include_deactivated=False):
+        if include_deactivated:
+            return cls.query.filter_by(connector_type=type).all()
         return cls.query.filter_by(connector_type=type, is_active=True).all()
```
pascal-vs: why not
``` python
return cls.query.filter_by(connector_type=type, is_active=!include_deactivated).all()
```
or if you prefer to avoid the '!', you can change all the "include_deactivated" in your program by "only_activated" and swap boolean values.

pbougue: This proposition returns `actives` **or** `inactives`.
The goal is to return `actives` **or** `all`. That's why there is no filter in the other request.

pascal-vs: actually you are building an SQL query, I understand better. Thanks!

PR #303, opened 2020-02-13
Adding CORS to API

prhod:


requirements.txt
```diff
@@ -4,6 +4,7 @@ Flask-RESTful==0.3.7
 Flask-SQLAlchemy==2.4.1
 Flask-Script==2.0.6
 Flask-Caching==1.7.2
+Flask-Cors==3.0.6
```
pbougue: Why not 3.0.8 on that one?

prhod: updated and tested :+1:

PR #304, opened 2020-02-13
Manage polling interval by contributor

pbougue:
https://jira.kisio.org/browse/ND-558

Difference with JIRA: store last_call_datetime in redis, as it is already used at the same place for lock.
Seemed like a simple way to do and keep conf separate from mechanic resources

:white_check_mark: Hand-tested

### DIFF
--- a/kirin/default_settings.py
+++ b/kirin/default_settings.py
@@ -156,17 +156,16 @@ REDIS_LOCK_TIMEOUT_POLLER = int(
 REDIS_LOCK_TIMEOUT_PURGE = int(os.getenv("KIRIN_REDIS_LOCK_TIMEOUT_PURGE", timedelta(hours=12).total_seconds()))
 
 TASK_LOCK_PREFIX = "kirin.lock"
+TASK_LAST_CALL_DATETIME_PREFIX = "kirin.last_exec_datetime"
 
 TASK_STOP_MAX_DELAY = int(os.getenv("KIRIN_TASK_STOP_MAX_DELAY", timedelta(seconds=10).total_seconds()))
 TASK_WAIT_FIXED = int(os.getenv("KIRIN_TASK_WAIT_FIXED", timedelta(seconds=2).total_seconds()))
 
-# Must be >= 1. Defines the minimal interval (seconds) between 2 tries for polling (can be longer if a task is running).
-POLLER_MIN_INTERVAL = int(os.getenv("KIRIN_POLLER_MIN_INTERVAL", timedelta(seconds=10).total_seconds()))
 CELERYBEAT_SCHEDULE = {
     "poller": {
         "task": "kirin.tasks.poller",
-        "schedule": timedelta(seconds=POLLER_MIN_INTERVAL),
-        "options": {"expires": timedelta(seconds=POLLER_MIN_INTERVAL).total_seconds()},
+        "schedule": timedelta(seconds=1),
+        "options": {"expires": timedelta(seconds=1).total_seconds()},
     },
     "purge_gtfs_trip_update": {
         "task": "kirin.tasks.purge_gtfs_trip_update",
--- a/kirin/default_settings.py
+++ b/kirin/default_settings.py
@@ -161,11 +161,16 @@ TASK_LAST_CALL_DATETIME_PREFIX = "kirin.last_exec_datetime"
 TASK_STOP_MAX_DELAY = int(os.getenv("KIRIN_TASK_STOP_MAX_DELAY", timedelta(seconds=10).total_seconds()))
 TASK_WAIT_FIXED = int(os.getenv("KIRIN_TASK_WAIT_FIXED", timedelta(seconds=2).total_seconds()))
 
+# Must be >= 1. Defines the general minimal interval (seconds) between 2 possible tries for polling.
+# WARNING: this is anyway also managed by retrieval_interval for each contributor
+# (can also  be longer if a task is running).
+POLLER_MIN_INTERVAL = int(os.getenv("KIRIN_POLLER_MIN_INTERVAL", timedelta(seconds=1).total_seconds()))
+
 CELERYBEAT_SCHEDULE = {
     "poller": {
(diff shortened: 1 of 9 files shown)


kirin/default_settings.py
```diff
@@ -156,17 +156,16 @@
 REDIS_LOCK_TIMEOUT_PURGE = int(os.getenv("KIRIN_REDIS_LOCK_TIMEOUT_PURGE", timedelta(hours=12).total_seconds()))
 
 TASK_LOCK_PREFIX = "kirin.lock"
+TASK_LAST_CALL_DATETIME_PREFIX = "kirin.last_exec_datetime"
 
 TASK_STOP_MAX_DELAY = int(os.getenv("KIRIN_TASK_STOP_MAX_DELAY", timedelta(seconds=10).total_seconds()))
 TASK_WAIT_FIXED = int(os.getenv("KIRIN_TASK_WAIT_FIXED", timedelta(seconds=2).total_seconds()))
 
-# Must be >= 1. Defines the minimal interval (seconds) between 2 tries for polling (can be longer if a task is running).
-POLLER_MIN_INTERVAL = int(os.getenv("KIRIN_POLLER_MIN_INTERVAL", timedelta(seconds=10).total_seconds()))
```
woshilapin: Do you really need to remove this configuration parameter? Leaving it configurable might come up handy in the future if doing all the check in a second start to be to short? (very unlikely, I agree, but who knows...)

kirin/gtfs_rt/tasks.py
```diff
@@ -106,15 +108,34 @@ def _retrieve_gtfsrt(config):
...
+        now = datetime.utcnow()
+
+        if last_exe_dt and now <= last_exe_dt + as_duration(retrieval_interval):
+            # do nothing if the last call is too recent
+            new_relic.ignore_transaction()
+            return
+
+        # store last_call_datetime in redis
+        # to avoid storing things forever in redis: set an expiration duration of (interval + "security" margin)
+        redis_client.set(last_exe_dt_name, now.strftime(str_dt_format), ex=retrieval_interval + 10)
```
woshilapin: Should this update of Reddis Last Call be updated later in the code, after the HTTP request has actually been done and successful? Because if the call happens and fails, then you won't pull until next `retrieval_interval`, right?

kirin/gtfs_rt/tasks.py
```diff
@@ -106,15 +108,34 @@ def _retrieve_gtfsrt(config):
...
+
+        last_exe_dt_name = make_kirin_last_call_dt_name(func_name, contributor)
+        last_exe_dt_str = redis_client.get(last_exe_dt_name)
+        last_exe_dt = datetime.strptime(last_exe_dt_str, str_dt_format) if last_exe_dt_str else None
+        now = datetime.utcnow()
+
+        if last_exe_dt and now <= last_exe_dt + as_duration(retrieval_interval):
+            # do nothing if the last call is too recent
+            new_relic.ignore_transaction()
+            return
```
woshilapin: How about extracting this section in a function? Because this whole block is only about checking if you actually need to do the request or not. Seems like a good candidate for a named block of code aka function :wink:

kirin/gtfs_rt/tasks.py
pbougue: After discussion, no action to be done on that, no strong opinion too.
The discussion is on how Kirin should react when something goes wrong on retrieval.
For now it's just a chance missed, and the new try is indeed going to be in `retrieval_interval` seconds. Small argument to keep it that way : this avoids DOS on provider when something goes wrong (a circuit breaker would even tend to larger interval).

pbougue:
Jenkins, test this please.

tests/integration/contributors_test.py
```diff
@@ -305,6 +311,7 @@ def test_post_get_put_to_ensure_API_consitency(test_client):
         "navitia_token": "blablablabla",
         "feed_url": "http://nihongo.jp",
         "connector_type": "gtfs-rt",
+        "retrieval_interval": 180,
```
kadhikari: Minor: You are not testing this value?

tests/integration/contributors_test.py
```diff
@@ -93,6 +93,7 @@ def test_get_contributors_with_specific_id(test_client, with_custom_contributors
     assert contrib[0]["connector_type"] == "gtfs-rt"
     assert contrib[0]["navitia_token"] == "my_other_token"
     assert contrib[0]["feed_url"] == "http://otherfeed.url"
+    assert contrib[0]["retrieval_interval"] == 10
```
kadhikari: Minor: You could also test the contributor "realtime.sherbrooke"

kirin/default_settings.py
```diff
@@ -156,12 +156,16 @@
 REDIS_LOCK_TIMEOUT_PURGE = int(os.getenv("KIRIN_REDIS_LOCK_TIMEOUT_PURGE", timedelta(hours=12).total_seconds()))
 
 TASK_LOCK_PREFIX = "kirin.lock"
+TASK_LAST_CALL_DATETIME_PREFIX = "kirin.last_exec_datetime"
```
kadhikari: Minor: I don't find "last_exec_datetime" ?

datanel: @kadhikari why do you search `last_exec_datetime` ? Don't you have to search by `TASK_LAST_CALL_DATETIME_PREFIX` instead?

kadhikari: I was looking the configuration part in fab_kirin and other git depos..

datanel: seen with @kadhikari `kirin.last_exec_datetime` is just a string used to generate a name

tests/integration/contributors_test.py
pbougue: It is tested below: https://github.com/CanalTP/kirin/pull/304/files#diff-0551f96323f530fc84374722cf78bbdeR325
But I'll add intermediate tests (after post and get, too)

PR #305, opened 2020-02-13
Remove unused received_at from RealTimeUpdate

pbougue:
Cleanup part of https://jira.kisio.org/browse/ND-558

TODO:
- [ ] **check that db migration is stacked on the last master's** migration just before merge
- [x] open follow-up PR (db cleanup) to be merged once this is deployed everywhere: #310 
- [x] merge (and rebase) #304 (as it is stacked over it to avoid migration conflicts)
- [x] report corrections from reviews on #304 if needed
- [x] merge (and rebase) #308 (as it is a db hotfix, and it is stacked over it)

:mag: review by commit

### DIFF
--- a/kirin/core/model.py
+++ b/kirin/core/model.py
@@ -74,7 +74,7 @@ def gen_uuid():
 
 class TimestampMixin(object):
     created_at = db.Column(db.DateTime(), default=datetime.datetime.utcnow, nullable=False)
-    updated_at = db.Column(db.DateTime(), default=None, onupdate=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime(), default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow)
 
 
 Db_TripEffect = db.Enum(*[e.name for e in TripEffect], name="trip_effect")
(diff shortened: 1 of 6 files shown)


pbougue:
Jenkins, test this please.

kirin/core/model.py
```diff
@@ -74,7 +74,7 @@ def gen_uuid():
 
 class TimestampMixin(object):
     created_at = db.Column(db.DateTime(), default=datetime.datetime.utcnow, nullable=False)
-    updated_at = db.Column(db.DateTime(), default=None, onupdate=datetime.datetime.utcnow)
+    updated_at = db.Column(db.DateTime(), default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow)
```
kadhikari: This modification will initialize the column value to now. If we need to know if the record is modified or not we must compare with the column created_at.

pbougue: True.
And values might differ a tiny-bit depending on when `utcnow()` is called.

Still, as far as I debugged it, I wanted to sort on `updated_at, defaulting to created_at` which is not super easy to do without this. For example on `trip_update`.

In my opinion, this is a more interesting (debug) feature than retrieving updated (or un-updated) objects.
And we could probably still compare that the 2 values are less than 1s different (more easily than sort issue mentionned above), no?

Open to remove this and be contradicted on any assumption here (please :pray: )

PR #309, opened 2020-02-25
Really clean old contributors from db

pbougue:
This time for good, we can remove "contributor" columns as they
are unused (write or read) in previous version (since 4.0.1)

Follow-up of #308 

Also add a tiny doc.

:mag: review by commit (last 2 only)

TODO:
- [x] rebase and **stack migration on top of master's one** just before merge (update "down revision")
- [x] wait to have `4.0.1` deployed everywhere (including in prod)
- [x] ~merge (and rebase) #305 before (as it is stacked over it to avoid migration conflicts)~
  This PR is now directly on top of master ~(as #305 doesn't do any migration)~ for readability (#305 has migration, we'll see who gets merged first.)
- [x] Test locally:  migrate to `3.0.0`, add a feed, migrate to `4.0.1`, migrate to this > prod case
  > everything migrated as planned, column "contributor" removed and retrieval_interval exists, working
- [x] Test locally: migrate to `4.0.0`, add a feed, migrate to this (`4.0.1`) > custo case
  > everything migrated as planned, column "contributor" removed and retrieval_interval exists, working

JIRA: https://jira.kisio.org/browse/NDTMA-122

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,8 +64,17 @@ honcho run ./manage.py db migrate
 ```
 This will generate a new migration file, that you can amend at your will.
 
+:bulb: Keep in mind that db upgrade is done **before** deployment of Kirin (and Kirin must work all the way).  
+Also, where Kirin is duplicated, at some point both Kirin version `n-1` and Kirin version `n` are
+running, using the same upgraded database.
+
 :warning: To ensure safe db migrations for both upgrade (deploy) and downgrade (rollback), please make sure that:  
-Kirin version `n` is able to read/write in db version `n+1` FILLED by Kirin `n+1` (it's the case on rollback)
+Kirin version `n` is able to read/write in db version `n+1` FILLED by Kirin `n+1` (it's the case on rollback).  
+As for a (column) removal, one should ensure that version `n-1` (and of course `n`) don't use (read or write)
+the "object" removed from db.
+So first (in version `n-1`) do a PR removing **any** use in python (especially in model.py) but keeping db
+almost as-is, with just a "nullable" migration if needed.
+Then (version `n`) do a PR with only the db migration, removing the "object" unused.
 
 
 ## Roles and architecture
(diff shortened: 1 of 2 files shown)


migrations/versions/1f13d13948dd_rm_rtu_received_at.py
```diff
@@ -0,0 +1,27 @@
...
+# revision identifiers, used by Alembic.
+revision = "1f13d13948dd"
+down_revision = "ab44c0f366df"
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.drop_column("real_time_update", "received_at")
```
kadhikari: As for deleting the column 'contributor', we will have the same problem while deleting this column. This should be done in the next version.

CONTRIBUTING.md
```diff
@@ -64,8 +64,16 @@ honcho run ./manage.py db migrate
 ```
 This will generate a new migration file, that you can amend at your will.
 
+:bulb: Keep in mind that db upgrade is done **before** deployment of Kirin (and Kirin must work all the way).  
+And that where Kirin is duplicated, at some point both Kirin version `n-1` and Kirin version `n` are
```
woshilapin: Did you missed a word?

```suggestion
And that is where Kirin is duplicated, at some point both Kirin version `n-1` and Kirin version `n` are
```

migrations/versions/1f13d13948dd_rm_rtu_received_at.py
pbougue: Very true!
Nice catch, I'm glad it wasn't merged :pray: 

Nota: this is a change contained in #305 so I'm gonna remove it from there (and this PR won't be stacked over it as there won't be any db migration in #305 )

CONTRIBUTING.md
pbougue: As discussed : rephrased :)

pbougue:
@kadhikari : rebased the work directly on 'master' branch (and will split work from #305 in 2 parts).
So this should be good for this PR (once 4.0.1 deployed on prod).

PR #313, opened 2020-03-06
Rename global variable rmq_handler

woshilapin:
To avoid problem with the Python module with the same name.


buildbotctp:

woshilapin:
> tada This will conflict with GHAction PR, though.

Yep, I know. Let's merge it and I update the other one.

PR #314, opened 2020-03-09
Fix protobuf-compiler dependency

woshilapin:
It seems that the `ubuntu-latest` that Github is using is not really the latest (or maybe it is the latest LTS?). Since our management of version dependencies for the CI is not ideal anyway, I think it will cause us less problem to remove completely the version for now...

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -20,7 +20,7 @@ jobs:
             - name: Install Python Dependencies
               run: pip install --upgrade --requirement requirements_dev.txt --requirement requirements_pre-commit.txt
             - name: Install Protobuf Dependencies
-              run: sudo apt install --yes protobuf-compiler=3.6.1.3-2
+              run: sudo apt install --yes protobuf-compiler
             - name: Build Protobuf
               run: python setup.py build_pbf
             - name: Setup pre-commit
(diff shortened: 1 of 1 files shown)


pascal-vs:
Yes, for Ubuntu, "latest" is for "latest LTS": https://hub.docker.com/_/ubuntu

.github/workflows/continuous-integration.yml
```diff
@@ -20,7 +20,7 @@ jobs:
             - name: Install Python Dependencies
               run: pip install --upgrade --requirement requirements_dev.txt --requirement requirements_pre-commit.txt
             - name: Install Protobuf Dependencies
-              run: sudo apt install --yes protobuf-compiler=3.6.1.3-2
+              run: sudo apt install --yes protobuf-compiler
```
datanel: The version installed here will be `3.0.0-9.1ubuntu1`.
If you want the version installed in production (v3.3.2) you can compile from source as it is done here https://github.com/CanalTP/navitia_docker_images/blob/master/python/Dockerfile#L21 but it will increase the build time

woshilapin: Yes, I found out what was the version but I realized that making it explicit will make the build fail when the next LTS is out.

woshilapin:
Jenkins, test this please

datanel:
Jenkins, test this please

datanel:
Jenkins, test this please

woshilapin:
Jenkins, test this please

PR #320, opened 2020-09-11
Fix deadlock on migrations

pbougue:
:mag: review by commit with comment may be better.

## On the last migration

Drop foreign_key constraint before migration to avoid deadlock (put FK back after)

Could not use:
* ALTER TYPE name ADD VALUE new_enum_value
  (Migrations are transactions.
   And we wouldn't solve downgrade but we may
   just have let 'piv' live there)
* hacky modifications on system tables equivalent of ADD VALUE enum
  (rights are preventing it and it would be super-dirty)
* column add, then switch, then drop
  (deadlocks too)
* lock on ALTER COLUMN operations without FK drop
  (deadlocks too)

:heavy_check_mark: tested locally (kirin being spammed during migration)
Don't see any easy way to add tests, and it will be tested a bit more on **dev** platform (which has the same amount of data than prod ~800K real_time_update)

Notes from experiment:
* Migration is ~3s locally for 20K real_time_updates
* During lock kirin just waits: tested locking during 70s and just ended-up with a slow response. So during migration kirin should work as intended on the side too (just slower).

## On other migrations

During local tests detected similar problems, so corrected as it was easy (see each commit).

### DIFF
--- a/migrations/versions/443587af1780_reference_contributor_in_other_tables.py
+++ b/migrations/versions/443587af1780_reference_contributor_in_other_tables.py
@@ -64,9 +64,11 @@ def downgrade():
     op.drop_index("realtime_update_contributor_id_and_created_at", table_name="real_time_update")
     op.drop_index("contributor_id_idx", table_name="trip_update")
 
-    # Drop foreignKey constraints
-    op.drop_constraint("fk_trip_update_contributor_id", "trip_update", type_="foreignkey")
-    op.drop_constraint("fk_real_time_update_contributor_id", "real_time_update", type_="foreignkey")
+    # Drop foreignKey constraints (lock required to avoid deadlock from Kirin's work on the side)
+    op.execute("LOCK TABLE trip_update; ALTER TABLE trip_update DROP CONSTRAINT fk_trip_update_contributor_id")
+    op.execute(
+        "LOCK TABLE real_time_update; ALTER TABLE real_time_update DROP CONSTRAINT fk_real_time_update_contributor_id"
+    )
 
     # Delete contributor_id in real_time_update and trip_update
     op.drop_column("trip_update", "contributor_id")
--- a/migrations/versions/1911dab5a030_nullable_received_at.py
+++ b/migrations/versions/1911dab5a030_nullable_received_at.py
@@ -20,5 +20,8 @@ def upgrade():
 
 
 def downgrade():
-    op.execute("UPDATE real_time_update SET received_at = created_at WHERE received_at IS NULL;")
-    op.execute("ALTER TABLE real_time_update ALTER COLUMN received_at SET NOT NULL;")
+    op.execute(
+        "LOCK TABLE real_time_update; \
+        UPDATE real_time_update SET received_at = created_at WHERE received_at IS NULL; \
+        ALTER TABLE real_time_update ALTER COLUMN received_at SET NOT NULL;"
+    )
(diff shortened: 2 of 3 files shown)


migrations/versions/443587af1780_reference_contributor_in_other_tables.py
```diff
@@ -64,9 +64,11 @@ def downgrade():
     op.drop_index("realtime_update_contributor_id_and_created_at", table_name="real_time_update")
     op.drop_index("contributor_id_idx", table_name="trip_update")
 
-    # Drop foreignKey constraints
-    op.drop_constraint("fk_trip_update_contributor_id", "trip_update", type_="foreignkey")
-    op.drop_constraint("fk_real_time_update_contributor_id", "real_time_update", type_="foreignkey")
+    # Drop foreignKey constraints (lock required to avoid deadlock from Kirin's work on the side)
+    op.execute("LOCK TABLE trip_update; ALTER TABLE trip_update DROP CONSTRAINT fk_trip_update_contributor_id")
```
woshilapin, line 68: I had to go see the documentation to understand when the table was `UNLOCK`. And it seems to happen until the end of the transaction. I don't see any transaction here. But maybe `alembic` wrap the `upgrade`/`downgrade` operation in a transaction?

woshilapin (approved):

migrations/versions/443587af1780_reference_contributor_in_other_tables.py
pbougue: From what I tested it is wrapping in a transaction... As the lock is released after the migration (Kirin running on the side is moving on once the transaction is done).
And at least one transaction is done for migration, as it prevents from using `ADD VALUE enum`.
Until now, I wasn't able to get a clear view of what alembic is doing.

That is before I learn about the option `--sql from_revision:to_revision`.
From that it's clear: alembic adds the final `;` to each SQL command, and wraps the whole migration (either upgrade or downgrade) between `BEGIN; <migration> COMMIT;`.

Which lead me to test some other ideas... More on that in an upcoming commit.

pbougue:
OK, now that I discovered `alembic --sql from_rev:to_rev` I have 3 solutions that are **working** locally, and we have to decide which is the chosen one (by order of chronological appearance):
1. [first commit](https://github.com/CanalTP/kirin/blob/f28f79405e725c72ccdfe12034bc558a9dfa1dbc/migrations/versions/85a968ce3ce5_add_piv_connector_type.py#L18-L38) : unique transaction (totally reverted if it fails), removes foreign key (lock needed here) during `ALTER tables` then put it back, quite "slow" (3s locally for 20K real_time_update).
2. [4th commit](https://github.com/CanalTP/kirin/blob/a1f524aaa675003f62257759eb7934ab1f9e955a/migrations/versions/85a968ce3ce5_add_piv_connector_type.py#L18-L38) : multiple transactions (partially reverted if it fails), isolate `ALTER contributor` with a `LOCK` in a separate transaction, no foreign_key dropped, quite "slow" (3s locally for 20K real_time_update).
3. [5th commit](https://github.com/CanalTP/kirin/blob/6624cd842711f6cea11875a524558bd1f1eb1186/migrations/versions/85a968ce3ce5_add_piv_connector_type.py#L18-L20) : no transaction (empty transactions actually), hacks the fact that migrations are transactional, quite quick (0.8s, probably independant of db size).

Then from there we might also choose what we do on downgrade transaction: the 3. option is not available to remove an enum_value, but we migth just not clean it.

pascal-vs (approved):

woshilapin (approved):

PR #321, opened 2020-09-22
Store PIV feed

pbougue:
Store PIV raw-data input in db when POSTing it (still no processing), and add tests on that.
Also an important refactor and some cleanup.

:mag: Please review by **commit** and read **commit message first** :information_source:

JIRA: https://jira.kisio.org/browse/ND-1013


kirin/core/abstract_builder.py
```diff
@@ -0,0 +1,139 @@
...
+        """Create a wrapping object around the feed received to store it"""
+        rt_update = None
+        log_dict = {}
+        return rt_update, log_dict
+
+    def build_trip_updates(self, rt_update):
+        """Convert realtime information into Kirin's internal model"""
+        trip_updates = []
+        log_dict = {}
+        return trip_updates, log_dict
```
woshilapin, line 139: I suppose these are abstract methods. Should we enforce any class that inherits from `AbstractKirinModelBuilder` to implement these with something like `raise NotImplementedError("Please Implement this method")`?

kirin/core/abstract_builder.py
```diff
@@ -0,0 +1,139 @@
...
+        record_call(status, **log_dict)
+        if status == "OK":
+            logging.getLogger(__name__).info(status, extra=log_dict)
+        elif status == "warning":
+            logging.getLogger(__name__).warning(status, extra=log_dict)
+        else:
+            logging.getLogger(__name__).error(status, extra=log_dict)
+
+
+class AbstractKirinModelBuilder(six.with_metaclass(ABCMeta, object)):
```
woshilapin, line 107: Would it be possible to hide to the user of any `AbstractKirinModelBuilder` the fact that they have to call `wrap_build()`? Because, if I understand, you provide a nice helper `wrap_build()` that needs to be called on anything that inherits `AbstractKirinModelBuilder`. But the user might forget to call `wrap_build()`. My Python expertise is not that good but, can we use `AbstractKirinModelBuilder` to automatically call `wrap_build()`?

kirin/core/abstract_builder.py
pbougue: Fixed using type annotations and comments too in https://github.com/CanalTP/kirin/pull/321/commits/152cce30bd6ab739b1a5230b75cf21cd787e1fe0

woshilapin (approved):

kirin/core/abstract_builder.py
pbougue: It is possible indeed, and I wanted to migrate `gtfs-rt` to `wrap_build` before considering it.
I finished `gtfs-rt` migration, and will open a new PR above this one for gtfs-rt.

After the merge of those 2 PRs I'm OK to hide `wrap_build` more if anyone thinks it's worth it (didn't think of how precisely, and I'm not especially aiming to do it myself).

So I think this is not blocking, and if one wants to keep track of this, it's possible.

datanel (approved):

PR #322, opened 2020-09-25
Use wrap_build() and AbstractKirinModelBuilder for gtfs-rt

pbougue:
~Only last (5) commits since the 25th of septembre to review.~ rebased
:mag: Please review by commit with message.

TODO:

- [x] merge  #321 and rebase current branch
- [x] tested locally that cots and gtfs-rt are still working all the way to navitia

Done also:
* update tests accordingly
* some cleanup (last 4 commits)

Related to https://jira.kisio.org/browse/ND-1013

### DIFF
--- a/kirin/gtfs_rt/tasks.py
+++ b/kirin/gtfs_rt/tasks.py
@@ -152,7 +152,7 @@ def gtfs_poller(self, config):
         # If the HEAD request or Redis get/set fail, we just ignore this part and do the polling anyway
         if not _is_newer(config):
             new_relic.ignore_transaction()
-            manage_db_no_new(connector=ConnectorType.gtfs_rt.value, contributor=contributor.id)
+            manage_db_no_new(connector_type=ConnectorType.gtfs_rt.value, contributor_id=contributor.id)
             return
 
         try:
@@ -161,8 +161,8 @@ def gtfs_poller(self, config):
         except Exception as e:
             manage_db_error(
                 data="",
-                connector=ConnectorType.gtfs_rt.value,
-                contributor=contributor.id,
+                connector_type=ConnectorType.gtfs_rt.value,
+                contributor_id=contributor.id,
                 error="Http Error",
                 is_reprocess_same_data_allowed=True,
             )
--- a/kirin/gtfs_rt/tasks.py
+++ b/kirin/gtfs_rt/tasks.py
@@ -36,7 +36,9 @@ from datetime import datetime
 import requests
 import six
 
-from kirin import gtfs_realtime_pb2
+from kirin.core import model
+from kirin.core.abstract_builder import wrap_build
+from kirin.core.types import ConnectorType
 from kirin.cots.model_maker import as_duration
 
 from kirin.tasks import celery
@@ -50,12 +52,10 @@ from kirin.utils import (
     record_input_retrieval,
     make_kirin_last_call_dt_name,
 )
-from kirin.gtfs_rt import model_maker
+from kirin.gtfs_rt import KirinModelBuilder
 from retrying import retry
 from kirin import app, redis_client
 from kirin import new_relic
-from google.protobuf.message import DecodeError
-import navitia_wrapper
 
 
 TASK_STOP_MAX_DELAY = app.config[str("TASK_STOP_MAX_DELAY")]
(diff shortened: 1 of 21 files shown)


kirin/gtfs_rt/tasks.py
```diff
@@ -125,17 +125,22 @@ def _retrieve_gtfsrt(config):
 @retry(stop_max_delay=TASK_STOP_MAX_DELAY, wait_fixed=TASK_WAIT_FIXED, retry_on_exception=should_retry_exception)
 def gtfs_poller(self, config):
     func_name = "gtfs_poller"
-    logger = logging.LoggerAdapter(logging.getLogger(__name__), extra={"contributor": config["contributor"]})
+    contributor = (
+        model.Contributor.query_existing()
+        .filter_by(id=config.get("contributor"), connector_type=ConnectorType.gtfs_rt.value)
+        .first()
```
woshilapin, line 131: I'm not sure I understand why you take only the first `gtfs_rt` contributor found in a function `gtfs_poller` that feels rather generic.

tests/check_utils.py
```diff
@@ -59,12 +58,12 @@ def api_post(url, display=False, check=True, *args, **kwargs):
...
     if check:
         assert resp.status_code == 200
-        return _to_json(resp.data, display)
+        return _to_json(resp.data)
     else:
-        return _to_json(resp.data, display), resp.status_code
+        return _to_json(resp.data), resp.status_code
 
 
-def _to_json(data, display):
```
woshilapin, line 67: Nice clean up.

tests/integration/gtfs_rt_test.py
```diff
@@ -215,35 +219,32 @@ def test_gtfs_model_builder(basic_gtfs_rt_data, basic_gtfs_rt_data_without_delay
         first_stop = trip_updates[0].stop_time_updates[0]
         assert first_stop.stop_id == "StopR1"
         assert first_stop.arrival_status == "none"
-        assert first_stop.arrival_delay is None
-        assert first_stop.departure_delay is None
+        assert first_stop.arrival_delay == timedelta(minutes=0)
```
woshilapin, line 222: :open_mouth: made me think of that https://github.com/CanalTP/tartare-tools/pull/145#discussion_r494168388 but maybe it's different here...

tests/integration/test_end_point.py
```diff
@@ -198,8 +197,8 @@ def test_status_with_database_ko(setup_database):
         assert len(resp["last_update"]) == 0
         assert len(resp["last_update_error"]) == 0
         assert len(resp["last_valid_update"]) == 0
-        resp["db_pool_status"] is None
-        resp["db_version"] is None
+        assert resp["db_pool_status"] is None
+        assert resp["db_version"] is None
```
woshilapin, line 201: This is probably much better... :open_mouth:

woshilapin (commented):
In this (huge!) PR, some general comments.

# ConnectorType
I'm seeing a lot of use of `ConnectorType.gtfs_rt.value` replacing a simple string. This `ConnectorType` has been existing for a long time. I'm assuming that what you did is improving something that should have been done a long time ago?

# `contributor` -> `contributor.id`
I'm seeing a lot of changes from `contributor` to `contributor.id` in function calls... but I can't find the corresponding changes in the functions themselves so this sounds really fishy to me. At the very least, I would expect the functions to have their argument changed from `contributor` to `contributor_id`.

kirin/gtfs_rt/tasks.py
pbougue: The function returns a list (it's a filter), but in the end we only get one, as id is a key.
So we take the only one we get (if any, as "id" provided might be wrong, and this behavior is tested here : https://github.com/CanalTP/kirin/blob/master/tests/integration/gtfs_rt_test.py#L185).

woshilapin: Oh, you're filtering by `id` and by `connector_type`. I see.

tests/integration/gtfs_rt_test.py
pbougue: Indeed :smile: 
Humm, not digging into the code, but I guess we just have to fill it at some point (at least for protobuf in the end?) and it was decided to do it explicitly at that point (in handler) :man_shrugging: 
It's OK to me.

pbougue:
@woshilapin indeed this PR shaves some yaks (even more after you review :stuck_out_tongue:)

On the `ConnectorType`: true, it should have been done a long ago (the enum appear after the use, to explain previous state).

On the `contributor` -> `contributor.id` changes:
Why:
* a lot of functions were using `contributor` params that were actually refering to `Contributor.id` (Contributor object appeared lately)
* in the first commit, the `contributor` param of `wrap_build()` is an actual `Contributor` object, which wasn't the case before.

Improvements after your review:
* renamed a lot of params, see https://github.com/CanalTP/kirin/pull/322/commits/1924b36d11e5d3ab00efde7d96db7c3c5dc2020a (and check commit's message :wink: )

woshilapin (approved):

datanel (approved):

woshilapin (approved):

PR #324, opened 2020-09-29
[ci] Put a fixed version for black

woshilapin:
- Stabilize the Black version so there is no different version between local development and the CI.
- Removing the `requirements_pre-commit.txt` which is documented nowhere and is not useful since `pre-commit` will install whatever is needed automatically


datanel (dismissed):

woshilapin:
> travis needs requirements_pre-commit.txt

Indeed but actually, we do not really need this file since `pre-commit install` is enough. I just moved the dependency for `pre-commit` into `requirements_dev.txt` and removed the file `requirements_pre-commit.txt`.

datanel (approved):

PR #325, opened 2020-09-30
Clarify precommit use

pbougue:
After first step in https://github.com/CanalTP/kirin/pull/324

As it requires a dedicated python 3.6+ virtualenv (hoping doc is clearer on that now)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,13 +18,13 @@ jobs:
             - name: Upgrade pip
               run: python -m pip install --upgrade pip
             - name: Install Python Dependencies
-              run: pip install --upgrade --requirement requirements_dev.txt
+              run: pip install --upgrade --requirement requirements.txt
             - name: Install Protobuf Dependencies
               run: sudo apt install --yes protobuf-compiler
             - name: Build Protobuf
               run: python setup.py build_pbf
             - name: Setup pre-commit
-              run: pre-commit install
+              run: pip install --upgrade pre-commit && pre-commit install
             - name: Quality Check
               run: pre-commit run --all --show-diff-on-failure
             - name: Code Linting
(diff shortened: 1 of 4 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -18,13 +18,13 @@ jobs:
...
             - name: Install Python Dependencies
-              run: pip install --upgrade --requirement requirements_dev.txt
+              run: pip install --upgrade --requirement requirements.txt
             - name: Install Protobuf Dependencies
               run: sudo apt install --yes protobuf-compiler
             - name: Build Protobuf
               run: python setup.py build_pbf
             - name: Setup pre-commit
-              run: pre-commit install
+              run: pip install --upgrade pre-commit && pre-commit install
```
woshilapin, line 27: I don't think `--upgrade` is useful in a CI environment since everything is installed from scratch anyway. From what I understand, `--upgrade` is useful only if the package has already been installed previously with an older version.

CONTRIBUTING.md
```diff
@@ -23,33 +23,34 @@ When absolutely needed, datetimes that are not UTC **must** be timezone-aware to
 being able to mix them with naive-UTC-datetimes (it's also the easiest and safest way to handle timezones).\
 Communication (request and read) with external services is done in UTC as much as possible (with Navitia especially).
 
-### Python formatting
+### Git hooks
 
-Python source code in this project is formatted with [Black](https://black.readthedocs.io/en/stable/)
-You should enable the pre-commit git hook to be sure. It's also the easier way to run black, you can simply run:
+The project provides a few git hooks that you should use to prevent any issues.
+Attention: they require python 3.6+ to run.
```
woshilapin, line 29: Not sure.
```suggestion
Warning: they require python 3.6+ to run.
```

woshilapin (approved):

.github/workflows/continuous-integration.yml
pbougue: Yep, I let it in case we keep cache between runs and doesn't hurt anyway (as it was there on the previous line too).

woshilapin (approved):

patochectp (approved):

PR #331, opened 2020-10-13
Update contributors status 

pbougue:
:mag: please review by commit, should be easier

* consider all types (including `piv` which was not displayed)
* do not consider config file as it's now deprecated and unused everywhere : https://jira.kisio.org/browse/ND-842
* update and add tests

TODO:

- [x] propagate config removal to fabric and conf repositories


kirin/core/model.py
```diff
@@ -456,49 +456,26 @@ def get_probes_by_contributor(cls):
...
-        for c in contributors:
+        contributor_ids = [contributor.id for contributor in Contributor.query_existing().all()]
+        for c_id in contributor_ids:
             sql = db.session.query(cls.created_at, cls.status, cls.updated_at, cls.error)
-            sql = sql.filter(cls.contributor_id == c)
+            sql = sql.filter(cls.contributor_id == c_id)
             sql = sql.order_by(desc(cls.created_at))
             row = sql.first()
             if row:
                 date = row[2] if row[2] else row[0]  # update if exist, otherwise created
```
woshilapin, line 468: Aren't we suppose to have the `updated_at` and `created_at` at the same value when the row is created? Corollary, `updated_at` is never empty...

woshilapin (approved):
A bunch of code removed, :neutral_face: :smiley: :smile: :joy: :hand:

kirin/core/model.py
pbougue: Indeed, but it's a "recent" change, not deployed on prod: #305.
So for so time this code is still a bit useful (as long as no new feed as been received on a contributor after deploying the version with #305).

We may do a separate PR for that, to be merge after `4.1` deployment (maybe some other things to move concerning that, but probably not much). I'm lazy on that one...

woshilapin (approved):

pbougue:
Jenkins, test this please.

patochectp (approved):

PR #332, opened 2020-10-13
[test] Add tests for PivWorker

woshilapin:
A few things to notice:
- The tests are only testing `kirin.command.piv_worker.PivWorker` meaning that the function `kirin.command.piv_worker.piv_worker` is not tested (this function is an infinite loop that would need to be launch into a thread and then never dies... therefore the test would hang up forever)
- PivWorkerTest exists in order to always clean up the thread at the end of the test (much better since when the test fails, raising an `AssertionError`, the thread is still destroyed)
- Testing that the RabbitMQ container is up has been completely rewritten
- Now, the docker wrapper for RabbitMQ create MQ handler on-demand (before, it was creating one and only one at initialization)
- Since these tests involve multi-threading, some of the assertions can't be done synchronously, therefore, I use a `Retrying` object to try asserting until the other thread has done the expected work

I was not able to create a test for a change of `broker_url`, any help is welcome but in the worst case, there might not be any case for that.

ref: ND-1066

### DIFF
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -75,17 +75,19 @@ class PivWorker(ConsumerMixin):
     def __enter__(self):
         self.connection = Connection(self.builder.contributor.broker_url)
         self.exchange = self._get_exchange(self.builder.contributor.exchange_name)
-        self.queue = self._get_queue(self.exchange, self.builder.contributor.queue_name)
+        self.queue = self._get_or_create_queue(self.builder.contributor.queue_name)
         return self
 
     def __exit__(self, type, value, traceback):
         self.connection.release()
 
     def _get_exchange(self, exchange_name):
-        return Exchange(exchange_name, "fanout", durable=True, no_declare=True)
+        return Exchange(name=exchange_name, type="fanout", durable=True, no_declare=True, auto_delete=False)
 
-    def _get_queue(self, exchange, queue_name):
-        return Queue(queue_name, exchange, durable=True, auto_delete=False)
+    def _get_or_create_queue(self, queue_name):
+        queue = Queue(name=queue_name, exchange=self.exchange, durable=True, auto_delete=False)
+        queue.declare(channel=self.connection.channel())
+        return queue
 
     def get_consumers(self, Consumer, channel):
         return [
@@ -112,27 +114,19 @@ class PivWorker(ConsumerMixin):
             db.session.expire(self.builder.contributor, ["broker_url", "exchange_name", "queue_name"])
             self.last_config_checked_time = datetime.now()
         contributor = get_piv_contributor(self.builder.contributor.id)
-        if not contributor:
+        if (
+            not contributor
+            or contributor.broker_url != self.broker_url
+            or contributor.exchange_name != self.exchange.name
+            or contributor.queue_name != self.queue.name
+        ):
             logger.info(
-                "contributor '{0}' doesn't exist anymore, let the worker die".format(self.builder.contributor.id)
+                "configuration of contributor '{0}' changed, let the worker die".format(
+                    self.builder.contributor.id
+                )
             )
             self.should_stop = True
             return
-        if contributor.broker_url != self.broker_url:
-            logger.info("broker URL for contributor '{0}' changed, let the worker die".format(contributor.id))
-            self.should_stop = True
-            return
-        if contributor.exchange_name != self.exchange.name:
-            logger.info(
-                "exchange name for contributor '{0}' changed, worker updated".format(contributor.exchange_name)
-            )
-            self.exchange = self._get_exchange(contributor.exchange_name)
-            self.queue = self._get_queue(self.exchange, contributor.queue_name)
-        if contributor.queue_name != self.queue.name:
-            logger.info(
-                "queue name for contributor '{0}' changed, worker updated".format(contributor.queue_name)
-            )
-            self.queue = self._get_queue(self.exchange, contributor.queue_name)
         self.builder = KirinModelBuilder(contributor)
 
 
--- a/kirin/rabbitmq_handler.py
+++ b/kirin/rabbitmq_handler.py
@@ -133,10 +133,12 @@ class RTReloader(ConsumerProducerMixin):
 
 
 class RabbitMQHandler(object):
-    def __init__(self, connection_string, exchange):
+    def __init__(self, connection_string, exchange_name, exchange_type="topic"):
         self._connection = BrokerConnection(connection_string)
         self._connections = {self._connection}  # set of connection for the heartbeat
-        self._exchange = Exchange(exchange, durable=True, delivery_mode=2, type="topic")
+        self._exchange = Exchange(
+            exchange_name, durable=True, delivery_mode=2, type=exchange_type, auto_delete=False, no_declare=False
+        )
         monitor_heartbeats(self._connections)
 
     @retry(wait_fixed=200, stop_max_attempt_number=3)
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -42,6 +42,8 @@ from kirin.core.types import ConnectorType
 
 COTS_CONTRIBUTOR_ID = "rt.tchoutchou"
 PIV_CONTRIBUTOR_ID = "rt.piv"
+PIV_EXCHANGE_NAME = "piv_schedule"
+PIV_QUEUE_NAME = "piv_schedule_kirin"
 GTFS_CONTRIBUTOR_ID = "rt.vroumvroum"
 COTS_CONTRIBUTOR_DB_ID = "rt.tchoutchou_db"
 PIV_CONTRIBUTOR_DB_ID = "rt.piv_db"
@@ -68,7 +70,8 @@ def bdd(init_flask_db):
         flask_migrate.downgrade(revision="base", directory=migration_dir)
 
 
-def clean_db():
+@pytest.fixture(scope="function", autouse=True)
+def clean_db(rabbitmq_docker_fixture):
     """
     before all tests the database is cleared
     """
@@ -94,9 +97,12 @@ def clean_db():
                     "sncf_piv",
                     ConnectorType.piv.value,
                     "piv_token",
-                    "piv_feed_url",
-                    10,
+                    None,
+                    None,
                     True,
+                    rabbitmq_docker_fixture.url,
+                    PIV_EXCHANGE_NAME,
+                    PIV_QUEUE_NAME,
                 ),
                 model.Contributor(
                     GTFS_CONTRIBUTOR_ID,
@@ -111,7 +117,16 @@ def clean_db():
                     COTS_CONTRIBUTOR_DB_ID, "idfm", ConnectorType.cots.value, "cots_db_token", "cots_db_feed_url"
                 ),
                 model.Contributor(
-                    PIV_CONTRIBUTOR_DB_ID, "tn", ConnectorType.piv.value, "piv_db_token", "piv_db_feed_url"
+                    PIV_CONTRIBUTOR_DB_ID,
+                    "tn",
+                    ConnectorType.piv.value,
+                    "piv_db_token",
+                    None,
+                    None,
+                    True,
+                    rabbitmq_docker_fixture.url,
+                    PIV_EXCHANGE_NAME,
+                    PIV_QUEUE_NAME,
                 ),
                 model.Contributor(
                     GTFS_CONTRIBUTOR_DB_ID,
@@ -124,11 +139,6 @@ def clean_db():
             ]
         )
         db.session.commit()
-
-
-@pytest.fixture(scope="function", autouse=True)
-def clean_db_fixture():
-    clean_db()
 
 
 @pytest.fixture(scope="function")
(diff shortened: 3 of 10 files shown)


patochectp:
jenkins test this please.

patochectp (approved):

woshilapin:
After PR #331 was merged, I resolved merged conflicts.

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+
+def test_mq_message_received_and_stored(
+    test_client, pg_docker_fixture, rabbitmq_docker_fixture, broker_connection, mq_handler
+):
+    with PivWorkerTest(test_client, rabbitmq_docker_fixture.url, pg_docker_fixture):
+        # Check that PivWorker is creating the queue
+        wait_until.call(lambda: fassert(is_queue_exists(broker_connection, queue_name)))
+
+        # Check that MQ message is received and stored in DB
+        mq_handler.publish(str('{"key": "Some valid JSON"}'), contributor_id)
```
pbougue, line 168: This `contributor_id` here makes me wonder if we should have another `publish()` method.
As it's not useful at all in that case, no?

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+        exchange = Exchange(
+            new_exchange_name,
+            durable=True,
+            delivery_mode=2,
+            type="fanout",
+            auto_delete=False,
+            no_declare=False,
+        )
+        exchange.declare(channel=broker_connection.channel())
+        assert is_exchange_exists(broker_connection, new_exchange_name)
```
pbougue, line 194: Can't you use init_exchange() here?

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+
+def test_exchange_changed(test_client, pg_docker_fixture, rabbitmq_docker_fixture, broker_connection):
+    with PivWorkerTest(test_client, rabbitmq_docker_fixture.url, pg_docker_fixture):
+        # Modify the 'exchange_name' in the contributor
+        new_exchange_name = "pivert"
+        modified_contrib = {
+            "id": contributor_id,
+            "exchange_name": new_exchange_name,
+        }
+        response = test_client.put("/contributors", json=modified_contrib)
```
pbougue, line 180: Minor: to me the "happy path" is creating exchange outside first, then branching PivWorker.

Otherwise, I expect the worker to check is exchange exist and retry until it does.
So it's OK to test like that, but it tests a bit more than happy-path.

TODO (probably): separate discussion to see if it should be checked, tested and all.

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+        assert response.status_code == 200
+
+        # Check the PivWorker did create the new queue
+        wait_until.call(lambda: fassert(is_queue_exists(broker_connection, new_queue_name)))
+
+        # Delete the old queue
+        channel = broker_connection.channel()
+        channel.queue_delete(queue_name)
+        fassert(not is_queue_exists(broker_connection, queue_name))
```
pbougue, line 224: more common to read
```suggestion
        assert not is_queue_exists(broker_connection, queue_name)
```

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def fassert(test):
+    """
+    `assert` is not supported into `lambda` function.
+    `lambda: assert True == False` is a syntax error.
+    With this function, you can replace it with
+    `lambda: fassert(True == False)`
+    """
+    assert test
```
pbougue, line 59: Not sure it's as easy: as you always use them together, can't you merge `wait_until` and `fassert` ?

```suggestion
@retry(stop_max_delay=20000, wait_exponential_multiplier=100)
def wait_until(test):
    assert test
```

pbougue (commented):
:clap: Really great and compact (given what is done) testing.
Not easy to read at first with `__init__`, `__enter__` and `__exit__`, but it's so much better when you have 2 tests... :tada:

kirin/rabbitmq_handler.py
```diff
@@ -133,10 +133,12 @@ def _on_request(self, message):
 
 
 class RabbitMQHandler(object):
-    def __init__(self, connection_string, exchange):
+    def __init__(self, connection_string, exchange_name, exchange_type="topic"):
         self._connection = BrokerConnection(connection_string)
         self._connections = {self._connection}  # set of connection for the heartbeat
-        self._exchange = Exchange(exchange, durable=True, delivery_mode=2, type="topic")
+        self._exchange = Exchange(
+            exchange_name, durable=True, delivery_mode=2, type=exchange_type, auto_delete=False, no_declare=False
```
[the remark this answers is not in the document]
woshilapin: I'm not sure I understand your point. There IS a `durable=True` already.

tests/integration/piv_worker_test.py
woshilapin: Might be done in another PR.

tests/integration/piv_worker_test.py
woshilapin: Might be done in another PR. I believe the `PivWorker` is not resilient (and most likely fails silently) if the `Exchange` doesn't exist before-hand.

kirin/rabbitmq_handler.py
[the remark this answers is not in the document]
pbougue: :smile: Indeed, that's a issue with my eyes :eyeglasses:

kirin/command/piv_worker.py
```diff
@@ -75,17 +75,17 @@ def __init__(self, contributor):
...
     def __exit__(self, type, value, traceback):
         self.connection.release()
 
     def _get_exchange(self, exchange_name):
-        return Exchange(exchange_name, "fanout", durable=True, no_declare=True)
+        return Exchange(name=exchange_name, type="fanout", durable=True, no_declare=True, auto_delete=False)
 
-    def _get_queue(self, exchange, queue_name):
-        return Queue(queue_name, exchange, durable=True, auto_delete=False)
+    def _create_queue(self, queue_name):
```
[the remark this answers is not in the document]
woshilapin: Done in d987d417f2516aae21c4077d4fbd6f786ba42f8f.

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+    try:
+        channel = connection.channel()
+        channel.queue_declare(queue=queue_name, nowait=False, passive=True)
+        return True
+    except NotFound as e:
+        return False
+    return False
+
+
+@pytest.yield_fixture(scope="session", autouse=True)
```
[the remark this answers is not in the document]
woshilapin: It does make sense and it is doable. Done in 8a2561c97c1e9020a6120ed818f9ecc2eca75b5c.

kirin/command/piv_worker.py
```diff
@@ -119,20 +119,28 @@ def on_iteration(self):
...
+                "broker URL '{1}' for contributor '{0}' changed, let the worker die".format(
+                    contributor.id, contributor.broker_url
+                )
+            )
             self.should_stop = True
             return
         if contributor.exchange_name != self.exchange.name:
             logger.info(
-                "exchange name for contributor '{0}' changed, worker updated".format(contributor.exchange_name)
+                "exchange name '{1}' for contributor '{0}' changed, worker updated".format(
```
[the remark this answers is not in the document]
woshilapin: Done in 8a0a7949ad4a9602bc7d39f0651f7d724a4d6eec.

tests/integration/piv_worker_test.py
woshilapin: Much better indeed although, I had to tweak it a little bit: `wait_until` function can't take a boolean but must take a function because it needs to be re-evaluated on every retry. Done in 7caa2336073dd4d08a5b354276066f064a81cbbc.

tests/integration/piv_worker_test.py
woshilapin: Done in 7caa2336073dd4d08a5b354276066f064a81cbbc.

woshilapin:
I also renamed `is_exchange_exists()` and `is_queue_exists()` into `is_exchange_created()` and `is_queue_created()` in 5831e8e474fc41d6efa1ea49ce49e9a654eaabd8.

tests/integration/piv_worker_test.py
woshilapin: Not really because the exchange's name is different and the function `init_exchange()` doesn't take an `exchange_name` as an input parameter. And since `init_exchange()` is a fixture function, I cannot really change the signature of this function to take an additional parameter that is not another fixture...

That said, I did it by extracting a `create_exchange()` function (see 50e2b78f38998d0aab92833b3f36e5667cb6db43).

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def is_exchange_exists(connection, exchange_name, exchange_type="fanout"):
+    try:
+        channel = connection.channel()
+        channel.exchange_declare(exchange_name, exchange_type, nowait=False, passive=True)
+        return True
+    except NotFound as e:
+        return False
+    return False
```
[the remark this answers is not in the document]
woshilapin: Much better indeed and solves the thing that was itching me with this function. Done in f301a794b922e7fa44cb6bf68636ddda3a7802d4. But then I realized that catching other `Exception` and raising them again would be better. So I also did 37d31c8e0d93a91ff6e349a4fa667b9374039681.

tests/integration/piv_worker_test.py
woshilapin: I did change the test though. See 648cd24adfbf45b0e701326013f32cd89dfe2c7a.

woshilapin:
I did the following modifications to the code:
- ad878bf7b7fb4e3d6c01c574be0b57ce6b9e11fa Now the `PivWorker` is killed for any configuration change (trying to hot reload `Exchange` and `Queue` was buggy)
- 3be22088eeb6238ecf0835bb6c72b4401457f367 Removed non-working tests (we'll put some manual testing in place)
- ca248bf564db3986b7acafba45c7708beb371648 Use existing `Contributor` instead of creating a new one for `PivWorker` tests
- af879882a13f3f653c0bfa5c351d6912045b206a Do not use `clean_db()` as a function anymore (should be a fixture and used only as a fixture).

tests/integration/piv_worker_test.py
woshilapin: Done

tests/integration/piv_worker_test.py
woshilapin: Done

kirin/command/piv_worker.py
```diff
@@ -119,20 +121,29 @@ def on_iteration(self):
...
+                )
             )
-            self.exchange = self._get_exchange(contributor.exchange_name)
-            self.queue = self._get_queue(self.exchange, contributor.queue_name)
+            self.should_stop = True
+            return
         if contributor.queue_name != self.queue.name:
             logger.info(
-                "queue name for contributor '{0}' changed, worker updated".format(contributor.queue_name)
+                "queue name '{1}' for contributor '{0}' changed, worker updated".format(
```
pbougue, line 141: As actions are the same, can you mutualize test and log too?

tests/integration/conftest.py
```diff
@@ -42,6 +42,8 @@
 
 COTS_CONTRIBUTOR_ID = "rt.tchoutchou"
 PIV_CONTRIBUTOR_ID = "rt.piv"
+PIV_EXCHANGE_NAME = "piv"
+PIV_QUEUE_NAME = "piv_kirin"
```
pbougue, line 46: To match naming decided:
```suggestion
PIV_EXCHANGE_NAME = "piv_schedule"
PIV_QUEUE_NAME = "piv_schedule_kirin"
```

tests/integration/piv_worker_test.py
```diff
@@ -0,0 +1,149 @@
...
+        auto_delete=False,
+        no_declare=False,
+    )
+    exchange.declare(channel=broker_connection.channel())
+
+
+# Use scope 'function' so the Exchange is recreated for every test.
+# It is useful because some tests are deleting the Exchange.
+@pytest.fixture(scope="function", autouse=True)
+def init_exchange(broker_connection):
```
pbougue, line 99:
```suggestion
def init_piv_exchange(broker_connection):
```

pbougue (commented):
And now, the end is near :notes:

kirin/command/piv_worker.py
woshilapin: Done in e493738d2f5f85947ae2ad98109e3f46dbf913f4.

tests/integration/conftest.py
woshilapin: Done in 4f5b6898bd4488427323e888838cbff87ae0a97c.

tests/integration/piv_worker_test.py
woshilapin: Done in 9b992effa3428e186bdfa44d183706523da3dd02.

woshilapin:
Note, I also rebased and squashed everything so it should be ready to merge.

pbougue (approved):

woshilapin:
Jenkins, test this please

PR #336, opened 2020-10-30
[feature] piv: trip removal

patochectp:
ref: ND-1085


kirin/piv/model_maker.py
```diff
@@ -269,15 +297,23 @@ def _get_navitia_vj(self, piv_key, ads):
...
         trip_update = model.TripUpdate(vj=vj, contributor_id=self.contributor.id)
         trip_update.headsign = get_value(json_train, "numero", nullable=True)
         company_id = get_value(get_value(json_train, "operateur"), "codeOperateur")
         trip_update.company_id = self._get_navitia_company(company_id)
         physical_mode = get_value(get_value(json_train, "modeTransport"), "typeMode")
         trip_update.physical_mode_id = self._get_navitia_physical_mode(physical_mode)
+        trip_update.message = trip_status[1]
+
+        # TODO: handle status/effect for creation, detour, back-to-normal
+        if trip_status[0].value == 2:  # SUPPRESSION or SUPPRESSION_TOTALE
```
woshilapin, line 310: How about this instead?
```suggestion
        if trip_status[0] == TripStatus.SUPPRESSION or trip_status[0] == TripStatus.SUPPRESSION_TOTALE:
```

kirin/piv/model_maker.py
```diff
@@ -269,15 +297,23 @@ def _get_navitia_vj(self, piv_key, ads):
...
-    def _make_trip_update(self, json_train, vj):
+    def _make_trip_update(self, json_train, vj, trip_status):
         trip_update = model.TripUpdate(vj=vj, contributor_id=self.contributor.id)
         trip_update.headsign = get_value(json_train, "numero", nullable=True)
         company_id = get_value(get_value(json_train, "operateur"), "codeOperateur")
         trip_update.company_id = self._get_navitia_company(company_id)
         physical_mode = get_value(get_value(json_train, "modeTransport"), "typeMode")
         trip_update.physical_mode_id = self._get_navitia_physical_mode(physical_mode)
+        trip_update.message = trip_status[1]
+
```
woshilapin, line 308: Maybe you can add a `trip_status_type` variable because we're gonna need it more than once in the below code.

```python
        trip_status_type = trip_status[0]
```

pbougue (commented):
Sorry for the mess with remarks :-/

kirin/piv/model_maker.py
```diff
@@ -51,6 +52,19 @@
 STOP_EVENT_DATETIME_MAP = {"arrivee": "arrival", "depart": "departure"}
 
 
+class TripStatus(Enum):
+    CREATION = 1
+    SUPPRESSION = 2
+    SUPPRESSION_TOTALE = 2
+    MODIFICATION_DETOURNEMENT = 3
+    MODIFICATION_DESSERTE_SUPPRIMEE = 4
+    MODIFICATION_LIMITATION = 4
```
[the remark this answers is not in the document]
patochectp: https://cpython-test-docs.readthedocs.io/en/latest/library/enum.html: 8.13.4. Duplicating enum members and values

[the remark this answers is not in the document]
pbougue: Can't we end up with problems when enumerating all cases or searching `in` (both ways) ?
This is an explicit import of 3.4 version (requirements.txt), so OK for the doc.

kirin/piv/model_maker.py
```diff
@@ -197,11 +211,22 @@ def build_trip_updates(self, rt_update):
             raise InvalidArguments('No object "evenement" or "planTransportSource" available in feed provided')
 
         if events:
-            event_types = [get_value(event, "type", nullable=True) for event in events]
-            if "RETARD" not in event_types:
-                raise UnauthorizedValue("None of the event-types {} are supported".format(event_types))
+            managed_events = [TripStatus.RETARD, TripStatus.SUPPRESSION]
```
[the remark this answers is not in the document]
pbougue: This could match naming of Exception if it changes: so a vote for `SUPPORTED_EVENTS`.

tests/integration/piv_test.py
```diff
@@ -256,3 +290,27 @@ def test_piv_delayed(mock_rabbitmq):
...
+
+def test_piv_trip_removal_simple_post(mock_rabbitmq):
+    """
+    simple trip removal post
+    """
+    piv_feed = get_fixture_data("piv/stomp_20201029_841252_trip_removal.json")
+    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    assert "PIV feed processed" in res.get("message")
+
+    _check_db_stomp_20201029_841252_trip_removal()
```
woshilapin, line 303: Any reason why these tests are split into 2 functions? I know there was already one like this but I never noticed before. Now, I'm wondering...

tests/integration/piv_test.py
```diff
@@ -244,6 +244,40 @@ def _check_db_stomp_20201022_23186_delayed_5min():
...
+        assert db_trip_removal.vj.navitia_trip_id == "PIVPP:2020-10-29:841252:1187:Train"
+        assert db_trip_removal.vj.start_timestamp == datetime(2020, 10, 29, 20, 5)
+        assert db_trip_removal.status == "delete"
+        assert db_trip_removal.effect == "NO_SERVICE"
+        assert db_trip_removal.message == "Indisponibilité d'un matériel"
+        # full trip removal : no stop_time to precise
+        assert len(db_trip_removal.stop_time_updates) == 0
+
+
+def _check_db_stomp_20201022_23186_removal():
```
woshilapin, line 264: I would use `assert` in the name of the function instead of `check`, it conveys better the idea the function will panic with an assertion exception (`check` could return a boolean, raise another kind of exception).

tests/integration/piv_test.py
patochectp: done cd2b5b04e47c2274847ee0d89d9ce2d2c466a90c

tests/integration/piv_test.py
woshilapin: Since you inlined the function, this comment doesn't apply anymore. I'm resolving.

woshilapin (approved):

kirin/piv/model_maker.py
```diff
@@ -269,15 +297,23 @@ def _get_navitia_vj(self, piv_key, ads):
             raise ObjectNotFound("no train found for key {}".format(piv_key))
         return vj
 
-    def _make_trip_update(self, json_train, vj):
+    def _make_trip_update(self, json_train, vj, trip_status):
```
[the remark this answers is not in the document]
pbougue: I don't like redundancy here.
But we have to pay attention to one thing if we amend json_train: avoid amending raw_data (we wouldn't be able to trust it for debug). And this might already be the case for `listeArretsDesserte`.

kirin/piv/model_maker.py
```diff
@@ -193,12 +221,23 @@ def build_trip_updates(self, rt_update):
...
+                if event_type and event_type in trip_piv_status_to_effect and event_type in MANAGED_EVENTS:
+                    trip_effect = trip_piv_status_to_effect[event_type]
+                    event_label = get_value(event, "texte", nullable=True)
+                    event_struct = (
+                        (trip_effect, event_label)
+                        if trip_effect_order[trip_effect] < trip_effect_order[event_struct[0]]
+                        else event_struct
+                    )
+
+            if event_struct[0] == TripEffect.UNKNOWN_EFFECT:
```
pbougue, line 237: As this is the value for "back-to-normal", using it like that prevents from dealing that king of feed.
That probably imply storing a default `None` or the `trip_effect_order` defaulted to `sys.maxint`.

kirin/piv/model_maker.py
```diff
@@ -266,17 +306,25 @@ def _get_navitia_vj(self, piv_key, ads):
...
         company_id = get_value(get_value(json_train, "operateur"), "codeOperateur")
         trip_update.company_id = self._get_navitia_company(company_id)
         physical_mode = get_value(get_value(json_train, "modeTransport"), "typeMode")
         trip_update.physical_mode_id = self._get_navitia_physical_mode(physical_mode)
+        trip_status_type = trip_status[0]
+        trip_update.message = trip_status[1]
+        trip_update.effect = trip_status_type.name
+
+        # TODO: handle status/effect for creation, detour, back-to-normal
+        if trip_status_type == TripEffect.NO_SERVICE:  # SUPPRESSION or SUPPRESSION_TOTALE
```
pbougue, line 321: `SUPPRESSION_TOTALE` is not a possible value anymore it seems, and all the comment can be removed I think:
```suggestion
        if trip_status_type == TripEffect.NO_SERVICE:
```

kirin/piv/model_maker.py
```diff
@@ -193,12 +221,23 @@ def build_trip_updates(self, rt_update):
         if not events and not plan_transport_source:
             raise InvalidArguments('No object "evenement" or "planTransportSource" available in feed provided')
 
+        event_struct = (TripEffect.UNKNOWN_EFFECT, "")
```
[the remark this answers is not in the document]
patochectp: me too but I didn’t know `collection.namedtuple`

kirin/piv/model_maker.py
```diff
@@ -221,24 +225,23 @@ def build_trip_updates(self, rt_update):
         if not events and not plan_transport_source:
             raise InvalidArguments('No object "evenement" or "planTransportSource" available in feed provided')
 
-        event_struct = (TripEffect.UNKNOWN_EFFECT, "")
+        higher_event = ujson.loads('{"type": "UNDEFINED", "texte": ""}')
```
woshilapin, line 228: Any reason to use a JSON loader instead of defining a simple dictionary in Python?

woshilapin (approved):

kirin/piv/model_maker.py
```diff
@@ -51,6 +52,37 @@
...
+    TripEffect.SIGNIFICANT_DELAYS: 5,
+    TripEffect.UNKNOWN_EFFECT: 6,
+    TripEffect.UNDEFINED: maxint,
+}
+
+MANAGED_EVENTS = ["RETARD", "SUPPRESSION"]
+
+
+def _get_trip_effect_order_from_piv_status(status):
+    return trip_effect_order[trip_piv_status_to_effect[status]]
```
pbougue, line 83: For safety, as we probably don't want to crash on those cases:
```suggestion
    return trip_effect_order.get(trip_piv_status_to_effect.get(status, TripEffect.UNDEFINED), maxint)
```

kirin/piv/model_maker.py
```diff
@@ -193,13 +225,23 @@ def build_trip_updates(self, rt_update):
...
+                        < _get_trip_effect_order_from_piv_status(get_value(higher_event, "type"))
+                        else higher_event
+                    )
+            if trip_piv_status_to_effect[higher_event.get("type")] == TripEffect.UNDEFINED:
+                raise UnsupportedValue("None of the event-types {} are supported".format(events))
         elif plan_transport_source and plan_transport_source not in ["PTP", "OPE"]:
-            raise UnauthorizedValue("planTransportSource {} is not supported".format(plan_transport_source))
+            raise UnsupportedValue("planTransportSource {} is not supported".format(plan_transport_source))
 
+        json_train["evenement"] = higher_event
```
pbougue, line 244: To keep the same struct, as it's a list initially :
```suggestion
        json_train["evenement"] = [higher_event]
```
And a `[0]` would be required when using it after.

kirin/piv/model_maker.py
```diff
@@ -268,15 +311,23 @@ def _get_navitia_vj(self, piv_key, ads):
 
     def _make_trip_update(self, json_train, vj):
         trip_update = model.TripUpdate(vj=vj, contributor_id=self.contributor.id)
-        trip_update.headsign = get_value(json_train, "numero", nullable=True)
-        company_id = get_value(get_value(json_train, "operateur"), "codeOperateur")
+        trip_update.headsign = json_train.get("numero")
+        company_id = json_train.get("operateur").get("codeOperateur")
```
pbougue, line 315: You change the behavior here : we are supposed to break and raise InvalidArguments if we don't have those fields (so that error is correctly handled up the stack). Same on all non-nullable `get_value`.

patochectp: the check is made before

pbougue (approved):

PR #337, opened 2020-10-30
Add newrelic to PIV worker

pbougue:
Also:
* improve contributor's configuration reload
* improve robustness to exceptions when processing feed
* minor improvements
* minor fixes

:mag: review by commit with messages is highly recommended

:heavy_check_mark: Hand-tested that newrelic displays informations about PIV worker

JIRA: https://jira.kisio.org/browse/ND-1078

### DIFF
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -29,7 +29,7 @@
 # [matrix] channel #navitia:matrix.org (https://app.element.io/#/room/#navitia:matrix.org)
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
-from kirin import manager, app
+from kirin import manager, app, new_relic
 from kirin.core.model import db
 from kirin.core.types import ConnectorType
 from kirin.core.abstract_builder import wrap_build
@@ -54,6 +54,7 @@ CONF_RELOAD_INTERVAL = timedelta(
 
 
 class PivWorker(ConsumerMixin):
+    @new_relic.agent.background_task(name="piv_worker-init", group="Task")
     def __init__(self, contributor):
         if contributor.connector_type != ConnectorType.piv.value:
             raise ConfigurationError(
@@ -82,6 +83,7 @@ class PivWorker(ConsumerMixin):
         self.navitia_coverage = deepcopy(contributor.navitia_coverage)
         self.navitia_token = deepcopy(contributor.navitia_token)
 
+    @new_relic.agent.background_task(name="piv_worker-enter", group="Task")
     def __enter__(self):
         self.connection = Connection(self.builder.contributor.broker_url)
         self.exchange = self._get_exchange(self.builder.contributor.exchange_name)
@@ -109,6 +111,7 @@ class PivWorker(ConsumerMixin):
             )
         ]
 
+    @new_relic.agent.background_task(name="piv_worker-process_message", group="Task")
     def process_message(self, body, message):
         try:
             wrap_build(self.builder, body)
@@ -121,6 +124,7 @@ class PivWorker(ConsumerMixin):
             # * we do not want to process this message after another one (produced later) on the same train
             message.ack()
 
+    @new_relic.agent.background_task(name="piv_worker-on_iteration", group="Task")
     def on_iteration(self):
         if datetime.now() - self.last_config_checked_time < CONF_RELOAD_INTERVAL:
             return
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -162,4 +162,8 @@ def piv_worker():
                 worker.run()
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
+        finally:
+            db.session.expire(
+                contributor
+            )  # force db-reload otherwise staying locked on previous contributor's config
             time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -33,6 +33,7 @@ from kirin import manager, app
 from kirin.core.model import db
 from kirin.core.types import ConnectorType
 from kirin.core.abstract_builder import wrap_build
+from kirin.exceptions import ConfigurationError
 from kirin.piv import KirinModelBuilder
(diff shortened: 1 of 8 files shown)


patochectp (approved):

kirin/utils.py
```diff
@@ -333,3 +335,23 @@ def get_lock(logger, lock_name, lock_timeout):
...
+
+def log_exception(exception, origin):
+    """
+    log all exceptions
+    """
+    logger = logging.getLogger(__name__)
+    message = ""
+    if hasattr(exception, "data"):
+        message = exception.data
+    error = "{ex} {data} {orig}".format(ex=exception.__class__.__name__, data=message, orig=origin)
```
woshilapin, line 348: How about using separators here? Because we don't know exactly what `exception.data` might look like. Something like that?

```suggestion
    error = "{ex}: {data} - {orig}".format(ex=exception.__class__.__name__, data=message, orig=origin)
```

kirin/utils.py
```diff
@@ -333,3 +335,23 @@ def get_lock(logger, lock_name, lock_timeout):
         if locked:
             logger.debug("releasing lock %s", lock_name)
             lock.release()
+
+
+def log_exception(exception, origin):
```
woshilapin, line 340: _minor_ - name `origin` doesn't feel clear to me. How about `message` since it could be any kind of additional message? If you want to keep it to only designing the source of where the exception comes from, `exception_origin` or `exception_source` would feel better I think.

kirin/command/piv_worker.py
```diff
@@ -100,10 +102,16 @@ def get_consumers(self, Consumer, channel):
...
-        wrap_build(self.builder, body)
-        # TODO: We might want to not acknowledge the message in case of error in
-        # the processing.
-        message.ack()
+        try:
+            wrap_build(self.builder, body)
+        except Exception as e:
+            log_exception(e, "piv_worker")
+        finally:
+            # We might want to not acknowledge the message in case of error in the processing.
```
woshilapin, line 121: It really is problematic. Because if it's an exception trying to write to the DB, we probably want to re-queue it and retry. But if it is because the JSON of the payload is incorrect, we probably want to dump it because we will never be able to parse it. Do you think we should discuss that more in details?

kirin/command/piv_worker.py
```diff
@@ -162,4 +162,8 @@ def piv_worker():
                 worker.run()
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
+        finally:
+            db.session.expire(
```
woshilapin, line 181: Good catch :+1:

kirin/command/piv_worker.py
```diff
@@ -71,8 +71,11 @@ def __init__(self, contributor):
         if not contributor.queue_name:
             raise ValueError("Missing 'queue_name' configuration for contributor '{0}'".format(contributor.id))
         self.last_config_checked_time = datetime.now()
-        self.broker_url = deepcopy(contributor.broker_url)
         self.builder = KirinModelBuilder(contributor)
+        # store config to spot configuration changes
```
woshilapin, line 81: I'm wondering, should we store directly `self.contributor = deepcopy(contributor)` and then in `on_iteration()`, compare directly `if self.contributor != contributor`? That also mean to use `self.contributor` everywhere instead of `self.builder.contributor`. Or maybe instead, we could directly initialize with `self.builder = KirinModelBuilder(deepcopy(contributor))`?

Note: I'm not even sure it would work because `deepcopy(contributor)` would probably bring with it all the ORM stuff so in the end, `deepcopy(contributor)` and `contributor` might still be updated at the same time... ORM magic!

kirin/exceptions.py
```diff
@@ -86,4 +86,9 @@ class UnauthorizedOnSubService(SubServiceError):
 
 class UnauthorizedValue(KirinException):
     code = 400
-    message = "Unauthorized value"
+    message = "unauthorized value"
+
+
+class ConfigurationError(KirinException):
+    code = 400
+    message = "configuration error"
```
woshilapin, line 94: Do we need these since if I understand correctly, this error is never gonna be used through an HTTP request?

kirin/command/piv_worker.py
```diff
@@ -166,8 +171,9 @@ def piv_worker():
                 worker.run()
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
+            if isinstance(e, ConfigurationError):
```
woshilapin, line 178: I'm wondering, instead of using this `ConfigurationError` only to catch errors in the constructor of `PivWorker`, how about using `@retry` on `PivWorker.__init__()` instead of with no timeout. I believe it should achieve a similar behavior, adding an exponential time-delta between each retry and avoiding our hand-made solution.

kirin/command/piv_worker.py
pbougue: I thought the same and ended up lazy on that... If you both think it's worthy I'm OK to dig.

kirin/command/piv_worker.py
pbougue: We have some means (exception types) to distinguish errors that are "retryable" from the ones from payload being bad, and it's already used (`allow_reprocess` system for GTFS-RT and POST).
But requeueing is the real issue here... And I'm not sure it's worth digging that for now :-/

kirin/utils.py
pbougue: `message` is already used below to bear `exception.data` (which is the more informative), so OK for `exception_source`.

kirin/command/piv_worker.py
woshilapin: The problem I see is that it's probably much more likely that an error happens somewhere in the system and that we want to retry, in order to not lose the message. I would say the probability that the payload is not JSON is low.

The default behavior you chose is to ignore the message but I wonder if the other default would be better?

kirin/command/piv_worker.py
pbougue: Ended up with https://github.com/CanalTP/kirin/pull/337/commits/55371e2d9a75c232c2a5684e9ab272ea0367c0ae
Which is still quite hand-made but without much exception handling...

woshilapin (approved):

PR #340, opened 2020-11-04
[feature][PIV] Bind company and physical mode

woshilapin:
:book: This PR is binding to the correct company instead of a random company. Also, default company and physical mode are used.

:test_tube: For the tests:
- `test_piv_delayed` is testing the nominal case (there is a company which is found in Navitia)
- `test_no_company_source_code_default_to_company_1187` is testing the case where the company is not found in Navitia but the fallback default company 1187 is found
	- This test does not work yet since it will be supported once trip add will be implemented (:bookmark: ND-1021)
- there is no test for when no company can be found **and** the fallback default company can't be found either
	- The _mock_ for the default company already exists and it doesn't seem possible to remove one _mock_ **only for one test**

:mag_right: You can review commit by commit.

:bookmark: ref: ND-1068.


patochectp (approved):

kirin/piv/model_maker.py
```diff
@@ -313,9 +316,17 @@ def _make_trip_update(self, json_train, vj):
...
-        trip_update.company_id = self._get_navitia_company(company_id)
+        trip_update.company_id = self._get_navitia_company_id(company_id)
+        if not trip_update.company_id:
+            trip_update.company_id = self._get_navitia_company_id(DEFAULT_COMPANY_CODE)
+            if not trip_update.company_id:
+                raise ObjectNotFound(
+                    "no company found for key {}, nor for the default company {}".format(
+                        company_id, DEFAULT_COMPANY_CODE
+                    )
+                )
```
pbougue, line 327: minor: how about moving that in a dedicated function (maybe even `_get_navitia_company_id`) to "lighten" `_make_trip_update` ?

kirin/piv/model_maker.py
```diff
@@ -313,9 +316,17 @@ def _make_trip_update(self, json_train, vj):
...
         physical_mode = json_train.get("modeTransport").get("typeMode")
-        trip_update.physical_mode_id = self._get_navitia_physical_mode(physical_mode)
+        trip_update.physical_mode_id = self._get_navitia_physical_mode_id(physical_mode)
```
pbougue, line 329: It looks OK: sending `None` if physical_mode is not found, so Navitia will fallback to "some physical mode" which seems better than no trip :-)

kirin/piv/model_maker.py
```diff
@@ -402,36 +413,33 @@ def _make_trip_update(self, json_train, vj):
...
-        return None
+        return companies[0] if companies else None
 
-    def _get_navitia_physical_mode(self, indicator=None):
+    def _get_navitia_physical_mode_id(self, indicator=None):
         """
         Get a navitia physical_mode for the codes present in PIV (FERRE / ROUTIER)
         """
         uri = {
             "FERRE": "physical_mode:LongDistanceTrain",
```
pbougue, line 434: minor (and not sure): should we reuse `DEFAULT_PHYSICAL_MODE_ID` here ?

pbougue (approved):
That motivation to add so much fixtures :clap: ... I probably would have fallback to a new 2-stops VJ.

kirin/piv/model_maker.py
woshilapin: I thought about that but for me, here is not the default value, it's the corresponding value for `FERRE`. That's the reason I didn't use the constant here. Unless you're strongly opinionated on this one, I'll probably leave it as it is.

kirin/piv/model_maker.py
woshilapin: I've inlined it in `_get_navitia_company_id()` function (see f0d17f99edfea2b3625dbb887e9f3c9e0381d36c).

kirin/piv/model_maker.py
pbougue: Nope, as I think of it, the only "valid" use would be to rename const to `LONG_DISTANCE_TRAIN_ID` then reuse it.
You can resolve anyway :+1:

PR #341, opened 2020-11-05
[feature] piv: partial removal

patochectp:
ref: ND-1084


tests/integration/piv_test.py
```diff
@@ -305,6 +305,14 @@ def test_piv_event_priority(mock_rabbitmq):
         assert db_trip_removal.message == "Indisponibilité d'un matériel"
         # full trip removal : no stop_time to precise
         assert len(db_trip_removal.stop_time_updates) == 0
+
+        for s in db_trip_removal.stop_time_updates[0:16]:
+            assert s.arrival_status == "none"
+            assert s.departure_status == "none"
+            assert s.message is None
+            assert s.arrival_status == "delete"
+            assert s.departure_status == "delete"
```
woshilapin, line 314: I do not understand how is it possible that both
```python
             assert s.arrival_status == "none"
```
and
```python
             assert s.arrival_status == "delete"
```
can be true :open_mouth:

woshilapin (approved):

tests/integration/piv_test.py
patochectp: db_trip_partial_removed.stop_time_updates len = 0 so the for loop is not executed

pbougue (approved):
Apart from the question it's OK, so we might also merge and add another PR if needed.

kirin/piv/model_maker.py
```diff
@@ -79,7 +79,8 @@
     TripEffect.UNDEFINED: maxint,
 }
 
-MANAGED_EVENTS = ["RETARD", "SUPPRESSION"]
+MANAGED_EVENTS = ["RETARD", "SUPPRESSION", "MODIFICATION_DESSERTE_SUPPRIMEE"]
```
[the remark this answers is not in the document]
pbougue: To be done in a separate ticket, but yes, it's the same behaviour.

PR #345, opened 2020-11-06
PIV: Add delay tests and minor status correction

pbougue:
:mag: You can review by commit with message.

Add some tests on partial delays and multiple posts (train `23187` is a shorter version of `23186`).
On the way, tying internal `stop.status` closer to PIV `evenement.type`

JIRA: https://jira.kisio.org/browse/ND-1082


tests/integration/piv_test.py
```diff
@@ -266,6 +266,25 @@ def test_piv_delayed(mock_rabbitmq):
...
+    piv_feed = get_fixture_data("piv/stomp_20201022_23186_delayed_5min.json")
+    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    assert "PIV feed processed" in res.get("message")
+    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    assert "PIV feed processed" in res.get("message")
+
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 2
+    db_trip_delayed = _check_db_stomp_20201022_23186_delayed_5min()
+    assert db_trip_delayed.effect == "SIGNIFICANT_DELAYS"
```
woshilapin, line 451: Why do you need to test this since it's already tested in `_check_db_stomp_20201022_23186_delayed_5min()`.

At the very least, you can probably use the enum to test it.

```suggestion
    assert db_trip_delayed.effect == TripEffect.SIGNIFICANT_DELAYS.name
```

tests/integration/piv_test.py
```diff
@@ -266,6 +266,25 @@ def test_piv_delayed(mock_rabbitmq):
...
+    double delayed stops post
+    """
+    piv_feed = get_fixture_data("piv/stomp_20201022_23186_delayed_5min.json")
+    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    assert "PIV feed processed" in res.get("message")
+    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    assert "PIV feed processed" in res.get("message")
+
+    with app.app_context():
+        assert len(RealTimeUpdate.query.all()) == 2
```
woshilapin, line 449:
```suggestion
        assert RealTimeUpdate.query.count() == 2
```

tests/integration/piv_test.py
```diff
@@ -266,6 +266,25 @@ def test_piv_delayed(mock_rabbitmq):
...
+    db_trip_delayed = _check_db_stomp_20201022_23186_delayed_5min()
+    assert db_trip_delayed.effect == "SIGNIFICANT_DELAYS"
+    assert len(db_trip_delayed.stop_time_updates) == 17
```
woshilapin, line 452: Same here, already tested in `_check_db_stomp_20201022_23186_delayed_5min()`.

woshilapin (approved):

tests/integration/piv_test.py
pbougue: I test chaining of feeds here.
And you're right about enum :+1:

tests/integration/piv_test.py
```diff
@@ -252,6 +252,175 @@ def _check_db_stomp_20201022_23186_delayed_5min():
...
+        # no specific functional constraint on last departure, except time consistency
+        assert last_st.arrival <= last_st.departure
+        assert last_st.message == None
+
+        assert db_trip_delayed.contributor_id == PIV_CONTRIBUTOR_ID
+
+        return db_trip_delayed  # for additional testing if needed
+
+
+def _check_db_stomp_20201022_23187_delayed_5min():
```
woshilapin, line 341: Would it be possible to merge this one with the other function that tests `delayed_5min`, I guess it would reduce the number of function, probably reduce the number of stops from the other function (17 to 6). Also, I'm still thinking these functions should be named like this

```suggestion
def _assert_db_stomp_20201022_23187_delayed_5min():
```

kirin/piv/model_maker.py
```diff
@@ -394,17 +394,14 @@ def _make_trip_update(self, json_train, vj):
...
-                            )  # minutes
+                        piv_event_delay = piv_disruption.get("retard", {}).get("duree", 0)
+                        setattr(st_update, DELAY_MAP[event_toggle], as_duration(piv_event_delay * 60))  # minutes
+
                         piv_stop_time_status = get_value(piv_disruption, "type", nullable=True)
-                        if piv_stop_time_status and piv_stop_time_status == "SUPPRESSION_PARTIELLE":
+                        if piv_stop_time_status in ["RETARD_OBSERVE", "RETARD_PROJETE"]:
+                            setattr(st_update, STATUS_MAP[event_toggle], ModificationType.update.name)
+                        elif piv_stop_time_status == "SUPPRESSION_PARTIELLE":
                             setattr(st_update, STATUS_MAP[event_toggle], ModificationType.delete.name)
```
woshilapin, line 404: Do we need something like the following?

```python
                        else:
                            setattr(st_update, STATUS_MAP[event_toggle], ModificationType.none.name)
```

tests/integration/piv_test.py
pbougue: I renamed it (didn't remember that valid suggestion until now).

It can't be merged as-is though, as the third stop is 10min later: to be able to have the second stop 5min late but not inconsistent in stop-times.
As the stops in the other one are partly tested in a loop and also lots are used to test messages...
I'm not super motivated to merge those :tired_face:

kirin/piv/model_maker.py
pbougue: We could, I'm OK to add it if it's clearer for everyone.
The default values lead to that, so I didn't add it and kept it like in cots connector.

woshilapin: In my mind, I had to go see the array of all the possible values to understand there was an `else` condition (said differently, I was not sure if the 2 existing `if` were covering all the possible cases or not). But this is a very small argument towards making it explicit. Either way is fine for me.

patochectp (approved):

PR #346, opened 2020-11-09
piv: test enhancement

patochectp:


tests/integration/piv_test.py
```diff
@@ -612,35 +613,71 @@ def test_piv_partial_removal(mock_rabbitmq):
...
+
 def test_piv_modification_limitation(mock_rabbitmq):
     """
-    the trip 23186 is deleted except the first two stop
+    the trip 23187 is deleted except the first two stop
     """
     # Simple modification limitation
-    piv_23186_removal = get_fixture_data("piv/stomp_20201022_23186_modification_limitation.json")
-    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_23186_removal)
+    piv_23187_removal = get_fixture_data("piv/stomp_20201022_23187_blank_fixture.json")
```
woshilapin, line 638: Maybe `get_fixture_data()` can directly returns the result of `ujson.load()` instead and you can modify the JSON object (instead of getting back-and-forth to the `str`).

woshilapin (commented):
Generally, I like this idea :+1:

tests/integration/piv_test.py
```diff
@@ -612,35 +616,71 @@ def test_piv_partial_removal(mock_rabbitmq):
...
+        if desserte["rang"] == rang:
+            desserte[dep_or_arr]["evenement"] = {"type": evt_type, "texte": message}
+            if motif_modification:
+                desserte[dep_or_arr]["motifModification"] = motif_modification
+
+
 def test_piv_modification_limitation(mock_rabbitmq):
     """
-    the trip 23186 is deleted except the first two stop
+    the trip 23187 is deleted except the first two stop
```
pbougue, line 638: minor
```suggestion
    the trip 23187 is deleted except the first two stops
```

tests/integration/piv_test.py
```diff
@@ -501,8 +502,11 @@ def test_piv_event_priority(mock_rabbitmq):
     """
     simple trip removal post
     """
-    piv_feed = get_fixture_data("piv/stomp_20201022_23186_trip_removal_delayed.json")
-    res = api_post("/piv/{}".format(PIV_CONTRIBUTOR_ID), data=piv_feed)
+    piv_feed = get_fixture_data_as_dict("piv/stomp_20201022_23187_blank_fixture.json")
+    _set_piv_disruption(piv_feed, evt_type="SUPPRESSION", message="Indisponibilité d'un matériel")
+    _set_piv_disruption(piv_feed, evt_type="RETARD", message="Absence inopinée d'un agent")
+
```
pbougue, line 508: Should you also update stops to have a realistic feed?
I'm afraid here we test only the changes managed in the current code, but we would be blind to undesirable side-effects of dealing with stops (or wrongly using fields that never appear with this method, so errors never spotted in tests).

In general, that's what I'm afraid of with this way of testing : it's closer to unit tests, so useful and more simple for a lot of things, but not replacing integration tests.
Maybe we should have both?

pbougue (commented):
This idea will make it lighter for more targeted tests to be added :+1: 
Although those are not integration test (or a bit less complete probably).

Just curious and trying to find some ideas: Did you try dumping the constructed json and compare to the "original" json to the one previously used?

PR #347, opened 2020-11-10
[tech] piv tests migration

patochectp:


woshilapin (approved):
Really nice to me. It does create a lot more code but:
- the code helps to focus on what really changes for the current test
- less fixtures

So for me, it's only better.

minor: Sometimes the `piv_feed` variable is a JSON object, and sometimes it's a `str`. Maybe you can homogenize the tests on that (put the `ujson.dumps()` always in the `api_post()` for example?).

tests/integration/piv_test.py
```diff
@@ -548,64 +669,67 @@ def test_no_company_source_code_default_to_company_1187(mock_rabbitmq):
 
 def test_piv_partial_removal(mock_rabbitmq):
     """
-    the trip 23186 is partially deleted
+    the trip 23187 is partially deleted
 
-    Normally there are 17 stops in this VJ, but 2 (Mies, Pont-Céard) are
-    removed (departure from Tannay deleted)
+    Normally there are 6 stops in this VJ, but 2 (Mies, Genève Eaux-Vives)
+    respectively rang 2 and 3
```
pbougue, line 675: Not sure:
```suggestion
    ,respectively rang 2 and 3, are removed.
```

pbougue (commented):
Some navitia fixtures can also be removed (23186 VJ, and probably some stops too, maybe 841252 VJ and the stomp too?), no?
:heavy_check_mark: Done

pbougue (approved):
:+1: :mushroom:

PR #356, opened 2020-12-01
[tech] piv worker: manage pending transaction

patochectp:
ref: ND-1110

### DIFF
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -176,6 +176,11 @@ def piv_worker():
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
         finally:
+            try:
+                db.session.commit()
+            except Exception as db_e:
+                logger.warning("Exception while db-commit: {0}".format(db_e))
+                db.session.rollback()
             if should_wait:
                 time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
             db.session.expire(


woshilapin (approved):

kirin/command/piv_worker.py
```diff
@@ -173,8 +173,10 @@ def piv_worker():
                 should_wait = False  # wait only after init crash
                 logger.info("launching the PIV worker for '{0}'".format(contributor.id))
                 worker.run()
+            db.session.commit()
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
+            db.session.rollback()
```
pbougue, line 179: Maybe we could try to commit then rollback if it blocks.
But it's OK like that as we can also assume that if a commit is really needed, it should be secured earlier :-)

pbougue (approved):

kirin/command/piv_worker.py
```diff
@@ -173,8 +173,10 @@ def piv_worker():
...
         finally:
```
pbougue, line 180: How about inserting here:
```suggestion
        finally:
            try:
                db.session.commit()
            except Exception as db_e:
                logger.warning("Exception while db-commit: {0}".format(db_e))
                db.session.rollback()
```
The idea would be to avoid direct rollback when another exception bring us in the except.

kirin/command/piv_worker.py
```diff
@@ -173,9 +173,15 @@ def piv_worker():
                 should_wait = False  # wait only after init crash
                 logger.info("launching the PIV worker for '{0}'".format(contributor.id))
                 worker.run()
+            db.session.commit()
```
pbougue, line 176: As you do it just after in the finally it doesn't hurt but this line should not be really useful.

pbougue (approved):

kirin/command/piv_worker.py
patochectp: Done

pbougue (approved):

PR #357, opened 2020-12-08
[tech] piv worker: manage concurrency

patochectp:
ref: ND-1141

### DIFF
--- a/kirin/command/piv_worker.py
+++ b/kirin/command/piv_worker.py
@@ -46,13 +46,23 @@ from copy import deepcopy
 import logging
 import time
 
-from kirin.utils import log_exception
+from kirin.utils import log_exception, get_lock, make_kirin_lock_name
+from kirin import redis_client
+from redis.exceptions import LockError
+
 
 logger = logging.getLogger(__name__)
 
 CONF_RELOAD_INTERVAL = timedelta(
     seconds=float(str(app.config.get(str("BROKER_CONSUMER_CONFIGURATION_RELOAD_INTERVAL"))))
 )
+# In case of postgresql disconnection, it seems that the timeout (so exception) takes 10 seconds to happen,
+# the lock expiration shall not occur during this time otherwise it can lead to incorrect state (still working but
+# unlocked)
+PIV_WORKER_REDIS_TIMEOUT_LOCK = max(
+    timedelta(seconds=float(str(app.config.get(str("REDIS_LOCK_TIMEOUT_POLLER"))))), timedelta(seconds=15)
+)
+PIV_LOCK_NAME = make_kirin_lock_name("piv_poller")  # only one contributor and one PIV worker at a time
 
 
 class PivWorker(ConsumerMixin):
@@ -80,6 +90,7 @@ class PivWorker(ConsumerMixin):
         self.broker_url = deepcopy(contributor.broker_url)
         self.navitia_coverage = deepcopy(contributor.navitia_coverage)
         self.navitia_token = deepcopy(contributor.navitia_token)
+        self.last_lock_update = datetime.now()
 
     @new_relic.agent.background_task(name="piv_worker-enter", group="Task")
     def __enter__(self):
@@ -92,7 +103,7 @@ class PivWorker(ConsumerMixin):
         self.connection.release()
 
     def _get_exchange(self, exchange_name):
-        return Exchange(name=exchange_name, type="fanout", durable=True, no_declare=True, auto_delete=False)
+        return Exchange(name=exchange_name, type=str("fanout"), durable=True, no_declare=True, auto_delete=False)
 
     def _get_or_create_queue(self, queue_name):
         queue = Queue(name=queue_name, exchange=self.exchange, durable=True, auto_delete=False)
@@ -124,6 +135,11 @@ class PivWorker(ConsumerMixin):
 
     @new_relic.agent.background_task(name="piv_worker-on_iteration", group="Task")
     def on_iteration(self):
+        if datetime.now() - self.last_lock_update > PIV_WORKER_REDIS_TIMEOUT_LOCK // 5:
+            redis_client.expire(PIV_LOCK_NAME, PIV_WORKER_REDIS_TIMEOUT_LOCK)
+            self.last_lock_update = datetime.now()
+            logger.debug("lock {%s} updated", PIV_LOCK_NAME)
+
         if datetime.now() - self.last_config_checked_time < CONF_RELOAD_INTERVAL:
             return
 
@@ -152,38 +168,47 @@ class PivWorker(ConsumerMixin):
 
 @manager.command
 def piv_worker():
-    import sys
-
     # We assume one and only one PIV contributor is going to exist in the DB
+    # Therefore, we enforce one and only one `PivWorker` is running at a time
     while True:
-        should_wait = True
         try:
-            contributors = get_piv_contributors()
-            if len(contributors) == 0:
-                logger.warning("no PIV contributor")
-                time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
-                continue
-            contributor = contributors[0]
-            if len(contributors) > 1:
-                logger.warning(
-                    "more than one PIV contributors: {0}; choosing '{1}'".format(
-                        map(lambda c: c.id, contributors), contributor.id
-                    )
-                )
-            with PivWorker(contributor) as worker:
-                should_wait = False  # wait only after init crash
-                logger.info("launching the PIV worker for '{0}'".format(contributor.id))
-                worker.run()
-        except Exception as e:
-            logger.warning("PIV worker died: {0}".format(e))
-        finally:
-            try:
-                db.session.commit()
-            except Exception as db_e:
-                logger.warning("Exception while db-commit: {0}".format(db_e))
-                db.session.rollback()
-            if should_wait:
-                time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
-            db.session.expire(
-                contributor
-            )  # force db-reload otherwise staying locked on previous contributor's config
+            with get_lock(logger, PIV_LOCK_NAME, PIV_WORKER_REDIS_TIMEOUT_LOCK.total_seconds()) as locked:
+                if locked:
+                    should_wait = True
+                    try:
+                        contributors = get_piv_contributors()
+                        if len(contributors) == 0:
+                            logger.warning("no PIV contributor")
+                            time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
+                            continue
+                        contributor = contributors[0]
+                        if len(contributors) > 1:
+                            logger.warning(
+                                "more than one PIV contributors: {0}; choosing '{1}'".format(
+                                    map(lambda c: c.id, contributors), contributor.id
+                                )
+                            )
+
+                        with PivWorker(contributor) as worker:
+                            should_wait = False  # wait only after init crash
+                            logger.info("launching the PIV worker for '{0}'".format(contributor.id))
+                            worker.run()
+
+                    except Exception as e:
+                        logger.warning("PIV worker died: {0}".format(e))
+                    finally:
+                        try:
+                            db.session.commit()
+                        except Exception as db_e:
+                            logger.warning("Exception while db-commit: {0}".format(db_e))
+                            db.session.rollback()
+                        if should_wait:
+                            time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
+                        db.session.expire(
+                            contributor
+                        )  # force db-reload otherwise staying locked on previous contributor's config
+                else:
+                    logger.debug("Lock {%s} un-acquired", PIV_LOCK_NAME)
+                    time.sleep(CONF_RELOAD_INTERVAL.total_seconds())
+        except LockError as lock_e:
+            logger.warning("Exception with lock : {0}".format(lock_e))
--- a/tests/integration/piv_worker_test.py
+++ b/tests/integration/piv_worker_test.py
@@ -30,18 +30,21 @@
 # www.navitia.io
 from __future__ import absolute_import, print_function, unicode_literals, division
 
-from kirin import app, db
-from kirin.command.piv_worker import PivWorker
+from kirin.command.piv_worker import PivWorker, PIV_LOCK_NAME, PIV_WORKER_REDIS_TIMEOUT_LOCK
 from kirin.core.model import RealTimeUpdate
-from kirin.core.types import ConnectorType
 from kirin.piv.piv import get_piv_contributor
 from tests.integration.conftest import PIV_CONTRIBUTOR_ID, PIV_EXCHANGE_NAME, PIV_QUEUE_NAME
 
 from amqp.exceptions import NotFound
-from kombu import Connection, Exchange, Queue
+from kombu import Connection, Exchange
 import pytest
+import logging
+
 import threading
 from retrying import retry
+from mock import patch
+from datetime import datetime
+import time
 
 
 @retry(stop_max_delay=20000, wait_exponential_multiplier=100)
@@ -53,7 +56,7 @@ def is_exchange_created(connection, exchange_name, exchange_type="fanout"):
     try:
         channel = connection.channel()
         channel.exchange_declare(exchange_name, exchange_type, nowait=False, passive=True)
-    except NotFound as e:
+    except NotFound:
         return False
     except Exception as e:
         raise e
@@ -64,7 +67,7 @@ def is_queue_created(connection, queue_name):
     try:
         channel = connection.channel()
         channel.queue_declare(queue=queue_name, nowait=False, passive=True)
-    except NotFound as e:
+    except NotFound:
         return False
     except Exception as e:
         raise e
@@ -86,7 +89,7 @@ def create_exchange(broker_connection, exchange_name):
         exchange_name,
         durable=True,
         delivery_mode=2,
-        type="fanout",
+        type=str("fanout"),
         auto_delete=False,
         no_declare=False,
     )
@@ -102,8 +105,7 @@ def init_piv_exchange(broker_connection):
 
 
 def launch_piv_worker(pg_docker_fixture):
-    import kirin
-    from kirin import app, db, manager
+    from kirin import app
     from tests.conftest import init_flask_db
     from tests.integration.conftest import PIV_CONTRIBUTOR_ID
 
@@ -111,6 +113,7 @@ def launch_piv_worker(pg_docker_fixture):
         # re-init the db by overriding the db_url
         init_flask_db(pg_docker_fixture)
         contributor = get_piv_contributor(PIV_CONTRIBUTOR_ID)
+
         with PivWorker(contributor) as worker:
             worker.run()
 
@@ -121,6 +124,7 @@ class PivWorkerTest:
         self.broker_url = broker_url
         self.broker_connection = broker_connection
         self.pg_docker_fixture = pg_docker_fixture
+        self.last_lock_update = datetime.now()
 
     def __enter__(self):
         # Launch a PivWorker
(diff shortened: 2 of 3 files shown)


kirin/command/piv_worker.py
```diff
@@ -112,7 +112,10 @@ def get_consumers(self, Consumer, channel):
     @new_relic.agent.background_task(name="piv_worker-process_message", group="Task")
     def process_message(self, body, message):
         try:
-            wrap_build(self.builder, body)
+            lock_name = make_kirin_lock_name("piv_poller", self.builder.contributor.id)
+            with get_lock(logger, lock_name, app.config[str("REDIS_LOCK_TIMEOUT_POLLER")]) as locked:
+                if locked:
```
pbougue, line 117: Whether Kirin gets the lock or not, `ack()` is called for the message, so the message is lost in case Kirin didn't get the lock (not treated but acknowledged), no?

I don't see any other outcome to that case, but it would be so nice to be able to test that (but maybe expensive).
Maybe first set the lock to mock a concurrent worker, then call piv_worker on a rabbit : reusing what's done into piv_worker_test.py + mock_debug + count message in queue if it's easy with kombu... or just see if the message gets consumed when removing the lock.

And then to fix that, probably more something like:
```python
try:
    lock_name = make_kirin_lock_name("piv_poller", self.builder.contributor.id)
    with get_lock(logger, lock_name, app.config[str("REDIS_LOCK_TIMEOUT_POLLER")]) as locked:
        if locked:
            try:
                wrap_build(self.builder, body)
            finally:
                #blabla
                message.ack()
except Exception as e:
    log_exception(e, "piv_worker")
```

tests/integration/piv_worker_test.py
```diff
@@ -147,3 +147,26 @@ def test_mq_message_received_and_stored(
...
+            wait_until(lambda: is_queue_created(broker_connection, PIV_QUEUE_NAME))
+
+            mq_handler.publish(str('{"key": "Some valid JSON"}'), PIV_CONTRIBUTOR_ID)
+            wait_until(lambda: RealTimeUpdate.query.count() == 1)
+            # Check lock acquired and released
+            calls = [
+                call("getting lock kirin.lock|piv_poller|rt.piv"),
+                call("releasing lock kirin.lock|piv_poller|rt.piv"),
+            ]
+            mock_debug.assert_has_calls(calls, any_order=False)
```
pbougue, line 172: :heart: sweet.
It may allow tests that would be super-expensive to do otherwise! It's sometimes a bit hacky but way better than nothing and quite cheap :+1:

kirin/command/piv_worker.py
```diff
@@ -112,7 +112,10 @@ def get_consumers(self, Consumer, channel):
...
+                    wrap_build(self.builder, body)
```
pbougue, line 118: I wonder if we should wait a bit (like 1s) before retrying when we don't get the lock.
To avoid stressing redis and rabbit too much doing nothing.
It would probably end with always the same worker doing the job but it's fine I think (the other is a backup).

So adding (not tested, didn't check time unit):
```python
                else:  # if not locked
                    sleep(1)
```

tests/integration/piv_worker_test.py
woshilapin: I agree, sweet test :heart_eyes:

woshilapin (dismissed):

kirin/utils.py
```diff
@@ -327,7 +327,7 @@ def get_database_pool_status():
 def get_lock(logger, lock_name, lock_timeout):
     from kirin import redis_client
 
-    logger.debug("getting lock %s", lock_name)
+    logger.debug("getting lock %s" % lock_name)
```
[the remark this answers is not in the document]
patochectp: with comma version I'm getting `getting lock %s', u'kirin.lock|piv_poller|rt.piv` instead of `getting lock kirin.lock|piv_poller|rt.piv`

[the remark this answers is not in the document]
woshilapin: Weird. Since we mostly use the following syntax usually, any reason to use the `%` operator here. This is truly a naive question since I do not really know the consequence of using one or the other.

```suggestion
    logger.debug("getting lock {}".format(lock_name)) 
```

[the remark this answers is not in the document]
patochectp: You're right, with a comma the message is properly formatted in the console but not with the mock. I’ll find another solution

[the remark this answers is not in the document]
woshilapin: Ah, that is the `mock`. So I think it's actually fine keeping how it was before and changing how the tests to something like `call("getting lock %s", "kirin.lock|piv_poller|rt.piv")`. If I understand correctly, the `mock` is only testing how the mock was called (with which arguments so it expects the same one).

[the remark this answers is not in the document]
patochectp: That's what I did

kirin/command/piv_worker.py
```diff
@@ -124,6 +129,11 @@ def process_message(self, body, message):
 
     @new_relic.agent.background_task(name="piv_worker-on_iteration", group="Task")
     def on_iteration(self):
+        if datetime.now() - self.last_lock_update > CONF_LOCK_EXPIRE - timedelta(seconds=2):
```
woshilapin, line 132: Why do you have a ` - timedelta(seconds=2)` here? It looks like you're will refresh the lock expiration value and want to catch it before it's outdated... but I'm not sure. Maybe a comment would help if that's the case. (and 1 second is not enough ? how about using 5 seconds or more to be sure ?).

kirin/command/piv_worker.py
```diff
@@ -124,6 +129,11 @@ def process_message(self, body, message):
...
+            redis_client.expire(self.lock_name, CONF_LOCK_EXPIRE)
```
woshilapin, line 133: Just me complaining about something we cannot do anything about...

This `.expire()` has really a misleading name. I would expect calling this method expires a lock right now... but it is actually to set the delay for expiration if I understand correctly! Weird API in my opinion but not your fault.

kirin/command/piv_worker.py
```diff
@@ -152,9 +162,9 @@ def on_iteration(self):
 
 @manager.command
 def piv_worker():
-    import sys
-
     # We assume one and only one PIV contributor is going to exist in the DB
+    lock_name = "{}|{}".format(app.config[str("TASK_LOCK_PREFIX")], PIV_POLLER)
```
woshilapin, line 166: Is this to enforce we're going to have only one `PivWorker` running at a time? If yes, I suggest the comment above to also make this precision. Something like

```
     # We assume one and only one PIV contributor is going to exist in the DB.
     # Therefore, we enforce one and only one `PivWorker` is running at a time.
```

kirin/command/piv_worker.py
```diff
@@ -46,18 +46,21 @@
...
     seconds=float(str(app.config.get(str("BROKER_CONSUMER_CONFIGURATION_RELOAD_INTERVAL"))))
 )
+CONF_LOCK_EXPIRE = timedelta(seconds=float(str(app.config.get(str("REDIS_LOCK_TIMEOUT_POLLER")))) / 5)
+PIV_POLLER = "piv_poller"
 
 
 class PivWorker(ConsumerMixin):
     @new_relic.agent.background_task(name="piv_worker-init", group="Task")
-    def __init__(self, contributor):
+    def __init__(self, contributor, lock_name):
```
woshilapin, line 63: I see how we might use this new parameter in the future but for now, there is only one `PivWorker`, only one `lock_name`, so the name of the lock could just be a global constant, and we wouldn't need this new parameter.

kirin/command/piv_worker.py
```diff
@@ -170,13 +180,21 @@ def piv_worker():
...
+                        logger.info("launching the PIV worker for '{0}'".format(contributor.id))
+                        worker.run()
+                else:
+                    logger.debug("Lock {%s} un-acquired", lock_name)
+
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
         finally:
+            if locked:
+                redis_client.delete(lock_name)
```
woshilapin, line 196: Should we also remove the lock inside the `PivWorker` itself when we quit it (line [15](https://github.com/CanalTP/kirin/pull/357/files#diff-ac2785653924e36dfb16946d557b7c72e51fea24bac926570c14e24589d796f4R159))? That said, I'm worried that we release the lock maybe too early if do this...

kirin/command/piv_worker.py
```diff
@@ -170,13 +180,21 @@ def piv_worker():
...
+                        should_wait = False  # wait only after init crash
+                        logger.info("launching the PIV worker for '{0}'".format(contributor.id))
+                        worker.run()
+                else:
+                    logger.debug("Lock {%s} un-acquired", lock_name)
+
         except Exception as e:
             logger.warning("PIV worker died: {0}".format(e))
         finally:
+            if locked:
```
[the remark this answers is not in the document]
patochectp: if we have 2 piv_worker we can try to delete a lock of the other worker and will fail

kirin/command/piv_worker.py
```diff
@@ -170,13 +180,21 @@ def piv_worker():
                         map(lambda c: c.id, contributors), contributor.id
                     )
                 )
-            with PivWorker(contributor) as worker:
-                should_wait = False  # wait only after init crash
-                logger.info("launching the PIV worker for '{0}'".format(contributor.id))
-                worker.run()
+            with get_lock(logger, lock_name, app.config[str("REDIS_LOCK_TIMEOUT_POLLER")]) as locked:
+                if locked:
```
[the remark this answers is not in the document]
patochectp: Code above don't write in db, just read and I think that the lock must be as close as possible to the modifications

kirin/command/piv_worker.py
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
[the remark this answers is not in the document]
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
```diff
@@ -46,18 +46,21 @@
...
+from kirin.utils import log_exception, get_lock
+from kirin import redis_client
 
 logger = logging.getLogger(__name__)
 
 CONF_RELOAD_INTERVAL = timedelta(
     seconds=float(str(app.config.get(str("BROKER_CONSUMER_CONFIGURATION_RELOAD_INTERVAL"))))
 )
+CONF_LOCK_EXPIRE = timedelta(seconds=float(str(app.config.get(str("REDIS_LOCK_TIMEOUT_POLLER")))) / 5)
+PIV_POLLER = "piv_poller"
```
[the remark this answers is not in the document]
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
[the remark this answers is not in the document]
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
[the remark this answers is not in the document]
patochectp: done in 498b92964bcdcbbf488b3abec5acd2ac7431b7f9

kirin/command/piv_worker.py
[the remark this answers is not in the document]
woshilapin: I'm not sure I agree with this because `on_iteration()` is not called every CPU clock cycle but more something like every second. So if `PIV_WORKER_REDIS_TIMEOUT_LOCK` is less than 5 seconds, there is a high chance you actually miss it. Maybe a minimum value of 1 second would help?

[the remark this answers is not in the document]
patochectp: Default REDIS_LOCK_TIMEOUT_POLLER is 5 min

kirin/command/piv_worker.py
woshilapin: I do not see what has been done for that comment. Note that I mistakenly wrote `line 15` instead of `line 159` (I fixed it since, but the link was correct). Maybe this has mislead you...

kirin/command/piv_worker.py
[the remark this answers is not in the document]
woshilapin: Sure, but it is configurable and it's actually very likely you'll put a very low value if you try to write tests on it.

kirin/command/piv_worker.py
patochectp: redis_client.delete has been removed at the request of @pbougue because the lock is removed when we released it

woshilapin: Ah OK, didn't understand that part. I'm resolving this conversation then.

kirin/command/piv_worker.py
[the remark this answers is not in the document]
woshilapin: As discussed together, maybe a fifth is a good idea, as long as it is bigger than at least 2 seconds. So something like this maybe?

```python
max(PIV_WORKER_REDIS_TIMEOUT_LOCK // 5, timedelta(seconds=2))
```

@pbougue What do you think?

tests/integration/piv_worker_test.py
```diff
@@ -147,3 +149,23 @@ def test_mq_message_received_and_stored(
...
+    logger = logging.getLogger("kirin.command.piv_worker")
+    with patch.object(logger, "debug") as mock_debug:
+        with PivWorkerTest(test_client, rabbitmq_docker_fixture.url, broker_connection, pg_docker_fixture):
+            # Check that PivWorker is creating the queue
+            wait_until(lambda: is_queue_created(broker_connection, PIV_QUEUE_NAME))
+
+            mq_handler.publish(str('{"key": "Some valid JSON"}'), PIV_CONTRIBUTOR_ID)
+            wait_until(lambda: RealTimeUpdate.query.count() == 1)
+            # Check lock acquired and released
+            mock_debug.assert_any_call("lock {%s} updated", PIV_LOCK_NAME)
```
pbougue, line 171: * Here you only test lock refresh, and it seems a bit lucky that the processing is longer than the interval between refresh, no?
* Can you test lock is acquired through `mock_debug.assert_any_call("getting lock %s", PIV_LOCK_NAME)`?
* It's nice to test that processing is still done, but it's not needed to acquire the lock (so it may be replaced by the sleep only if it's already tested elsewhere)
* No check that it's released (should only be the case if conf changes I guess and doable via `mock_debug.assert_any_call("releasing lock %s", PIV_LOCK_NAME)`).

```suggestion
            # Check lock acquired and refreshed
            mock_debug.assert_any_call("getting lock %s", PIV_LOCK_NAME)
            sleep((PIV_WORKER_REDIS_TIMEOUT_LOCK // 5) + 1)
            mock_debug.assert_any_call("lock {%s} updated", PIV_LOCK_NAME)
```

kirin/command/piv_worker.py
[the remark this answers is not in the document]
pbougue: Not sure about the subject here (looking at [**the current code**](https://github.com/CanalTP/kirin/pull/357/files#diff-ac2785653924e36dfb16946d557b7c72e51fea24bac926570c14e24589d796f4R131-R132)):
1. If it's about the value of `PIV_WORKER_REDIS_TIMEOUT_LOCK` OK to have a minimum on it, but for now it's based on `REDIS_LOCK_TIMEOUT_POLLER` directly (the one from GTFSRT-pollers), so we should be safe.
I'm OK to have a dedicated parameter, separated from the one on GTFSRT-pollers, too.
2. If it's about the interval when refresh happens I don't think it's a good idea, as we would have a lock not refresh before it expires.

pbougue (commented):
:+1: The cinematic is good to me, I don't think of tricky breaking case.

tests/integration/piv_worker_test.py
patochectp: We can't test getting and releasing lock because we only test the worker and not the wrapper

pbougue: Ah, true!
Then only the comment needs a "refresh" :stuck_out_tongue_closed_eyes:

kirin/command/piv_worker.py
```diff
@@ -152,38 +161,44 @@ def on_iteration(self):
...
-        except Exception as e:
-            logger.warning("PIV worker died: {0}".format(e))
-        finally:
-            try:
-                db.session.commit()
-            except Exception as db_e:
-                logger.warning("Exception while db-commit: {0}".format(db_e))
-                db.session.rollback()
-            if should_wait:
+        with get_lock(logger, PIV_LOCK_NAME, app.config[str("REDIS_LOCK_TIMEOUT_POLLER")]) as locked:
```
pbougue, line 167: reminder: reuse the const TIMEOUT

kirin/command/piv_worker.py
```diff
@@ -46,13 +46,16 @@
...
-from kirin.utils import log_exception
+from kirin.utils import log_exception, get_lock, make_kirin_lock_name
+from kirin import redis_client
 
 logger = logging.getLogger(__name__)
 
 CONF_RELOAD_INTERVAL = timedelta(
     seconds=float(str(app.config.get(str("BROKER_CONSUMER_CONFIGURATION_RELOAD_INTERVAL"))))
 )
+PIV_WORKER_REDIS_TIMEOUT_LOCK = timedelta(seconds=float(str(app.config.get(str("REDIS_LOCK_TIMEOUT_POLLER")))))
```
pbougue, line 57: As discussed :
```suggestion
PIV_WORKER_REDIS_TIMEOUT_LOCK = timedelta(seconds=max(float(str(app.config.get(str("REDIS_LOCK_TIMEOUT_POLLER"))))), 10)
```

tests/integration/piv_worker_test.py
```diff
@@ -165,7 +166,9 @@ def test_redis_lock_update(
             # Check that PivWorker is creating the queue
             wait_until(lambda: is_queue_created(broker_connection, PIV_QUEUE_NAME))
 
+            time.sleep((PIV_WORKER_REDIS_TIMEOUT_LOCK.total_seconds() // 5) + 1)
             mq_handler.publish(str('{"key": "Some valid JSON"}'), PIV_CONTRIBUTOR_ID)
             wait_until(lambda: RealTimeUpdate.query.count() == 1)
-            # Check lock acquired and released
+            # Check lock refreshed
             mock_debug.assert_any_call("lock {%s} updated", PIV_LOCK_NAME)
+            # mock_debug.assert_called_once_with("lock {%s} updated", PIV_LOCK_NAME)
```
woshilapin, line 174: Maybe this can be removed.

woshilapin (approved):

tests/integration/piv_worker_test.py
patochectp: done

kirin/command/piv_worker.py
```diff
@@ -46,13 +46,22 @@
...
+from redis.exceptions import LockError
+
 
 logger = logging.getLogger(__name__)
 
 CONF_RELOAD_INTERVAL = timedelta(
     seconds=float(str(app.config.get(str("BROKER_CONSUMER_CONFIGURATION_RELOAD_INTERVAL"))))
 )
+# In case of postgresql disconnection, it seems that the error handling seems take 10 seconds, the lock expiration
+# shall not occur during this time otherwise the piv worker definitely died
```
pbougue, line 60: minor:
```suggestion
# In case of postgresql disconnection, it seems that the timeout (so exception) takes 10 seconds to happen,
# the lock expiration shall not occur during this time otherwise it can lead to incorrect state (still working but unlocked)
```

pbougue (approved):

kirin/command/piv_worker.py
```diff
@@ -124,6 +134,11 @@ def process_message(self, body, message):
 
     @new_relic.agent.background_task(name="piv_worker-on_iteration", group="Task")
     def on_iteration(self):
+        if datetime.now() - self.last_lock_update > PIV_WORKER_REDIS_TIMEOUT_LOCK // 5:
+            redis_client.expire(PIV_LOCK_NAME, PIV_WORKER_REDIS_TIMEOUT_LOCK)
```
pbougue, line 138:
```suggestion
            redis_client.expire(PIV_LOCK_NAME, PIV_WORKER_REDIS_TIMEOUT_LOCK.total_seconds())
```

pbougue (changes requested):

kirin/command/piv_worker.py
patochectp: ``time`` can be represented by an integer or a Python timedelta object and tilmedelta has microsecond resolution thus total_seconds is a float

pbougue: Ah yep, I didn't check https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/client.py#L1651
:hand_over_mouth: sorry!

pbougue (approved):

PR #358, opened 2020-12-08
[tech] GithubActions: build only master, release and PRs

pbougue:
For now, all the deployment system is based on `release` branch and `git describe`, tags are only used for clearer information.
So they are built as they are always on `release` branch, but it's a side effect.
The goal is to avoid duplicate builds.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,5 +1,13 @@
 name: Continuous Integration
-on: [push, pull_request]
+
+on:
+    push:
+        branches:  # Build 'master' and 'release' branches
+            - 'master'
+            - 'release'
+        tags-ignore:  # No build on tags
+            - '**'
+    pull_request:  # Build any PR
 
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -1,5 +1,13 @@
 name: Continuous Integration
-on: [push, pull_request]
+on:
+    push:
+        branches:  # Build 'master' and 'release' branches
+            - 'master'
+            - 'release'
+        tags-ignore:  # No build on tags
+            - '**'
```
woshilapin, line 8: Why don't we want to build tags too?

pbougue: For now, all the deployment system is based on `release` branch and `git describe`, tags are only used for clearer information.
So they are built as they are always on `release` branch, but it's a side effect, and we don't want to duplicate builds.

woshilapin (approved):

PR #360, opened 2020-12-17
[tech] Improve error management (especially newrelic)

pbougue:
Also for logs.

* Document log-level separations, and matching with newrelic
* Refactor exception management for better dispatch in log-levels

JIRA: https://jira.kisio.org/browse/ND-841

### DIFF
--- a/kirin/new_relic.py
+++ b/kirin/new_relic.py
@@ -34,9 +34,10 @@ import logging
 import flask
 import os
 
+from navitia_wrapper import NavitiaException
 from werkzeug.exceptions import NotFound, MethodNotAllowed
 
-from kirin.exceptions import InvalidArguments
+from kirin.exceptions import InvalidArguments, ObjectNotFound, UnsupportedValue
 
 try:
     from newrelic import agent
@@ -50,8 +51,8 @@ def init(config_file):
     if agent and config_file and os.path.exists(config_file):
         agent.initialize(config_file)
     else:
-        logger = logging.getLogger(__name__)
-        logger.warning("newrelic hasn't been initialized")
+        log = logging.getLogger(__name__)
+        log.warning("newrelic hasn't been initialized")
 
 
 def must_never_log(exception):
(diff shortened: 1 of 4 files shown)


kirin/new_relic.py
```diff
@@ -65,18 +66,30 @@ def must_never_log(exception):
...
 
 
-def is_invalid_input_exception(exception):
-    if isinstance(exception, InvalidArguments):
+def is_only_warning_exception(exception):
+    if (
+        isinstance(exception, InvalidArguments)
+        or isinstance(exception, ObjectNotFound)
+        or isinstance(exception, NavitiaException)
+        or isinstance(exception, UnsupportedValue)
```
woshilapin, line 74: Should we create an array of all `WARNING_EXCEPTIONS`?

kirin/new_relic.py
```diff
@@ -65,18 +66,30 @@ def must_never_log(exception):
...
+    ):
         return True
-    # Any other exception is not related to an invalid input
+    # Any other exception is a failure that should alert developers and/or OPS
     return False
 
 
-def allow_apm_logging(exception):
+def is_reprocess_allowed(exception):
+    if isinstance(exception, InvalidArguments):
```
woshilapin, line 82: I'm questioning this. If you have an `UnsupportedValue`, I'm not sure I understand how reprocessing could work?

Also, I'm wondering about this black list vs white list here? Do we want to reprocess by default? Or do we want not reprocess by default? (I'm thinking about the case where we would add a fifth kind of exception in `is_only_warning_exception()`).

pbougue: `UnsupportedValue` won't have a different output unless a new version of Kirin is released. And those `UnsupportedValue` are not supposed to change (once the PIV development is finished, which is now the case). So they should not be reprocessed (and this reprocessing only affects processing when they are one just after the other, currently... mainly for gtfsrt).

As for the blacklist, I see 2 points that are soft PROS (but no big blocker for whitelist either):
* laziness: it's shorter and easier that way :grin: 
* the default allowing reprocessing implies more useless work at worse (but we are in a case where exceptions are managed and no serious crash expected). It also implies taking the chance of doing better next time (if navitia was not available for exemple) which is a good thing.

kirin/new_relic.py
pbougue: A tuple looks like a cleaner way indeed for documentation, I will amend the commit with that.

kirin/new_relic.py
woshilapin: `Unsupported` will be reprocessed here because the function returns `true`.

pbougue:
As discussed, I amended the commit considering your remarks @woshilapin.
:pray: thanks!

kirin/new_relic.py
```diff
@@ -65,18 +66,31 @@ def must_never_log(exception):
     return False
 
 
-def is_invalid_input_exception(exception):
-    if isinstance(exception, InvalidArguments):
+NOT_REPROCESSABLE_EXCEPTIONS = [InvalidArguments, UnsupportedValue]
+
+WARNING_EXCEPTIONS = [ObjectNotFound, NavitiaException]
+WARNING_EXCEPTIONS.extend(NOT_REPROCESSABLE_EXCEPTIONS)
```
woshilapin, line 72:
```suggestion
WARNING_EXCEPTIONS = [ObjectNotFound, NavitiaException] + NOT_REPROCESSABLE_EXCEPTIONS
```

woshilapin (approved):

kirin/new_relic.py
pbougue: :+1: Amended the commit once more.

woshilapin (approved):

patochectp (approved):

PR #362, opened 2020-12-22
[feature] Keep 'nb_days_to_keep_*' fields optional in API

woshilapin:
Related to this comment in our channel #navitia on Element
https://matrix.to/#/!FsygHZEAWANEgbeCMB:matrix.org/$1608631737617468xVRTf:matrix.org?via=matrix.org&via=bitma.st&via=kde.org

To summarize, `nb_days_to_keep_*` fields are documented as optional and indeed, the database does have a [default value](https://github.com/CanalTP/kirin/blob/master/kirin/core/model.py#L531-L532)... but our API makes it mandatory.

### DIFF
--- a/documentation/internal_format.md
+++ b/documentation/internal_format.md
@@ -29,8 +29,8 @@ is_active | Boolean, Optional | Used to activate/deactivate the kirin service fo
 broker_url| String, Optional | Url of the AMQP broker to listen for realtime information queue
 exchange_name| String, Optional | Exchange of the AMQP broker to listen for realtime information
 queue_name| String, Optional | Queue of the AMQP broker to listen for realtime information queue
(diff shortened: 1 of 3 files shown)


pbougue (approved):
I added a PR on your branch to precise doc too: https://github.com/woshilapin/kirin/pull/1

Originally, this was intended to be mandatory (and forgot doc update) to force conscious choice for our internal processes, but in an OSS vision it makes more sense to have it optional and a default value (and it's now documented internally to do an explicit choice for that).

Let you merge my PR if you're fine with it and merge this after whenever you're ready.

pbougue (changes requested):
I missed some points, will add them in my PR (still OK with the idea).

More changes (and test) are needed to have it optional in API.

pbougue (approved):

pbougue (approved):

PR #372, opened 2021-06-01
[tech] SonarCloud Integration

Moomoo-Seukla:
Add file properties and CI update. 
Project SonarCloud : https://sonarcloud.io/summary/overall?id=CanalTP_kirin 

_**Generate token  :**_ 
 1. Click on a project
 2. On the right hand side, Analysis method, click on the pen
 3. Click on ‘Follow the tutorial’ link under `GitHub Actions`
 4. Click on the pen next to the token ‘Value field’ and paste the custom token
 5. Click on Continue

NB: This is a new PR. The old PR was linked to my fork and that caused a problems

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -8,7 +8,6 @@ on:
         tags-ignore:  # No build on tags
             - '**'
     pull_request:  # Build any PR
-        types: [opened, synchronize, reopened]
 
 jobs:
     quality:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -8,6 +8,7 @@ on:
         tags-ignore:  # No build on tags
             - '**'
     pull_request:  # Build any PR
+        types: [opened, synchronize, reopened]
 
 jobs:
     quality:
@@ -68,3 +69,8 @@ jobs:
               with:
                   token: ${{ secrets.CODECOV_TOKEN }}
                   file: ./coverage.xml
+            - name: SonarCloud Scan
+              uses: SonarSource/sonarcloud-github-action@master
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,15 @@
+# sonar-project.properties
+
+# Project
+sonar.host.url=https://sonarcloud.io
+sonar.organization=canaltp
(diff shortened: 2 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -8,6 +8,7 @@ on:
         tags-ignore:  # No build on tags
             - '**'
     pull_request:  # Build any PR
+        types: [opened, synchronize, reopened]
```
pbougue, line 11: This is not required, as discussed before.
(if this actually is, I want to understand why)

pbougue (commented):
I don't get what you did with the token in the end to have it work.
The step-by-step comment probably misses details (or has too much of them):
* Why do you mention GitLab (I don't see it on my page, and we are using Github)?
* Why repeating step 1 to 3?
* Is it explaining how you debugged it, or how to obtain a correct token? If it's both please separate for clarity :pray: (or just focus on how to get the correct token).

It's nice that it works, but I would like anyone reading the PR to understand how it did (and have a bit of elements to debug if needed).

.github/workflows/continuous-integration.yml
Moomoo-Seukla: I am removing the line as discussed earlier. Just an oversight when writing the PR.

Moomoo-Seukla:
@pbougue : I have modified the comments to describe how to get a correct token.

sonar-project.properties
```diff
@@ -0,0 +1,15 @@
...
+# Project
+sonar.host.url=https://sonarcloud.io
+sonar.organization=canaltp
+sonar.projectKey=CanalTP_kirin
+sonar.projectName=kirin
+
+# Language
+sonar.language=py
+sonar.sources=kirin
+sonar.python.coverage.reportPaths=./coverage.xml
```
pbougue, line 12: I am having doubt on that from https://sonarcloud.io/dashboard?id=CanalTP_kirin&pullRequest=372, it returns a 0% coverage, which doesn't match at all codecov value (it's OK that it differs as files included are different, but 0% is not OK).
Maybe it requires a first run on `master` branch to work good, but we must at least check that after merge and if it's still 0% we will have to dig that specifically.

pbougue (commented):
The process described to get a token is way too complicated for this kind of need, and it's part of a larger tutorial (that is not so clear on the token part). So I am sure there is an easier way to get a working token and we should document that simpler process.

As for the code itself, if you want to dig later on coverage, it's OK to me.

pbougue (approved):
OK, the "tutorial" way is a bit clearer now in your comment.
I was aiming for a simpler and more conventional token generation in "My Account"/"Security" that works as well.

As for the coverage, let's just "QA" it when merging on master and then work on it again if needed (if 0%).

PR #379, opened 2021-06-09
[doc] update readme to switch from COTS to PIV

pbougue:


woshilapin (approved):
I'm a little worried about that PR since it removes all the documentation about COTS... but all the code relative to COTS is still there. Any reason for cleaning up the documentation and not the code?

ArnaudOggy (approved):

pbougue:
The reason is code is still used in artemis and at least tests that exchanges with navitia stay correct (testing the way COTS is processed is still done, but no more useful).
So yes, this just manages one step of COTS deprecation (the other steps comes with moving tests from COTS to GTFSRT in Artemis).

woshilapin:
> @woshilapin: The reason is COTS code is still used in artemis and at least tests that exchanges with navitia stay correct (testing the way COTS is processed is still done, but no more useful).
> So yes, this just manages one step of COTS deprecation (the other steps comes with moving tests from COTS to GTFSRT in Artemis sweat ).

Then isn't removing documentation before the code a bit weird. We could just leave around (maybe in the arcane of the `README` with big WARNING flag about deprecation)?

pbougue:
I agree that it would be cleaner for code/doc management.
But the documentation is already quite big, and letting unused and deprecated part (just waiting to be removed) would be confusing for new users or new developers, and the work is not worth it.
(and this is best effort as I was just improving bits of doc on the go)

woshilapin (approved):

PR #382, opened 2021-07-13
Add tests to ensure the expected behaviour of Kirin's API

vchabot:
Adding some tests to ensure the expected behaviour of Kirin's API:
- If Redis and RabbitMQ are down, Kirin should still work for the endpoints that don't need Redis nor RabbitMQ.
- The endpoints where Kirin needs Redis or RabbitMQ will throw an error.

This Pull Request adds tests to ensure that this behaviour is expected and should always be expected with this situation.

Ticket: ND-1543


tests/integration/test_end_point.py
```diff
@@ -222,6 +218,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
...
+    with pytest.raises(ConnectionError):
+        tester.post("/gtfs_rt/{}".format(GTFS_CONTRIBUTOR_ID), data=basic_gtfs_rt_data)
+
+    call_root()
+    call_contributors()
+    call_health_with_ko_and_503_response()
+    call_status()
+
+
+def call_root():
```
woshilapin, line 273: Since all this functions `call_*()` actually do assertions and may raise an `AssertionError`, I would usually name them with an `assert_` prefix or something similar.

woshilapin (approved):
Nice factorization of all the assertion methods and nice to see that much new tests!

tests/integration/test_end_point.py
vchabot: :+1:  I make the changes

tests/integration/test_end_point.py
```diff
@@ -222,6 +218,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
...
+    assert "navitia_url" in resp
+    assert "last_update" in resp
+    assert GTFS_CONTRIBUTOR_DB_ID in resp["last_update"]
+    assert "2015-11-04T08:02:00Z" in resp["last_update"][GTFS_CONTRIBUTOR_DB_ID]
+    assert COTS_CONTRIBUTOR_DB_ID in resp["last_update"]  # config does not affect status anymore
+    assert PIV_CONTRIBUTOR_ID in resp["last_update"]
+
+
+@pytest.fixture()
+def basic_gtfs_rt_data():
```
pbougue, line 309: It would be nicer to have it mutualized in `tests/integration/utils_test.py` (to use it in `gtfs_rt_test.py` too).

tests/integration/test_end_point.py
```diff
@@ -311,3 +425,14 @@ def setup_database():
         model.db.session.add(rtu_piv)
 
         model.db.session.commit()
+
+
+@pytest.fixture(scope="function", autouse=True)
+def navitia(monkeypatch):
```
pbougue, line 431: Not sure if it's just me having old habits: I'd move monkeypatch magic at the top of the file like for other tests to have it consistent (and primarily to emphasize that magic is being done "before" running tests :mage:).

tests/integration/test_end_point.py
```diff
@@ -222,6 +218,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
         assert "navitia_url" in resp
 
 
+def test_status_with_redis_and_rabbitmq_down(setup_database):
+    kirin.redis_client.connection_pool = ConnectionPool(host="wrong_url_status")
+    kirin.rmq_handler = RabbitMQHandler("wrong_rabbit_url", "wrong_exchange", "wrong_exchange_type")
+    assert_call_status_200()
+
+
+def test_health_with_redis_and_rabbitmq_down():
```
pbougue, line 227: Should you `setup_database` here too? As no db also makes /health down.

tests/integration/test_end_point.py
```diff
@@ -222,6 +218,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
...
+    kirin.redis_client.connection_pool = ConnectionPool(host="wrong_url_health")
+    kirin.rmq_handler = RabbitMQHandler("wrong_rabbit_url", "wrong_exchange", "wrong_exchange_type")
+    assert_call_health_with_ko_and_503()
```
pbougue, line 230: For all those, it could be nice to separate tests between rabbit and redis unplugs (it's likely that redis only doesn't break /health currently, which is OK for now).

tests/integration/test_end_point.py
```diff
@@ -222,6 +218,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
...
+    with pytest.raises(ConnectionError):
+        tester.post("/gtfs_rt/{}".format(GTFS_CONTRIBUTOR_ID), data=basic_gtfs_rt_data)
+
+    assert_call_root_200()
+    assert_call_contributors_200()
+    assert_call_health_with_ko_and_503()
+    assert_call_status_200()
+
+
+def assert_call_root_200():
```
pbougue, line 273: Minor (do it only if you want) : I'd be more specific about the call (same on all).
```suggestion
def assert_get_root_200():
```

pbougue (commented):
Nice to add all those tests!

We should probably discuss it a bit on how to move on, but for rabbit, this PR only tests lack of navitia-rabbit, not celery-rabbit (I'm not sure it's easy to auto-test that, but hand-test is doable if not already done).

tests/integration/test_end_point.py
Moomoo-Seukla: The tests I have run are not conclusive.
By removing the line that should do a 503 error, the test remains OK when it should not
 @pbougue can you look at

tests/integration/test_end_point.py
pbougue: For the record (not pushed yet) I chose to put it back into `gtfs_rt_test.py` to be more consistent with what's done elsewhere in tests. But it still removes redundant fixture :+1:

tests/integration/test_end_point.py
pbougue: After some digging, I pushed a lot of changes in https://github.com/CanalTP/kirin/pull/382/commits/e4e731764b44751ab2edd3d33557ab95b336d4e3 (read commit message too).
It's not perfect but probably OK for the need, needs thorough review to be sure that we at least share understanding.

I resolve this conversation (tests are down separately and joined), feel free to reopen this or a new one.

pbougue (approved):
We all are judges and contributors on this one :grin:

Moomoo-Seukla (approved):
@vchabot : I let you merge if the changes made by me and @pbougue are OK

PR #383, opened 2021-07-13
[tech] Delegate purge of TripUpdate entirely to PG

Moomoo-Seukla:
Use cascade deletions on the table vehicle_journey.

[ND-1397](https://jira.kisio.org/browse/ND-1397)


kirin/core/model.py
```diff
@@ -401,15 +401,14 @@ def find_by_contributor_period(cls, contributors, start_date=None, end_date=None
...
 
     @classmethod
-    def remove_by_contributors_and_period(cls, contributors, start_date=None, end_date=None):
-        trip_updates_to_remove = cls.find_by_contributor_period(
-            contributors=contributors, start_date=start_date, end_date=end_date
+    def remove_by_contributors_until(cls, contributors, until=None):
+        sub_query = (
+            db.session.query(VehicleJourney.id)
+            .join(TripUpdate, cls.vj_id == VehicleJourney.id)
+            .filter(cls.contributor_id.in_(contributors), VehicleJourney.start_timestamp < until)
```
woshilapin, line 408: Is it ok to compute `VehicleJourney.start_timestamp < until` if `until` is `None` (default value in the function's parameters).

woshilapin (approved):

kirin/core/model.py
Moomoo-Seukla: I remove the default value in function

pbougue: I think it would be better to let the params, but test them and filter only if they are `not None`.

kirin/core/model.py
```diff
@@ -401,15 +401,14 @@ def find_by_contributor_period(cls, contributors, start_date=None, end_date=None
         return query.all()
 
     @classmethod
-    def remove_by_contributors_and_period(cls, contributors, start_date=None, end_date=None):
-        trip_updates_to_remove = cls.find_by_contributor_period(
-            contributors=contributors, start_date=start_date, end_date=end_date
+    def remove_by_contributors_until(cls, contributors, until):
+        sub_query = (
+            db.session.query(VehicleJourney.id)
+            .join(TripUpdate, cls.vj_id == VehicleJourney.id)
```
pbougue, line 407: How about an outerjoin like what's done on real_time_update purge sub_query?
Looks like the outer key is used which is what we want, and seems more robust (no code "duplication" of outerkey)

kirin/core/model.py
pbougue: Could you separate in 2 `filter()` calls like in RTU purge for readability?

pbougue (commented):
Can you update the main comment of your PR and add reference (or link) to the related ticket?

Nice to have those two purges smart and consistent now :+1:

kirin/core/model.py
Moomoo-Seukla: This is I wanted, my opinion the outer-join is not necessary because all key-fields are populate, they are never null.The inner join is enough.

kirin/core/model.py
```diff
@@ -401,15 +401,14 @@ def find_by_contributor_period(cls, contributors, start_date=None, end_date=None
         return query.all()
 
     @classmethod
-    def remove_by_contributors_and_period(cls, contributors, start_date=None, end_date=None):
-        trip_updates_to_remove = cls.find_by_contributor_period(
-            contributors=contributors, start_date=start_date, end_date=end_date
+    def remove_by_contributors_until(cls, contributors, until):
+        sub_query = (
```
pbougue, line 405: Can you also add a comment that it will cascade on `trip_update`, that will itself cascade on `associate_realtimeupdate_tripupdate` and `stop_time_update` ?
I wonder if we could add some tests + log on it really cascading, too.

kirin/core/model.py
pbougue: :+1: Maybe we can still remove the join condition, though: to be checked.

kirin/core/model.py
pbougue: After discussion:
* comment to be added
* no need for a specific cascade check in purge method (code would break in other places before)
* add some asserts in tests to check that cascade is done

Moomoo-Seukla: @pbougue  : A test of deleting on cascade exist , see `test_purge_trip_and_rt`. It's enough or a new test is necessary ?

pbougue: Well spotted.
Can you just count the rows in the 5 tables before any purge (add 3 asserts)?
It would be nice to have objects in all tables before purge, so add one/some stop_time_update when filling db.

tests/integration/model_test.py
```diff
@@ -142,12 +138,16 @@ def test_find_activate():
...
+            "70866ce8-0638-4fa1-8556-1ddfa22d09d4",
+            "70866ce8-0638-4fa1-8556-1ddfa22d09d5",
+        ]
         assert len(rtu) == 3
-        assert rtu[0].vj_id == "70866ce8-0638-4fa1-8556-1ddfa22d09d3"
-        assert rtu[1].vj_id == "70866ce8-0638-4fa1-8556-1ddfa22d09d4"
-        assert rtu[2].vj_id == "70866ce8-0638-4fa1-8556-1ddfa22d09d5"
+        result = [rtu[0].vj_id, rtu[1].vj_id, rtu[2].vj_id]
+        result.sort()
+        assert bool(result == expected_result) == True
```
pbougue, line 150: Wouldn't this work? (same below)
```suggestion
        assert rtu.sort() == expected_result
```
The bool cast and the intermediate result look useless.

tests/integration/tasks_test.py
```diff
@@ -33,6 +33,8 @@ def test_purge_trip_and_rt(mock_rabbitmq):
 
         # Check there's really something before purge
         assert TripUpdate.query.count() == 1
+        assert VehicleJourney.query.count() == 1
+        assert StopTimeUpdate.query.count() == 3
```
pbougue, line 37: Can you please also add the third assert, as requested https://github.com/CanalTP/kirin/pull/383/files#r679204764 ?

pbougue (approved):

PR #384, opened 2021-07-13
Add redis rabbitmq status

vchabot:
Add Redis and RabbitMQ status to the `/status` route.

- [X] add Redis status
- [x] add RabbitMQ status
- [ ] ~update `/health` to add the Redis status (Not sure if it's a good idea)~
- [ ] ~update `/health` to add the RabbitMQ status~
- [x] update `/status` documentation

For Redis, I'm just using the `ping` feature from Redis ([link here, scroll to "Check if Redis is working"](https://redis.io/topics/quickstart)), to check that the Redis server is well responding. For the `is_active` key, I set a key with a value, I try to get it and delete it. If everything went fine, I assume the Redis server is active.

Master ticket: ND-565
Slave ticket for Redis: ND-1545
Slave ticket for RabbitMQ: ND-1544


kirin/utils.py
```diff
@@ -361,3 +361,27 @@ def log_exception(exception, exception_source):
...
+        return redis_client.ping()
+    except Exception:
+        return False
+
+
+def check_redis_activity():
+    try:
+        redis_client.set("check_connection", "OK")
+        result = redis_client.get("check_connection") is not None
+        redis_client.delete("check_connection")
```
woshilapin, line 377: I'm guessing there is a risk that this is never deleted if an exception happen on line 375 and leave Redis with some objects we don't want? However, I'm not sure if this is really possible to make the 3 lines above act as an Atomic Transaction.

vchabot: Yes we have a potential risk for this. But when the Redis server will be back, the next time we will request the `/status` endpoint, this key will be removed from the Redis server. 
I think it won't be a big problem as there is only one key on the Redis server.

woshilapin: I understand better. Indeed, it shouldn't be a problem.

woshilapin:
Don't forget to update the documentation that just had been merged in https://github.com/CanalTP/kirin/blob/master/documentation/kirin_web_api.md#status-get. Also, I'm wondering if Redis should also be part of the response for `/health`?

woshilapin:
After some discussion with @vchabot, here is a summary. `/health` should probably be `KO` when the Celery RabbitMQ is down. However, it's not yet clear for Reddis.

When Reddis is down, Kirin will continue to be functional (see #382). However, this is not a normal situation either so returning `KO` would help to notify that something is wrong. Almost feel like we should invent something in-between `OK` and `KO` (like `Unstable`?). Any opinion on that @pbougue?

tests/integration/test_end_point.py
```diff
@@ -67,6 +67,9 @@ def test_status(setup_database):
     resp = api_get("/status")
 
     assert "version" in resp
+    assert "redis" in resp
+    assert "connection" in resp["redis"]
+    assert "is_active" in resp["redis"]
```
pbougue, line 72: With the tests added below on the value, it's redundant (value tests would break), so I think we should remove those lines.

tests/integration/test_end_point.py
```diff
@@ -85,8 +91,26 @@ def test_status(setup_database):
...
 
-    assert "rabbitmq_info" in resp
-    assert "password" not in resp["rabbitmq_info"]
+    assert "celery_rabbitmq" in resp
+    assert "connection" in resp["celery_rabbitmq"]
+    assert "is_active" in resp["celery_rabbitmq"]
+    assert "output_rabbitmq" in resp
+    assert "connection" in resp["output_rabbitmq"]
+    assert "is_active" in resp["output_rabbitmq"]
+    assert "info" in resp["output_rabbitmq"]
```
pbougue, line 100: We should probably just test values (which would test existence too).

tests/integration/test_end_point.py
```diff
@@ -85,8 +91,26 @@ def test_status(setup_database):
...
+    assert "password" not in resp["output_rabbitmq"]["info"]
+
+
+def test_status_with_redis_down():
+    resp = api_get("/status")
+    assert "redis" in resp
```
pbougue, line 106: Same here and below: it's "redundant" with value tests.

tests/integration/test_end_point.py
```diff
@@ -222,6 +242,124 @@ def test_status_with_database_and_navitia_ko(setup_database):
...
+
+
+def assert_call_health_with_ko_and_503():
+    resp, status_code = api_get("/health", False)
+    assert 503 == status_code
+    assert "message" in resp
+    assert "KO" in resp["message"]
+
+
+def assert_call_status_200():
```
pbougue, line 319: To be completed now that more info are available, no? (tests in this PR will be clearer after merge of #382  and rebase).

pbougue:
As for the discussion on `/health` being impacted when redis/celery are KO, I chose to separate tasks, as it's not so obvious indeed.

Maybe we should just improve doc for now:
The goal of `/health` is to tell if kirin's webservice is working. Not all of kirin's services (celery, piv-worker, reloader are out of the scope). It's then used by LB to direct web traffic.
The goal of `/status` is to display information on all of kirin's services, though.

From that, critical use of webservice is "only" on `/health` (of course) and on endpoints processing data (POST to `/cots` was the only way to process SNCF's feeds). So `/health` should not be dependent of celery, but it is depending on redis (locks, cache when processing feed), navitia-webservice and PG.

There are 2 gray areas (that may lead to a third status for health: `use-only-if-no-better-available` :smile: ):
* on output-rabbitmq (feeding navitia after feed processing). It is not critical as data is not lost, but it's not provided right away to Navitia if output-rabbitmq is KO. But a Navitia full-reload would work and consider processed data (if at that time, the reloader service is fully plugged).
* on `/status` itself, as it works and gives relevant info when redis/navitia are KO, but in those cases `/health` being KO would have LB "deactivate" node(s). It's acceptable IMO, as we may still request nodes directly (less convenient).

Moomoo-Seukla (approved):

documentation/kirin_web_api.md
```diff
@@ -38,18 +38,35 @@ For details on errors or statuses... check `/status` :-)
...
+  - is_active: test the Redis connection by adding a key, getting that key, and deleting it
+- output_rabbitmq: state of the rabbitmq connection
+  - info: dictionnary containing connection information
+  - connection: state of the rabbitmq connection
+- navitia: information for the navitia connection
+  - url: root url of the navitia server used to consolidate real-time information received by Kirin.
+  - connection: state of the connection to request navitia about base-schedule (condition for /health to be "OK")
+- postgres_db: state of the postgres database
+  - pool_status: state of the pool
+  - migration_version: migration version
```
pbougue, line 63: minor: not sure but it's more a db-schema version (we could also go for `alembic_version`)
I'd vote for:
```suggestion
  - schema_version: migration version
```
Any of the 3 (with `migration`) is fine, but if those ideas look good, go for it.

pbougue (commented):
:information_source: "minor" comments are up to your motivation/agreement, take them into consideration only if you feel like it.

This looks good, only remarks on the form.
:pray: Thanks for the :sweat: on celery and libs (and sad about the discoveries but glad to know).

pbougue (approved):

PR #387, opened 2021-07-27
Update pre-commit doc and black version

pbougue:
Did not upate `mypy` to latest version (tried) as it requires deps to be installed, which we do not want to do system-wide (and tricky to do otherwise).


CONTRIBUTING.md
```diff
@@ -25,13 +25,15 @@ Communication (request and read) with external services is done in UTC as much a
...
 
-The project provides a few git hooks that you should use to prevent any issues.
-Warning: they require **python 3.6+** to run (as opposed to Kirin itself, still using python 2).
+The project provides a few git hooks that you should use to prevent any issues.\
+Warning: they require **python3.8** to be available on the system (as opposed to Kirin itself, still using python 2.7).\
+[pyenv](https://github.com/pyenv/pyenv) may help you manage multiple python versions.
 
 These hooks are managed by [pre-commit](https://pre-commit.com/)
 that you need to [install](https://pre-commit.com/#install).
+[pre-commit](https://pre-commit.com/) itself requires python >=3.6.
```
woshilapin, line 34: This is not a typo, right?
```suggestion
[pre-commit](https://pre-commit.com/) itself requires python >=3.8.
```

woshilapin (approved):

CONTRIBUTING.md
pbougue: Nope, pre-commit requires 3.6+ and updated mypy 3.8.
Which is why it works when installing pre-commit in a 3.6 venv, with python3.8 available on the system (yes, pre-commit is quite twisted, creating another venv for its actions than the one on which pre-commit is installed).

Moomoo-Seukla (approved):

PR #389, opened 2021-08-13
acceleration status response

Moomoo-Seukla:
Delete useless index for status (status_idx). 
Optimization index for accelerated response of api status.

[ND-1560](https://jira.kisio.org/browse/ND-1560)


vchabot (approved):
Looks good to me!

kirin/core/model.py
```diff
@@ -458,8 +457,14 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
     )
 
     __table_args__ = (
-        db.Index("realtime_update_created_at", "created_at"),
-        db.Index("realtime_update_contributor_id_and_created_at", "created_at", "contributor_id"),
+        db.Index("realtime_update_created_at", sqlalchemy.text("created_at desc")),
+        db.Index(
+            "realtime_update_contributor_id_and_created_at",
```
pbougue, line 462: Why calling it only `[...]_contributor_id_and_created_at` when it's on `contributor_id`, `created_at` and **`status`** ?

Shouldn't we have 2 indexes here then: one for the request that's not filtered on status and one for the filter status==OK ? Then we would have a total of 3 indexes.
Maybe the index with status works fine for both requests (don't know if you checked that it's used in that case), but IMO it deserves a code-comment if it's the case (and an index renaming).

migrations/versions/27a3d7e96d1f_del_index_useless_optimize_index_real_time_update.py
```diff
@@ -0,0 +1,45 @@
...
+revision = "27a3d7e96d1f"
+down_revision = "1f13d13948dd"
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("realtime_update_contributor_id_and_created_at", table_name="real_time_update")
```
pbougue, line 20: It looks like dropping after creation could help the new index being created, no?
It shouldn't hurt much for disk space (and not for long), so I'd go for this (same for all).

migrations/versions/27a3d7e96d1f_del_index_useless_optimize_index_real_time_update.py
```diff
@@ -0,0 +1,45 @@
...
+# revision identifiers, used by Alembic.
+revision = "27a3d7e96d1f"
+down_revision = "1f13d13948dd"
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
pbougue, line 19: We are used to remove those comments once checked and adusted.

kirin/core/model.py
```diff
@@ -458,8 +457,14 @@ class RealTimeUpdate(db.Model, TimestampMixin):  # type: ignore
...
+            contributor_id,
+            status,
+            sqlalchemy.text("created_at desc"),
+            unique=False,
```
pbougue, line 466: It's the default value, so this explicit param is not necessary IMO (when reading this I was afraid all others implied uniqueness).

kirin/core/model.py
Moomoo-Seukla: There is no significant difference in performance between the two indexes. One index will do the job

migrations/versions/27a3d7e96d1f_del_index_useless_optimize_index_real_time_update.py
Moomoo-Seukla: I have done some tests and the order will not change the processing time

migrations/versions/27a3d7e96d1f_del_index_useless_optimize_index_real_time_update.py
pbougue: Please also remove the corresponding ones at the bottom of each section :pray:

kirin/core/model.py
pbougue: As mentioned, then can you add a code-comment to say it's efficient whether filter is done only on contrib, or on status and contrib, please ?

pbougue (approved):
Is there a reason for this PR to be in "Draft"?
Anyway, once switched to "ready for review", it should be OK to merge.

PR #390, opened 2021-08-27
Refactor the manage_consistency function

vchabot:


kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
     )
 
 
+def manage_arrival_consistency(stu, trip_update, previous_stop_event):
```
pbougue, line 363:
```suggestion
def manage_arrival_consistency(stu, trip_update, previous_stop_event):
"""
check and adjust arrival time and delay
returns False is feed is not consistent for that stop, True otherwise
"""
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+        return False
+    log_stu_modif(trip_update, stu, "arrival = {v}".format(v=stu.arrival))
+    if not stu.arrival_delay and stu.departure_delay:
+        stu.arrival_delay = stu.departure_delay
+        log_stu_modif(trip_update, stu, "arrival_delay = {v}".format(v=stu.arrival_delay))
+
+    return True
+
+
+def manage_departure_consistency(stu, trip_update):
```
pbougue, line 388: For consistency, we should return booleans in departure and arrival checks:
```suggestion
def manage_departure_consistency(stu, trip_update):
"""
check and adjust departure time and delay
returns False is feed is not consistent for that stop, True otherwise (currently always True)
"""
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    if stu.departure is not None:
+        return
```
pbougue, line 390:
```suggestion
        return True
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    stu.departure = stu.arrival
+    log_stu_modif(trip_update, stu, "departure = {v}".format(v=stu.departure))
+    if not stu.departure_delay and stu.arrival_delay:
+        stu.departure_delay = stu.arrival_delay
+        log_stu_modif(trip_update, stu, "departure_delay = {v}".format(v=stu.departure_delay))
+
```
pbougue, line 396:
```suggestion
return True

```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
-
-        if stu.departure is None:
-            stu.departure = stu.arrival
-            log_stu_modif(trip_update, stu, "departure = {v}".format(v=stu.departure))
-            if not stu.departure_delay and stu.arrival_delay:
-                stu.departure_delay = stu.arrival_delay
-                log_stu_modif(trip_update, stu, "departure_delay = {v}".format(v=stu.departure_delay))
+        if not manage_arrival_consistency(stu, trip_update, previous_stop_event):
+            return False
+        manage_departure_consistency(stu, trip_update)
```
pbougue, line 430: It can't happen, but I would probably do the same here (matching fn signature):
```suggestion
        if not manage_departure_consistency(stu, trip_update):
            return False
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    if not stu.departure_delay and stu.arrival_delay:
+        stu.departure_delay = stu.arrival_delay
+        log_stu_modif(trip_update, stu, "departure_delay = {v}".format(v=stu.departure_delay))
+
+
+def manage_deleted_stop_event(event_type, stu, previous_stop_event, trip_update):
+    # not considering deleted {event_type} (event_type can be departure or arrival)
+    if stu.is_stop_event_deleted(event_type):
+        return previous_stop_event
+    delay_field = "{}_delay"
```
pbougue, line 402: Do it only once (and remove other `format()`:
```suggestion
    delay_field = "{}_delay".format(event_type)
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    # if {event_type} is before previous stop-event's time:
+    # push {event_type} time so that its delay is the same than for previous time
```
pbougue, line 404:
```suggestion
    # if current stop-event is before previous stop-event's time:
    # push current stop-event time so that its delay is the same than for previous time
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    if previous_stop_event.time is not None and previous_stop_event.time > getattr(stu, event_type):
+        delay_diff = previous_stop_event.delay - getattr(stu, delay_field.format(event_type))
+        stu.departure_delay += delay_diff
+        setattr(stu, delay_field.format(event_type), (getattr(stu, delay_field.format(event_type)) + delay_diff))
```
pbougue, line 408: The second line is the new version of the first line, you need to remove it or you add delay twice (in the departure case)
```suggestion
        setattr(stu, delay_field.format(event_type), (getattr(stu, delay_field.format(event_type)) + delay_diff))
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+        setattr(stu, event_type, (getattr(stu, event_type) + delay_diff))
+        log_stu_modif(
+            trip_update,
+            stu,
+            "departure = {t} and departure_delay = {d}".format(
+                t=getattr(stu, event_type), d=getattr(stu, delay_field.format(event_type))
```
pbougue, line 414:
```suggestion
            "{e} = {t} and {e}_delay = {d}".format(e=event_type,
                t=getattr(stu, event_type), d=getattr(stu, delay_field.format(event_type))
```

kirin/core/merge_utils.py
```diff
@@ -360,37 +360,74 @@ def log_stu_modif(trip_update, stu, string_additional_info):
...
+    if stu.departure is not None:
+        return
+    stu.departure = stu.arrival
+    log_stu_modif(trip_update, stu, "departure = {v}".format(v=stu.departure))
+    if not stu.departure_delay and stu.arrival_delay:
+        stu.departure_delay = stu.arrival_delay
+        log_stu_modif(trip_update, stu, "departure_delay = {v}".format(v=stu.departure_delay))
+
+
+def manage_deleted_stop_event(event_type, stu, previous_stop_event, trip_update):
```
pbougue, line 398:
```suggestion
def manage_stop_event_order(event_type, stu, previous_stop_event, trip_update):
"""
adjust stop-event's time/delay so that it doesn't go back in time considering the previous stop-event
pushes to have delay equal to previous stop-event's delay in case of inconsistency
"""
```

pbougue (changes requested):
I think we should merge with the piv_worker part.

For the merge_utils part, I'm not sure it's worth correcting it, I let you see.
As the plan is to completely re-write it, we may as well forget that part and let it as-is.
And I'm glad tests were able to detect the mixup on departure_delay which is tricky :sweat_smile:

PR #391, opened 2021-09-01
create index concurrently

Moomoo-Seukla:
Change to avoid blocking the db when executing an upgrade/downgrade


migrations/versions/27a3d7e96d1f_del_index_useless_optimize_index_real_time_update.py
```diff
@@ -16,26 +16,46 @@
...
-    op.create_index("realtime_update_created_at", "real_time_update", [sa.text("created_at desc")], unique=False)
+    op.create_index(
+        "realtime_update_created_at",
+        "real_time_update",
+        [sa.text("created_at desc")],
+        unique=False,
+        postgresql_concurrently=True,
+    )
+    op.drop_index("realtime_update_contributor_id_and_created_at", table_name="real_time_update")
     op.drop_index("status_idx", table_name="real_time_update")
```
pbougue, line 38: Same for all drops: should we do it concurrently too?

pbougue (approved):
Looks good, to be tested :crossed_fingers:

PR #392, opened 2021-09-03
Filter TN and TNRER brand code from the PIV feed to avoid charge issues

vchabot:
Filter the PIV feeds and reject the "TN" and "TNRER" marque code.

Ticket: ND-1599


pbougue (commented):
Except for the litlle log improvment it's good for me :heavy_check_mark: 

We will then see how it behaves on dev for QA :mag:

kirin/piv/model_maker.py
```diff
@@ -255,6 +255,9 @@ def build_trip_updates(self, rt_update):
 
         dict_objects = get_value(json, "objects")
         json_train = get_value(dict_objects[0], "object")  # TODO: can we get more than 1 relevant in objects[]?
+        brand_code = json_train.get("marque", {}).get("code", "")
```
Moomoo-Seukla, line 258: To be uniform with the tests below. see L293
```suggestion
        brand_code = get_value(get_value(json_train, "marque"), "code", nullable=True)
```

Moomoo-Seukla, line 258: To be uniform with the tests below. see L293. Just suggest
```suggestion
        brand_code = get_value(get_value(json_train, "marque"), "code", nullable=True)
```

pbougue: I didn't check it closely at first, but raw `get_value()` is not usable here, as the fields are "nullable".
There is a "nullable" param available to the function, but on this part of the code we use `get()` or `get_value(nullable=True)` in nullable cases... So it's fine IMO.

We should be more consistent (which is more important that what option we choose IMO) about this, and I would tend to use `get()` on nullable cases (it was prefered by Patoche and Woshilapin too).

pbougue (approved):

PR #393, opened 2021-09-06
Try again the worker pings when celery could not reach any worker

vchabot:
Follow-up on #384

Ticket ref. ND-1599


kirin/utils.py
```diff
@@ -54,6 +54,10 @@
 from kirin.new_relic import must_never_log, record_exception
 
 
+class CeleryWorkerNotAvailableException(Exception):
+    pass
```
pbougue, line 58: Those go to `kirin/exceptions.py`. Or you might raise a `ValueError` to avoid creating a new `KirinException`, whatever suits you best.

pbougue (approved):

PR #396, opened 2021-09-07
ETag improvement to avoid useless request

vchabot:
Handle the ETag feature with the `If-None-Match` header.

You receive a 304 status-code response if the resource has not changed, and a 200 status-code with a new `ETag` response header when the resource has changed.

The `If-None-Match` header resource:
https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/If-None-Match

Ticket: ND-1548


Moomoo-Seukla (approved):

kirin/gtfs_rt/tasks.py
```diff
@@ -114,15 +84,55 @@ def _is_last_call_too_recent(func_name, contributor, minimal_call_interval):
     return False
 
 
+def _update_etag(resp, config):
```
pbougue, line 87:
```suggestion
def _update_redis_etag(resp, config):
```

Also, can you protect it to just log if this raises, please? This ETag management is a "bonus", and we want this to work even if redis is KO.

kirin/gtfs_rt/tasks.py
```diff
@@ -114,15 +84,55 @@ def _is_last_call_too_recent(func_name, contributor, minimal_call_interval):
...
+
+
 @new_relic.agent.function_trace()  # trace it specifically in transaction times
 def _retrieve_gtfsrt(config):
     start_dt = datetime.utcnow()
-    resp = requests.get(config["feed_url"], timeout=config.get("timeout", 1))
+    # Adding a If-None-Match header to check if the feed has changed or not
+    # https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/If-None-Match
+    etag_key = build_redis_etag_key(config["contributor"])
+    old_etag = redis_client.get(etag_key)
```
pbougue, line 99: If there is nothing in redis, we should not add the headers.
And this is true also if redis is unavailable or fails, we should still request properly here, as it's (quite) easy.

So this misses an try/except and a check that we get a proper value in redis IMO.

kirin/gtfs_rt/tasks.py
```diff
@@ -148,28 +158,5 @@ def gtfs_poller(self, config):
...
-                contributor_id=contributor.id,
-                error="Http Error",
-                is_reprocess_same_data_allowed=True,
-            )
-            logger.debug(six.text_type(e))
-            return
-
-        wrap_build(KirinModelBuilder(contributor), response.content)
+        _poll_gtfs_rt_feed(config, logger, contributor)
         logger.info("%s for %s is finished", func_name, contributor.id)
```
pbougue, line 162: Not your code, but on the way (these logs are heavy when looking for purge finish-logs):
```suggestion
        logger.debug("%s for %s is finished", func_name, contributor.id)
```

kirin/gtfs_rt/tasks.py
```diff
@@ -114,15 +84,55 @@ def _is_last_call_too_recent(func_name, contributor, minimal_call_interval):
...
+            _update_etag(response, config)
+    except Exception as e:
+        manage_db_error(
+            data="",
+            connector_type=ConnectorType.gtfs_rt.value,
+            contributor_id=contributor.id,
+            error="Http Error",
+            is_reprocess_same_data_allowed=True,
+        )
+        logger.debug(six.text_type(e))
```
pbougue, line 130: Not fan of adding params for log only: I'd like more to rebuild it only when needed here.

pbougue (commented):
This looks good, less code.
(As for being resilient to redis, I'm not sure it was the case before, but I'm pretty sure we want it if possible).

tests/integration/gtfsrt_tasks_test.py
```diff
@@ -0,0 +1,141 @@
...
+            _poll_gtfs_rt_feed(config)
+            assert len(RealTimeUpdate.query.all()) == 2
+            assert len(TripUpdate.query.all()) == 1
+            assert len(StopTimeUpdate.query.all()) == 4
+            assert "etag2" == kirin.redis_client.get(etag_key)
+
+
+def test_poller_with_no_etag_header():
+    responses_list = [
+        {"content": basic_gtfs_rt_data(), "status_code": 200},
```
pbougue, line 100: It would be nice to chain things more: here add a delay still with no ETag

tests/integration/gtfsrt_tasks_test.py
```diff
@@ -0,0 +1,141 @@
...
+        with app.app_context():
+            _poll_gtfs_rt_feed(config)
+            assert len(RealTimeUpdate.query.all()) == 1
+            assert len(TripUpdate.query.all()) == 1
+            assert len(StopTimeUpdate.query.all()) == 4
+            assert kirin.redis_client.get(etag_key) is None
+
+
+def test_poller_with_error_response():
+    responses_list = [{"status_code": 400}]
```
pbougue, line 123: Still on chaining tests : maybe check with correct responses before and after (check we recover too).

tests/integration/gtfsrt_tasks_test.py
```diff
@@ -0,0 +1,141 @@
...
+        "feed_url": "http://feed_url",
+        "retrieval_interval": "retrieval_interval",
+        "timeout": app.config.get(str("GTFS_RT_TIMEOUT"), 1),
+    }
+    etag_key = build_redis_etag_key(config["contributor"])
+    with requests_mock.Mocker() as m:
+        m.get("http://feed_url", response_list=responses_list)
+        with app.app_context():
+            _poll_gtfs_rt_feed(config)
+            assert len(RealTimeUpdate.query.all()) == 1
```
pbougue, line 138: minor: A bit astonished by that at first sight, but just to confirm (if you have the answer):
In that case we create a RTU with empty data to store error, right?

It's not properly related to ETag management, but to adding tests around polling:
Then we may test the content of the RTU a bit (`error`, `status` and `raw_data`).

pbougue (commented):
:heavy_check_mark: just final touch to tests.
And :clap:  for increasing code-coverage on that part of the code, thanks for the motivation !

tests/integration/gtfsrt_tasks_test.py
vchabot: Yes it's stored with a KO status.

You're right, adding some assertions to ensure that the status is KO will be more understandable.

pbougue (approved):

PR #400, opened 2021-09-15
Add CSRF protection to Kirin and more secure CORS settings

vchabot:
Add some configuration to be compliant with sonarcloud security hotspots.


kirin/__init__.py
```diff
@@ -40,7 +40,10 @@
 import logging.config
 
 app = Flask(__name__)
-CORS(app)
+CORS(app, resources={r"/*": {"origins": "*", "send_wildcard": "False"}})
```
Moomoo-Seukla, line 43: You cant add `methods` with [GET, POST, PUT ] see [here](https://flask-cors.corydolphin.com/en/latest/api.html#flask_cors.cross_origin)

kirin/__init__.py
```diff
@@ -40,7 +40,10 @@
 import logging.config
 
 app = Flask(__name__)
-CORS(app)
+CORS(app, resources={r"/*": {"origins": "*", "send_wildcard": "False", "methods": ["GET", "POST", "PUT"]}})
```
pbougue, line 43: There is DELETE verb on /contributors/<toto>

kirin/api.py
```diff
@@ -44,7 +46,10 @@
 
 flask_restful.representations.json.settings = {"indent": 4}  # we always want pretty json
 
-api = flask_restful.Api(app, catch_all_404s=True)
+csrf = CSRFProtect()
+csrf.init_app(app)
+
+api = flask_restful.Api(app, catch_all_404s=True, decorators=[csrf.exempt])
```
pbougue, line 52: As discussed: we may just aknowledge sonar alert and add a code-comment on the line sonar warns about.
As we are not managing any authentication, the CSRF attack is not relevant in Kirin's case.

The auth would currently be managed on tools wrapping kirin's webservice (it's not planned at all to ever manage auth in kirin).

kirin/api.py
```diff
@@ -34,6 +34,7 @@
 from flask.signals import got_request_exception
 import flask_restful
 from flask import request
+
```
pbougue, line 37: minor: useless change

kirin/__init__.py
```diff
@@ -39,8 +39,12 @@
 from flask_cors import CORS
 import logging.config
 
+# As Kirin is an API without authentication, we don't need to add CSRF protection
 app = Flask(__name__)
-CORS(app)
+CORS(
+    app,
+    resources={r"/*": {"origins": "*", "send_wildcard": "False", "methods": ["GET", "POST", "PUT", "DELETE"]}},
```
pbougue, line 46: Sonar still complains about it, which I can understand, as we finally just made things more explicit, but equal to what we had.
* We may try to detail a bit by path what method is allowed or not, it would be a bit more secure.
* We may also (not exclusive) precise origins, but it would require some dinging (Kisio possible origins), and probably outsource this part to conf (this is opensource).
* Finally, we may add some comment on why we did "only that (depending on where we stop)" and then acknowledge the sonar warning if still here (would be nice to keep it visible though, but not have it buzz at every PR).

pbougue (approved):

PR #402, opened 2021-09-22
Change error and avoid useless critical logs when a feed is not well formed

vchabot:
Ticket: ND-1646

Fixing the `unsupported operand type(s) for -: 'NoneType' and 'datetime.timedelta'` error when a feed has only brand new stops on a base-schedule trip.


tests/integration/piv_test.py
```diff
@@ -2118,3 +2119,81 @@ def test_piv_feed_is_rejected_with_tn_brand_code(input):
         assert "invalid arguments" in res.get("message")
     else:
         assert "PIV feed processed" in res.get("message")
+
+
+def test_piv_fails_with_no_base_schedule_stops_in_a_creation():
+    piv_feed = _get_stomp_20201022_23187_stop_time_added_at_the_beginning_fixture()
```
pbougue, line 2125: Can't you use `get_fixture_data_as_dict("piv/stomp_20201022_23187_blank_fixture.json")` directly?

tests/integration/piv_test.py
```diff
@@ -2118,3 +2119,81 @@ def test_piv_feed_is_rejected_with_tn_brand_code(input):
...
+    piv_feed["objects"][0]["object"]["listeArretsDesserte"]["arret"] = [
+        {
+            "depart": {
+                "numeroCirculation": "KHAN64",
+                "dateHeure": "2021-09-19T09:36:00+02:00",
+                "dateHeureReelle": "2021-09-19T09:36:00+02:00",
+                "typeAffichage": "HORAIRE",
+                "indicateurTempsReel": True,
+                "statutModification": "CREATION",
+                "planTransportSource": "OPE",
```
pbougue, line 2149: minor: To replicate inital "bug", it would probably be enough to change fixture on just those 2 fields at all stops.

pbougue (approved):

Moomoo-Seukla (approved):

pbougue (approved):

PR #404, opened 2021-09-29
Request only base-schedule VJ from Navitia

pbougue:
Just explicitly do it for COTS and GTFSRT (it was only implicit before).
For PIV, add this to request, using since/until, as it's the only way currently.

JIRA: https://jira.kisio.org/browse/ND-1665


vchabot (approved):
LGTM :shipit:

Moomoo-Seukla (approved):

PR #405, opened 2021-10-01
Remove old flask libraries

vchabot:
Remove old python2 like libraries used by Flask to the new generation :new:


pbougue:
Fully comprised in #406

PR #419, opened 2021-11-23
docs: 'virtualenv' must be installed for 'make test'

woshilapin:
We could also just add a dependency in `requirements_dev.txt`, I can go either way.


tests/readme.md
```diff
@@ -2,7 +2,11 @@
 
 ## Run
 
-To run the tests, please setup dependencies described in [main README](../readme.md#setup).
+To run the tests, please setup the following:
+
+* dependencies described in [main README](../readme.md#setup)
+* `virtualenv` must be installed (used in `tests/launch_tests.sh`)
```
vchabot, line 8: Perhaps adding a bit of code at the end to easily copy/paste the command to install `virtualenv`?

woshilapin: Done in 44cda1c.

vchabot (approved):
LGTM :shipit:

pbougue (approved):
I like it like that, as it's required only if you launch tests using the script.

PR #422, opened 2021-11-26
tests: do not redefine the same fixture everywhere

woshilapin:
Use dependencies in fixtures of `pytest` to achieve factorization of code.


pbougue (approved):
Looks better :mushroom: 
Still not one for all, but still modular :+1:

vchabot (approved):
:heart_eyes:

PR #424, opened 2021-11-26
refactor: inline some useless functions

woshilapin:
To be merged after #421 (since most of the cleanup impact what has been done there).


kirin/command/piv_worker.py
```diff
@@ -32,11 +32,12 @@
 
 
 from kirin import app, new_relic
+from kirin.core import model
```
pbougue, line 35: Minor, to shorten code a bit (and keep it clear):
To be done the same way everywhere IMO, so it may not be a good idea if `model` is used "from scratch" in a lot of places.
```suggestion
from kirin.core.model import Contributor
```

pbougue (approved):

kirin/command/piv_worker.py
woshilapin: Done in a81f9d8.

woshilapin: Done (better) in 6689ed1.

pbougue (approved):

vchabot (approved):

PR #425, opened 2021-11-30
fix: remove MonkeyPatchWarning error

woshilapin:
Remove the following warning

```
/usr/src/app/kirin/__init__.py:76: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['newrelic.packages.urllib3.util (/usr/local/lib/python3.8/site-packages/newrelic/packages/urllib3/util/__init__.py)', 'urllib3.util.ssl_ (/usr/local/lib/python3.8/site-packages/urllib3/util/ssl_.py)', 'urllib3.util (/usr/local/lib/python3.8/site-packages/urllib3/util/__init__.py)'].
```

The hint to fix it was [here](https://github.com/gevent/gevent/issues/1016#issuecomment-328529454).


kirin/__init__.py
```diff
@@ -32,15 +32,8 @@
...
 
-from flask_migrate import Migrate
-
-from kirin import exceptions
-
-
 from flask import Flask
 from flask_cors import CORS
-import logging.config
-from kirin import command
```
woshilapin, line 43: Note: this has been removed (not moved) because we already import `command.piv_worker` and `command.load_realtime` below.

kirin/__init__.py
```diff
@@ -32,15 +32,8 @@
...
 import sys
 
-from flask_migrate import Migrate
-
-from kirin import exceptions
-
-
 from flask import Flask
 from flask_cors import CORS
-import logging.config
```
pbougue, line 42: Don't we risk a complete crash instead of a log if we import it after something that's crashing?
For example, if there is an issue with `KIRIN_CONFIG_FILE`

pbougue (commented):
Maybe a discussion about the moves could be useful here :sweat_smile:

kirin/__init__.py
woshilapin: Where exactly is called `purge_rt`?

woshilapin: So, `purge_rt` has been removed (poor API, unused command) but `purge_contributor` has been reactivated with `import kirin.command.purge` (the file has been moved from `purge_rt.py` to `purge.py`).

kirin/__init__.py
woshilapin: Moved it back to where it was, doesn't seems to trigger the warning.

pbougue (approved):

kirin/__init__.py
```diff
@@ -71,10 +63,14 @@
         del sys.modules["threading"]
     # end of conflict's patch
 
+    # MUST be imported before `requests`, ideally, import it as soon as possible
+    # https://www.gevent.org/api/gevent.monkey.html
```
vchabot, line 67: :+1: 
Good to know, indeed.

vchabot (approved):

PR #426, opened 2021-12-01
feat: remove useless field in 'real_time_update'

woshilapin:
Remove the use of `RealTimeUpdate.connector` everywhere. The `connector_type` is available in `Contributor`.

PR #428, opened 2021-12-02
feat: remove useless field in 'real_time_update' (db migration)

woshilapin:


migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
```diff
@@ -0,0 +1,33 @@
...
+
+Revision ID: 50975c469693
+Revises: 1639faa28bc4
+Create Date: 2021-12-01 17:00:00.0
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "50975c469693"
+down_revision = "1639faa28bc4"
```
pbougue, line 12: As this may be merged later (after another migration), we must pay attention to that down revision that is likely to change.

pbougue (approved):

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
```diff
@@ -0,0 +1,33 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.execute("ALTER TABLE real_time_update DROP connector")
+
+
+def downgrade():
+    op.execute("ALTER TABLE real_time_update ADD COLUMN connector connector_type")
```
Moomoo-Seukla, line 23: In the last version, column `connector` can't be `null`, I think you forget add this constraint after apply `UPDATE`. What do think ?

woshilapin: Indeed, done in e27202b.

pbougue: Hmm, it makes me think that the SQL part of having the column nullable (and not nullable on downgrade) should probably be done in previous PR.
Because in previous PR, the column is already left empty.

Moomoo-Seukla: Do something like after UPDATE `alter table real_time_update alter column connector set not null;` 
'''

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
```diff
@@ -0,0 +1,33 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.execute("ALTER TABLE real_time_update DROP connector")
+
+
+def downgrade():
+    op.execute("ALTER TABLE real_time_update ADD COLUMN connector connector_type NOT NULL")
```
pbougue, line 23: It can only be done after (once filled) and more importantly, it must be done in previous PR (that already stops filling the column)

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
woshilapin: > Hmm, it makes me think that the SQL part of having the column nullable (and not nullable on downgrade) should probably be done in previous PR.
Because in previous PR, the column is already left empty.

Should be fixed by https://github.com/CanalTP/kirin/pull/431.

pbougue (changes requested):

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
woshilapin: > Do something like after UPDATE alter table real_time_update alter column connector set not null;


Right, I don't know what I was thinking. Should be better with this d5c595a205ce6f8b557dd23110b552d0ce951c6b.

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
woshilapin: Should be fixed by d5c595a.

woshilapin: But for #426, do you think we should do an intermediate migration where we remove the `NOT NULL` constraint?

pbougue: Yes, otherwise after merging #426 you are not filling a column that can't be null :boom:

Moomoo-Seukla (approved):

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
woshilapin: I removed the `NOT NULL` constraint in the downgrade since this constraint should have disappear in https://github.com/CanalTP/kirin/pull/426 (see https://github.com/CanalTP/kirin/pull/426#issuecomment-985579549).

migrations/versions/50975c469693_no_connector_type_in_real_time_update.py
```diff
@@ -0,0 +1,33 @@
...
+    op.execute(
+        "\
+        UPDATE real_time_update \
+            SET connector = ( \
+                SELECT c.connector_type \
+                FROM contributor AS c \
+                WHERE contributor_id = c.id \
+            ) \
+            WHERE connector IS NULL"
+    )
```
pbougue, line 33: I just thought of it: actually, this should also be done just before setting column as `NOT NULL` too.
So on #426 too.
:sweat_smile:

woshilapin: Indeed, removed from the current PR in 1843305 and moved to PR #426 in 1dddbd6.

PR #429, opened 2021-12-02
Add a new connector: siri-et-xml-tn

vchabot:
Ticket : ND-1732

Add a new connector: `siri-et-xml-tn`.

Based on the piv connector, which has a worker and consume a RabbitMQ queue.

TODO:
- [x] Add tests
- [x] Refactor the common things between this connector and the `piv` connector


kirin/workers/__init__.py
```diff
@@ -156,71 +162,78 @@ def on_iteration(self):
...
+    with get_lock(logger, lock_name, WORKER_REDIS_TIMEOUT_LOCK.total_seconds()) as locked:
         if locked:
             should_wait = True
             contributor = None
             try:
-                contributor = pick_piv_contributor()
+                contributor = pick_first_contributor_by_connector_type(logger, connector_type)
                 if not contributor:
-                    logger.warning("no PIV contributor")
+                    logger.warning("no SIRI_ET_XML_TN contributor")
```
pbougue, line 189:
```suggestion
                    logger.warning("no contributor for {}".format(connector_type)
```

kirin/workers/http_utils.py
```diff
@@ -0,0 +1,51 @@
...
+    contributor = (
+        Contributor.query_existing().filter_by(id=contributor_id, connector_type=connector_type).first()
+    )
+    if not contributor:
+        abort(404, message="Contributor '{}' not found".format(contributor_id))
+
+    raw_xml = get_request_data(flask.globals.request, connector_type)
+
+    if connector_type == ConnectorType.siri_et_xml_tn.value:
+        from kirin.workers.siri_et_xml_tn.siri_et_xml_tn_model_maker import KirinModelBuilder
```
pbougue, line 44: Looks like a circular dependency, probably better to ask for a method in class (or a new param IMO).

pbougue (commented):
I did not review it all, but wanted to avoid case where code moves too much during review and if needed other review will add comments.

kirin/api.py
```diff
@@ -51,6 +52,10 @@
 api.add_resource(cots.Cots, "/cots", endpoint="cots")
 api.add_resource(piv.PivIndex, "/piv", endpoint="piv_index")
 api.add_resource(piv.Piv, "/piv/<string:contributor_id>", endpoint="piv")
+api.add_resource(siri_et_xml_tn.SiriEtXmlTnIndex, "/siri_et_xml_tn", endpoint="siri-et-xml-tn_index")
+api.add_resource(
+    siri_et_xml_tn.SiriEtXmlTn, "/siri_et_xml_tn/<string:contributor_id>", endpoint="siri-et-xml-tn"
+)
```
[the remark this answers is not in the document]
vchabot: I remove them for now. I had to keep the dash and not the underscore to be fully automatic with the `connector_type` argument below (which uses dash and not underscore). We will have the time when getting them back in the code.

kirin/__init__.py
```diff
@@ -88,8 +88,9 @@
 
 # activate a command
 import kirin.command.load_realtime
-import kirin.command.piv_worker
+import kirin.workers.piv.piv_worker
```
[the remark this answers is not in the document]
vchabot: I'm not a fan of names based on a technology, but I'll make an exception for this one :)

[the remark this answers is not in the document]
pbougue: queue_workers and poller_workers ?

kirin/workers/http_utils.py
vchabot: As I removed the siri_et_xml_tn API, this is no longer an issue (for now, this will be an issue when we will add the siri_et_xml_tn and siri_et_lite_idfm APIs).

kirin/workers/__init__.py
```diff
@@ -156,71 +162,78 @@ def on_iteration(self):
...
-    # Therefore, we enforce a maximum of one `PivWorker` is running at a time
-    while True:
-        try:
-            attempt_piv_worker_launch()
-        except LockError as lock_e:
-            logger.warning("Exception with lock : {0}".format(lock_e))
+def get_worker_class(connector_type: str):
+    if connector_type == ConnectorType.piv.value:
+        module = "kirin.workers.piv.piv_worker"
+        return import_module(module), "PivWorker"
```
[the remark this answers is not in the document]
vchabot: Done :+1:

tests/integration/piv_worker_test.py
```diff
@@ -115,7 +114,8 @@ def launch_piv_worker(pg_docker_fixture):
         # re-init the db by overriding the db_url
         reinit_flask_db(pg_docker_fixture)
 
-        contributor = get_contributor_by_connector_type(PIV_CONTRIBUTOR_ID, ConnectorType.piv.value)
+        contributors = Contributor.find_by_connector_type(ConnectorType.piv.value)
+        contributor = [c for c in contributors if c.id == PIV_CONTRIBUTOR_ID][0]
```
pbougue, line 118: You can probably do a bit more pythonic, like:
```suggestion
        contributor = next(filter(lambda c: c.id == PIV_CONTRIBUTOR_ID, contributors))
```

tests/integration/siri_et_xml_tn_worker_test.py
```diff
@@ -123,10 +121,8 @@ def launch_siri_et_xml_tn_worker(pg_docker_fixture):
         # re-init the db by overriding the db_url
         reinit_flask_db(pg_docker_fixture)
 
-        contributor = get_contributor_by_connector_type(
-            SIRI_ET_XML_TN_CONTRIBUTOR_DB_ID, ConnectorType.siri_et_xml_tn.value
-        )
-
+        contributors = Contributor.find_by_connector_type(ConnectorType.siri_et_xml_tn.value)
+        contributor = [c for c in contributors if c.id == SIRI_ET_XML_TN_CONTRIBUTOR_DB_ID][0]
```
pbougue, line 125: Same, you can probably use `next(filter(lambda))`

pbougue (commented):
Partial review still, focus on the refactor/inheritance is for later :sleeping_bed:

migrations/versions/2489aef96da8_add_siri_et_xml_tn.py
```diff
@@ -53,12 +43,18 @@ def downgrade():
...
+
+    op.execute("CREATE TYPE old_connector_type AS ENUM('cots', 'gtfs-rt', 'piv', 'siri-et-lite-idfm')")
+
+    op.execute("ALTER TABLE real_time_update ADD COLUMN old_connector old_connector_type")
+    op.execute("UPDATE real_time_update SET old_connector = connector::text::old_connector_type")
+    op.execute("ALTER TABLE real_time_update DROP connector")
+    op.execute("ALTER TABLE real_time_update RENAME COLUMN old_connector TO connector")
+
+    op.execute("ALTER TABLE contributor ADD COLUMN old_connector_type old_connector_type")
+    op.execute("UPDATE contributor SET old_connector_type = connector_type::text::old_connector_type")
```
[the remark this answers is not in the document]
vchabot: Only on the contributor table, as we have the migration merged with #426 (https://github.com/CanalTP/kirin/pull/426/files#diff-6a32d92142830e8f12c994565153b67e0876cc12ab4c255f30190822e388924f)

kirin/queue_workers/piv/__init__.py
```diff
@@ -28,5 +28,3 @@
 # [matrix] channel #navitia:matrix.org (https://app.element.io/#/room/#navitia:matrix.org)
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
-
-from kirin.piv.model_maker import KirinModelBuilder
```
woshilapin, line 32: So we don't export the `ModelBuilder` in the `__init__` anymore? This is what's done in `poller_workers`. That said, when I used `autoflake`, it was mentioned as an unused `import` and was sanitized. I'm not sure which way to go but at least, I believe we should probably have something consistent with `poller_workers`.

kirin/queue_workers/piv/piv.py
```diff
@@ -0,0 +1,45 @@
...
+from kirin.queue_workers.http_utils import post_feed, get_contributors_by_connector_type
+
+
+class PivIndex(Resource):
+    def get(self):
+        return get_contributors_by_connector_type(ConnectorType.piv.value)
+
+
+class Piv(Resource):
+    def post(self, contributor_id=None):
```
woshilapin, line 44: Do we really need a default value for `contributor_id`? What does it mean to call `post` without a `contributor_id`? Actually, if I understand well, this will lead to an `abort(400)`. Maybe we should remove the default value.
```suggestion
    def post(self, contributor_id):
```

That said, we also have the following comment in `kirin/queue_workers/piv/piv_worker.py`.

```python
    # We assume a maximum of one PIV contributor is going to exist in the DB
    # Therefore, we enforce a maximum of one `PivWorker` is running at a time
```

kirin/queue_workers/piv/piv_worker.py
```diff
@@ -0,0 +1,89 @@
...
+class PivWorkerLauncher(AbstractQueueWorkerLauncher):
+    def get_connector_type(self):
+        return ConnectorType.piv.value
+
+    def process(self, contributor: Contributor) -> bool:
+        with PivWorker(contributor) as worker:
+            should_wait = False  # wait only after init crash
+            logger.info("launching the piv worker for '{0}'".format(contributor.id))
+            worker.run()
+            return should_wait
```
woshilapin, line 81: I wonder if we could improve this code design. `should_wait` has collateral effects, which makes it hard to understand at its role first sight. I don't know if it's coming from your refactor or if it was already here, but it'd be a good time to improve it. Just a (possible dumb) idea would be that the method `.run()` takes `should_wait` as a parameter.

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
```diff
@@ -0,0 +1,51 @@
...
+from kirin.utils import make_rt_update
+
+
+class KirinModelBuilder(AbstractKirinModelBuilder):
+    def __init__(self, contributor):
+        super(KirinModelBuilder, self).__init__(contributor)
+
+    def build_rt_update(self, input_raw):
+        rt_update = make_rt_update(
+            str(input_raw), connector_type=self.contributor.connector_type, contributor_id=self.contributor.id
```
woshilapin, line 42: I'm seeing you're using `str(input_raw)` instead of just `input_raw` here which is probably good. Should we do the same in `kirin/queue_workers/piv/piv_model_maker.py`?

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_worker.py
```diff
@@ -0,0 +1,93 @@
...
+        super().process_message(body, message)
+
+    @new_relic.agent.background_task(name="siri_et_xml_tn_worker-on_iteration", group="Task")
+    def on_iteration(self):
+        super().on_iteration()
+
+
+class SiriEtXmlTnWorkerLauncher(AbstractQueueWorkerLauncher):
+    def get_connector_type(self):
+        return ConnectorType.siri_et_xml_tn.value
```
woshilapin, line 77: Still my [rusty](https://www.rust-lang.org/) site speaking, but how about returning the type `ConnectorType` instead of a `str` which is less typed. Being typed on the interface of the API can usually help catch so logical bug inside the code.

kirin/queue_workers/__init__.py
```diff
@@ -84,14 +94,13 @@ def __init__(self, contributor):
         if not contributor.queue_name:
             raise ValueError("Missing 'queue_name' configuration for contributor '{0}'".format(contributor.id))
         self.last_config_checked_time = datetime.now()
-        self.builder = KirinModelBuilder(contributor)
+        self.builder = builder
```
woshilapin, line 97: In `poller_workers`, I had an hesitation to have the `builder` as part of the object (and therefore in the constructor). I chose to make it a parameter of the [`poll()` method](https://github.com/CanalTP/kirin/blob/master/kirin/pollers/__init__.py#L153). I know `poller_workers` and `queue_workers` are different, but I wonder if the difference is enough to explain this different choice of architecture or if we should make them consistent.

kirin/queue_workers/__init__.py
```diff
@@ -84,14 +94,13 @@ def __init__(self, contributor):
...
         # store config to spot configuration changes
         self.broker_url = deepcopy(contributor.broker_url)
         self.navitia_coverage = deepcopy(contributor.navitia_coverage)
         self.navitia_token = deepcopy(contributor.navitia_token)
         self.last_lock_update = datetime.now()
 
-    @new_relic.agent.background_task(name="piv_worker-enter", group="Task")
```
woshilapin, line 94: Is it not possible to make something like this (I'm totally making this up, and I truly have no idea if such thing would work).

```python
     @new_relic.agent.background_task(name="{}_worker-enter".format(self.connector_type), group="Task")
```

kirin/queue_workers/__init__.py
```diff
@@ -156,71 +165,84 @@ def on_iteration(self):
...
+    def get_contributor_by_connector_type(self):
+        """
+        Get a contributor by its ID and the associated connector type
+        """
+        contributors = [
+            c
+            for c in Contributor.find_by_connector_type(self.connector_type)
+            if c.id == self.builder.contributor.id
+        ]
+        return contributors[0] if contributors else None
```
woshilapin, line 184: This seems almost like a duplicate from what is in `http_utils.py`, maybe they can be merged together.

https://github.com/CanalTP/kirin/pull/429/files#diff-bda5ddc5793e32a0ac229c179dcdc65e4afed38d8e6d439a6a5d965b83446831R19

kirin/queue_workers/__init__.py
```diff
@@ -156,71 +165,84 @@ def on_iteration(self):
...
+
+    def pick_first_contributor_by_connector_type(self):
+        contributors = Contributor.find_by_connector_type(self.connector_type)
+        if len(contributors) > 1:
+            self.logger.warning(
+                "more than one {0} contributors: {1}; choosing '{2}'".format(
+                    self.connector_type, map(lambda c: c.id, contributors), contributors[0].id
+                )
+            )
+        return contributors[0] if len(contributors) > 0 else None
```
woshilapin, line 248: Here again, I'm wondering if there is an opportunity for factorizing with https://github.com/CanalTP/kirin/pull/429/files#diff-bda5ddc5793e32a0ac229c179dcdc65e4afed38d8e6d439a6a5d965b83446831R19.

tests/integration/siri_et_xml_tn_worker_test.py
```diff
@@ -0,0 +1,189 @@
+# coding: utf8
```
woshilapin, line 1: It seems like most of the tests in here are not really specific to `siri-et-xml-tn` or `piv`. Should we factorize them in a `queue_worker_test.py`

kirin/queue_workers/__init__.py
vchabot: That's what we asked ourselves yesterday with @pbougue, but I'm not sure this would be more readable, and I don't know if it would work.

kirin/queue_workers/piv/piv_worker.py
vchabot: We talked about this piece of code yesterday with @pbougue. It was already here before, but it should be refactored to a more readable code. I will try your solution.

kirin/queue_workers/__init__.py
```diff
@@ -1,7 +1,6 @@
-#!/usr/bin/env python
 # coding=utf-8
 
-# Copyright (c) 2020, Canal TP and/or its affiliates. All rights reserved.
+# Copyright (c) 2001, Canal TP and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
vchabot: Only 4 files have the 2020 copyright, so I will update these 4 files to change them to a 2001 copyright.

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
vchabot: Yes, I added the `str` cast before pushing the first commit, because I saw you did the same on the json connector. It's not necessary and I'll remove it.

woshilapin: Isn't it, I thought it was necessary actually to escape problematic characters and avoid blowing up the `INSERT` in DB.

kirin/queue_workers/piv/piv_worker.py
woshilapin: OK, it seems I was not correct on understanding how `should_wait` actually is used. Your PR doesn't really make it worse than before, so I propose we do not address the problem now. However, we might need to come back to it to improve the understandability of this piece of code in another PR at some point.

kirin/queue_workers/__init__.py
```diff
@@ -30,50 +29,61 @@
...
+    """
+    Class factorizing all the common functionality between queue_workers.
+    """
 
-class PivWorker(ConsumerMixin):
-    @new_relic.agent.background_task(name="piv_worker-init", group="Task")
-    def __init__(self, contributor):
-        if contributor.connector_type != ConnectorType.piv.value:
+    def __init__(
+        self, logger, contributor: Contributor, connector_type: str, builder: AbstractKirinModelBuilder
```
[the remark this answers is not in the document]
pbougue: Agreed on typing, but it should not be a parameter, as real implems of this abstract don't offer choice on this, they should provide a "get_connector_type" that returns this and should be used at init, I guess.

kirin/queue_workers/piv/piv.py
pbougue: If flask output's a message on missing contributor that's clear, I'm OK to remove the default value.

Removing the whole param doesn't seem like a good path, as the idea is more to provide handling of multiple contributors for the same connector than the opposite (the need just doesn't exist currently on those like it exists for GTFSRT... and it's a important architectural work to do).
It's also more consistent with GTFSRT.
Finally, it's about to be refactored in a separate PR to be handled once and for all (if possible).

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
pbougue: It's actually a `bytes` in `make_rt_update` typing (and I guess in sqlalchemy too).
So it might be better to be explicit on that right here from the start?

kirin/queue_workers/__init__.py
pbougue: Generally speaking, we don't want to re-instanciate KirinModelBuilder for each message that we process for performances.
So here, it's better to keep it associated to the worker that lives long.

That said, on celery workers, it's not as easy as the task is just a function so we re-instanciate at each processing (but a processing is very less frequent than for PIV). We might store Builders in some sort of cache, but the difference here is probably not so important (although it would save a navitia call and I did not check the rest).

kirin/queue_workers/__init__.py
pbougue: IMO the only thing common is the use of find_by_connector_type, no?
Here we search for a precise contributor while in `http_utils` we want to list them all an format the result.

We could probably shorten/pythonify this too (I guess it returns `None` if no match is found, did not check):
```python
        return next(filter(lambda c: c.id == self.builder.contributor.id, Contributor.find_by_connector_type(self.connector_type)))
```

kirin/queue_workers/__init__.py
pbougue: We may factorize a bit by offering the possibility to pass no contributor_id.
But it's in self... so depending on if `self.contributor` is None, maybe?

tests/integration/siri_et_xml_tn_worker_test.py
pbougue: And launch them for both with a parametrized test ? Why not.
Please note that they are quite "large" and test also that de KirinModelBuilder does things (not super smart things but still).

kirin/queue_workers/__init__.py
vchabot: It does not work, as it does not recognize self, nor `connector_type`

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
woshilapin: Now that I see every `make_rt_update` being called with `str(input_raw)`, I wonder if this should be internalized inside the `make_rt_update` directly?

woshilapin:
It seems that Sonar is complaining about duplicated block between https://github.com/CanalTP/kirin/blob/siri_et_xml_tn/kirin/queue_workers/piv/piv_model_maker.py#L657-L678 and https://github.com/CanalTP/kirin/blob/siri_et_xml_tn/kirin/poller_workers/gtfs_rt/gtfs_rt_model_maker.py#L459-L481. It might be a good idea to actually follow the lead of Sonar on this and factorize the code.

woshilapin (approved):
Good for me. There is a merge conflict on your PR though.

tests/integration/siri_et_xml_tn_worker_test.py
```diff
@@ -0,0 +1,189 @@
...
+        self.thread.start()
+        wait_until(lambda: self.thread.is_alive())
+        # Check that SiriEtXmlTnWorker is ready (a good hint is when queue is created)
+        wait_until(lambda: is_queue_created(self.broker_connection, SIRI_ET_XML_TN_QUEUE_NAME))
+
+    def __exit__(self, type, value, traceback):
+        # Remove the contributor
+        self.test_client.delete("/contributors/{}".format(SIRI_ET_XML_TN_CONTRIBUTOR_DB_ID))
+        # SiriEtXmlTnWorker should die eventually when no SIRI_ET_XML_TN contributors is available
+        wait_until(lambda: not self.thread.is_alive())
```
pbougue, line 151: All the above is duplicate with PIV and should be mutualized (parametrized)

pbougue (commented):
Full review.
Let's split the work to manage it!

kirin/queue_workers/http_utils.py
```diff
@@ -0,0 +1,42 @@
+from typing import Tuple
+
+import flask
+from flask import url_for, Request
+from flask_restful import abort
+
+from kirin.core.build_wrapper import wrap_build
+from kirin.core.model import Contributor
+from kirin.exceptions import InvalidArguments
+from kirin.queue_workers.piv.piv_model_maker import KirinModelBuilder
```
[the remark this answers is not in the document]
vchabot: For now on this PR, only piv uses this, that's why I left this here.

kirin/core/model.py
```diff
@@ -576,6 +576,10 @@ def find_by_connector_type(cls, type, include_deactivated=False):
             return cls.query.filter_by(connector_type=type).all()
         return cls.query.filter_by(connector_type=type, is_active=True).all()
 
+    @classmethod
+    def find_one_by_connector_type_and_id(cls, type, contributor_id=None):
+        return Contributor.query_existing().filter_by(id=contributor_id, connector_type=type).first()
```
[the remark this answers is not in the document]
vchabot: The `query_existing` already filters only active contributors ( `filter_by(is_active=True)` )

kirin/queue_workers/__init__.py
```diff
@@ -119,108 +128,112 @@ def get_consumers(self, Consumer, channel):
...
+        """
 
         # SQLAlchemy is not querying the DB for read (uses cache instead),
         # unless we specifically tell that the data is expired.
         db.session.expire(self.builder.contributor)
-        self.last_config_checked_time = datetime.now()
 
-        contributor = get_piv_contributor(self.builder.contributor.id)
-        if (
+        contributor = Contributor.find_one_by_connector_type_and_id(
```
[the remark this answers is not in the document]
vchabot: Yes I do, the `query_existing` inside `find_one_by_connector_type_and_id` filters only active contributors.

kirin/queue_workers/http_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        c.id: {"href": url_for(connector_type, contributor_id=c.id, _external=True)} for c in contributors
+    }
+    return response, 200
+
+
+def post_feed(contributor_id: str, connector_type: str) -> Tuple[dict, int]:
+    if contributor_id is None:
+        abort(400, message="Contributor's id is missing")
+
+    contributor = Contributor.find_one_by_connector_type_and_id(connector_type, contributor_id)
```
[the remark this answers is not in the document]
woshilapin: I don't change this part of the code at all. So I'm guessing that any fix in the current PR should be good for #433?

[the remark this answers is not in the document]
pbougue: Yep, after discussion, query_existing is doing the job already (my bad :pray: )
And it's OK that it doesn't change in #433

kirin/queue_workers/http_utils.py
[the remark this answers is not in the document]
pbougue: OK, this is gonna be removed in #433, let's leave it like that.

kirin/queue_workers/__init__.py
```diff
@@ -119,108 +128,118 @@ def get_consumers(self, Consumer, channel):
...
 
-def try_db_commit():
-    try:
-        db.session.commit()
-    except Exception as db_e:
-        logger.warning("Exception while db-commit: {0}".format(db_e))
-        db.session.rollback()
 
+class QueueWorkerLauncher:
+    def __init__(self, logger: Logger, worker: Worker):
```
pbougue, line 187: I'm not sure that it works as expected: are you sure you rebuild it each time in `process()`
Otherwise we would end up in a situation where the config is not updated.
That's why I suggested passing the `class` itself, not an instance.

kirin/queue_workers/__init__.py
```diff
@@ -119,108 +131,113 @@ def get_consumers(self, Consumer, channel):
...
-    while True:
+                self.attempt_worker_launch()
+            except LockError as lock_e:
+                self.logger.warning("Exception with lock : {0}".format(lock_e))
+
+    def process(self) -> bool:
+        contributors = Contributor.find_by_connector_type(self.connector_type.value)
+        contributor = contributors[0] if len(contributors) > 0 else None
+        if not contributor:
+            raise NoContributorException
```
pbougue, line 208: You can probably avoid the raise here (and the try/except/return wrapping it):
```suggestion
            self.logger.warning("no contributor for {}".format(self.connector_type.value))
            return True  # Wait before next check that contributor exists
```

kirin/queue_workers/__init__.py
```diff
@@ -119,108 +131,113 @@ def get_consumers(self, Consumer, channel):
...
+        with self.worker_class(contributor) as worker:
+            should_wait = False  # wait only after init crash
+            self.logger.info(
+                "launching the {0} worker for '{1}'".format(worker.get_connector_type().value, contributor.id)
+            )
+            worker.run()
+            return should_wait
```
pbougue, line 215: To match suggestion above:
```suggestion
            return False  # No wait when run() was interrupted quietly (because a config changed)
```

kirin/queue_workers/__init__.py
```diff
@@ -119,108 +131,113 @@ def get_consumers(self, Consumer, channel):
...
-    # We assume a maximum of one PIV contributor is going to exist in the DB
-    # Therefore, we enforce a maximum of one `PivWorker` is running at a time
-    while True:
+                self.attempt_worker_launch()
+            except LockError as lock_e:
+                self.logger.warning("Exception with lock : {0}".format(lock_e))
+
+    def process(self) -> bool:
+        contributors = Contributor.find_by_connector_type(self.connector_type.value)
+        contributor = contributors[0] if len(contributors) > 0 else None
```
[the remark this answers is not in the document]
pbougue: OK, hand-tested that what you did works OK, and it's also doable to widen it to line 172 of the file, to be done in #433

kirin/exceptions.py
```diff
@@ -86,3 +86,7 @@ class UnauthorizedOnSubService(SubServiceError):
 class UnsupportedValue(KirinException):
     code = 400
     message = "unsupported value"
+
+
+class NoContributorException(KirinException):
```
pbougue, line 91: To be removed in #433

pbougue (approved):

PR #431, opened 2021-12-03
fix: add 'SET NULL' constraint to migration

woshilapin:
Should address comment https://github.com/CanalTP/kirin/pull/428#discussion_r761918635. I've applied the `NOT NULL` constraint on `dev` DB manually too (so this migration should not be played again on next deployment).


pbougue (changes requested):
This should only be part of #426, it's fully related and only related to #426 PR (and #426 PR won't work without it).

pbougue (approved):
Ahhh, the mixup was partly in my head on this, it's better with this indeed.

migrations/versions/1639faa28bc4_add_siri_et_lite_idfm.py
```diff
@@ -49,11 +49,13 @@ def downgrade():
 
     op.execute("ALTER TABLE real_time_update ADD COLUMN old_connector old_connector_type")
     op.execute("UPDATE real_time_update SET old_connector = connector::text::old_connector_type")
+    op.execute("ALTER TABLE realtime_update ALTER COLUMN old_connector SET NOT NULL")
```
pbougue, line 52:
```suggestion
    op.execute("ALTER TABLE real_time_update ALTER COLUMN old_connector SET NOT NULL")
```

migrations/versions/1639faa28bc4_add_siri_et_lite_idfm.py
```diff
@@ -49,11 +49,13 @@ def downgrade():
...
     op.execute("ALTER TABLE real_time_update DROP connector")
     op.execute("ALTER TABLE real_time_update RENAME COLUMN old_connector TO connector")
 
     op.execute("ALTER TABLE contributor ADD COLUMN old_connector_type old_connector_type")
     op.execute("UPDATE contributor SET old_connector_type = connector_type::text::old_connector_type")
+    op.execute("ALTER TABLE realtime_update ALTER COLUMN old_connector_type SET NOT NULL")
```
pbougue, line 58:
```suggestion
    op.execute("ALTER TABLE real_time_update ALTER COLUMN old_connector_type SET NOT NULL")
```

migrations/versions/1639faa28bc4_add_siri_et_lite_idfm.py
woshilapin: :facepalm:

migrations/versions/1639faa28bc4_add_siri_et_lite_idfm.py
woshilapin: :facepalm:

PR #433, opened 2021-12-10
feat!: standardize /feeds web API

woshilapin:
Changes all the web API to send a feed and to get the contributors for a certain type of feeds.

* `POST /contributors/<contributor_id>/feeds[?data_format=[json|text|pbf]]`
* ~`GET /feeds/<feed_type> -> [Contributor]` (note: `feed_type` can be any of `gtfs_rt`, `cots`, `piv`, `siri_et_lite_idfm` or `siri_et_xml_tn`)~


vchabot (approved):

kirin/queue_workers/http_utils.py
```diff
@@ -16,18 +16,6 @@ def get_request_data(request: Request, connector_type: str) -> str:
...
-    if not contributors:
-        return {"message": "No {} contributor defined".format(connector_type)}, 200
-
-    response = {
-        c.id: {"href": url_for(connector_type, contributor_id=c.id, _external=True)} for c in contributors
-    }
-    return response, 200
-
-
 def post_feed(contributor_id: str, connector_type: str) -> Tuple[dict, int]:
```
vchabot, line 19: It seems like this is not used anymore as you're using your own function inside `kirin/resources/feeds.py`

woshilapin: I entirely removed `http_utils.py` since it's not used anymore. Thank you for the hint.

kirin/resources/feeds.py
```diff
@@ -0,0 +1,156 @@
...
+
+@unique
+class DataFormat(Enum):
+    def _generate_next_value_(name, start, count, last_values):
+        return name
+
+    PBF = auto()
+    TEXT = auto()
+    JSON = auto()
+    XML = auto()
```
pbougue, line 56: If this is maintained, casing consistency with the rest of enums in codebase should be checked.

pbougue (commented):
As discussed, I will try to handle those :snowboarder:

pbougue (commented):
Some factorization will probably be required with `Worker/WorkerLauncher.get_connector_type` when #429 is finished.

kirin/poller_workers/gtfs_rt/gtfs_rt_model_maker.py
```diff
@@ -104,6 +140,28 @@ def __repr__(self):
         """Allow this class to be cacheable"""
         return "{}.{}.{}".format(self.__class__, self.navitia.url, self.instance_data_pub_date)
 
+    @classmethod
+    def convert_source_to_target_format(cls, source_feed, source_format):
+        target_fmt = cls.SourceFormat.pbf
```
vchabot, line 145: In this method I see that you chose to use some short names. I'm not a big fan of short names, especially when you are not forced to use them (forced = like a really big variable name, for example).

An additional argument to this is you have some short names and long names alongside, and it's not really consistent.

kirin/resources/feeds.py
```diff
@@ -45,77 +42,22 @@
...
+    kirin.poller_workers.gtfs_rt.KirinModelBuilder,
+    kirin.poller_workers.siri_et_lite_idfm.KirinModelBuilder,
+    kirin.queue_workers.siri_et_xml_tn.siri_et_xml_tn_model_maker.KirinModelBuilder,
+]
 
-    proto = gtfs_realtime_pb2.FeedMessage()
-    json_data = ujson.loads(data)
-    json_format.ParseDict(json_data, proto)
-    return proto.SerializeToString()
+builder_cls_per_connector_type = {b.connector_type: b for b in builder_cls_list}
```
vchabot, line 53: As we discussed earlier, maybe another check to ensure a connector type has only one builder linked to it, to avoid a builder overwriting another. This is a nice to have, when someone will find the motivation.

vchabot (commented):
2 little things but I think we are almost good to go!

vchabot (approved):

kirin/core/abstract_builder.py
```diff
@@ -63,6 +64,25 @@ def __init__(self, contributor):
...
+    @property
+    def connector_type(self) -> ConnectorType:
+        """
+        format that's handled by builder
+        :return: ConnectorType enum value
+        """
+        raise NotImplementedError(self.NOT_IMPLEMENTED_ERROR)
+
+    @classmethod
+    def convert_source_to_target_format(cls, source_feed: Any, source_format: Optional[str]) -> Any:
```
woshilapin, line 77: I find the name of the function not good. When I see `convert`, I expect a source and a target, so mentioning them in the name is useless to me. I'd be more interested to know WHAT do we convert. So I'd say `convert_feed()` would be better (since `feed` seems to be the name we're using in the API)?

pbougue (approved):
This is kind of a self-review, but still :angel:

woshilapin:
If I could, I'd approve it too, but as the author, I can't. Therefore, I'll merge it.

PR #436, opened 2021-12-14
feat: add href links for '/feeds'

woshilapin:


woshilapin:
Getting nowhere in a reasonable amount of time for a feature which is not absolutely necessary. Cutting down the loss now.

PR #437, opened 2021-12-20
Small refactor for SIRI-TN connector

pbougue:
After comments in #429 

:mag: review by commit + commit's comments


kirin/core/model.py
```diff
@@ -422,6 +435,20 @@ def find_stop(self, stop_id, order=None):
             return first
         return next((st for st in self.stop_time_updates if st.stop_id == stop_id), None)
 
+    def copy_processed_attributes(self, other):
```
vchabot, line 438: You can add an explicit type for `other`

```python
def copy_processed_attributes(self, other: TripUpdate):
```

kirin/core/model.py
```diff
@@ -341,6 +349,11 @@ def __init__(
         self.headsign = headsign
         self.contributor_id = contributor_id
 
+        if self.effect is None and self.status is not None:
```
vchabot, line 352: Maybe add a comment to explain why we are doing this.

Effect and status are link to each other, if some one is None and the other isn't, we deduce the one from the other.

kirin/core/merge_utils.py
```diff
@@ -808,35 +808,41 @@ def adjust_trip_update_consistency(trip_update, stus):
         )
 
 
-def clean_trip_update(
+def build_trip_update_if_modified(
```
vchabot, line 811: :+1:

kirin/core/merge_utils.py
```diff
@@ -808,35 +808,41 @@ def adjust_trip_update_consistency(trip_update, stus):
...
-    """
-    old_tu = db_trip_update
-    if not old_tu:
-        old_tu = model.TripUpdate(vj=None, contributor_id=contributor_id)
-        old_tu.effect = get_effect_by_stop_time_status(old_tu.status)
+) -> Optional[model.TripUpdate]:
+    """
+    Build and return a consolidated TripUpdate (with its attached StopTimeUpdates) if
+    modifications appeared in new one
+    :param db_trip_update: TripUpdate obtained from db if previous realtime processing result was stored (or None)
```
vchabot, line 821: It seems like `db_trip_update` cannot be null (see the function signature). Maybe add an `Optional[model.TripUdpate]` type and a default value to `None` might be more consistent with your comment?

kirin/core/merge_utils.py
```diff
@@ -808,35 +808,41 @@ def adjust_trip_update_consistency(trip_update, stus):
...
+    :param old_stus: ordered STU list of previous realtime processing result (or corresponding list from base-schedule)
+    :param new_trip_update: TripUpdate from new feed being processed
+    :param new_stus: ordered STU list of new feed being processed
+    :param contributor_id: contributor's id
+    :return: TripUpdate with attached StopTimeUpdates if new is different from old (of from base-schedule),
```
vchabot, line 826: Looks like a typo

```suggestion
    :return: TripUpdate with attached StopTimeUpdates if new is different from old (or from base-schedule),
```

kirin/core/merge_utils.py
pbougue: Here I like to enforce caller to provide a value, so `Optional` is enough IMO (and required, yes!)

kirin/core/model.py
```diff
@@ -28,6 +28,7 @@
 # [matrix] channel #navitia:matrix.org (https://app.element.io/#/room/#navitia:matrix.org)
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+from __future__ import annotations
```
vchabot, line 31: Why do you need this import?

vchabot (approved):
LGTM :shipit:

kirin/core/model.py
pbougue: Annotating with type of enclosing class requires future feature: https://stackoverflow.com/a/33533514

PR #444, opened 2022-01-20
[tech] Improve dockerignore

pbougue:
Remove tests and cache files from docker build


.dockerignore
```diff
@@ -1,2 +1,14 @@
...
+**/.*
+!**/.git/
+**/*.pyc
+**/__pycache__
+tests/
+**/venv
+**/.env
+**/*_pb2.py.bak
+kirin/*_settings.py
+!kirin/default_settings.py
```
woshilapin, line 10: Why is there no `dev_settings.py` like in `.gitignore`?

.gitignore
```diff
@@ -3,9 +3,9 @@
 *_pb2.py
```
woshilapin, line 3: Should we also add `**/venv` like in `.gitignore`?

.dockerignore
```diff
@@ -1,2 +1,14 @@
-*.pyc
-venv/**
+**/.*
+!**/.git/
+**/*.pyc
+**/__pycache__
+tests/
+**/venv
+**/.env
```
woshilapin, line 7: Do we need this since we already do have `**/.*` above?

.dockerignore
```diff
@@ -1,2 +1,14 @@
-*.pyc
-venv/**
+**/.*
+!**/.git/
+**/*.pyc
+**/__pycache__
```
woshilapin, line 4: Should this be also in `.gitignore`?

.dockerignore
```diff
@@ -1,2 +1,14 @@
-*.pyc
-venv/**
+**/.*
```
woshilapin, line 1: Sounds a bit too wide. I think I would rather go for an exhaustive solution... although not a strong opinion on that.

.dockerignore
pbougue: `dev_settings` is ignored (contains developper's info), that's what we want.
`test_settings` is also ignored as we don't want to test anything using this image.

.dockerignore
pbougue: OK, I'll add this... This is probably ignored by each dev's general gitignore file, currently.

.dockerignore
woshilapin: Whoah, I meant `test_settings.py`. I actually just compared both `.dockerignore` and `.gitignore` and they're different... but now, I understand why it's not ignored in `.gitignore` but still ignored in `.dockerignore`. Sorry about the noise.

.dockerignore
pbougue: I checked on my computer and thought "better safe than sorry" but it's open to discussion (I have `.idea`, `.vscode` ...).

woshilapin: Yes, I see since there is an opportunity for leaking sensitive data into a Docker image, I think I do agree with you on that one. Let's keep it like it is then!

.dockerignore
```diff
@@ -1,2 +1,13 @@
...
+!**/.git/
+**/*.pyc
```
woshilapin, line 3: This can probably be removed too because of the above rule `**/.*`.

woshilapin (approved):

PR #450, opened 2022-03-21
feat: siri-et-xml-tn trip removal

woshilapin:
Add support for trip removal in the siri-et-xml-tn feed.

There is still some missing work about error handling, I'll work on that.

ref: NAV-1104


kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
```diff
@@ -47,12 +65,140 @@ def convert_feed(cls, source_feed, source_format):
...
+        line_ref = splits[3]  # TODO: IndexError
+        return "FR1:Line:{}:".format(line_ref)
+
+    def _get_since_until_dt(self, feed_vj):
+        origin_time = str_to_utc_naive_dt(feed_vj.find("./OriginAimed-DepartureTime").text)  # TODO: ValueError
+        since_dt = origin_time - timedelta(minutes=60)
+        until_dt = origin_time + timedelta(minutes=60)
+        return since_dt, until_dt
+
+    def _request_navitia_vj(self, vj_name, line_ref, since_dt, until_dt):
```
kadhikari, line 125: This method can be externalized as it is used in other model_maker. We can pass filter, since, until, data_freshness, depth (default=2).

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
```diff
@@ -47,12 +65,140 @@ def convert_feed(cls, source_feed, source_format):
...
+                "filter": "vehicle_journey.name={} and line.has_code(source, {})".format(vj_name, line_ref),
+                "since": to_navitia_utc_str(since_dt),
+                "until": to_navitia_utc_str(until_dt),
+                "data_freshness": "base_schedule",
+                "depth": "2",  # we need this depth to get the stoptime's stop_area
+            }
+        )
+        return navitia_vjs
+
+    @app.cache.memoize(timeout=1500)
```
kadhikari, line 151: If it's in second = 1500/60 = 25 minutes, i think it's a huge value. we can also use configured value non?

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
```diff
@@ -47,12 +65,140 @@ def convert_feed(cls, source_feed, source_format):
...
+    def _get_navitia_vj(self, vj_name, line_ref, since_dt, until_dt):
```
kadhikari, line 152: Same here, this method can also be externalized.

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
```diff
@@ -47,12 +65,140 @@ def convert_feed(cls, source_feed, source_format):
...
+        )
+        vj_name = self._get_vj_name(feed_vj)
+        line_ref = self._get_line_ref(feed_vj)
+        since_dt, until_dt = self._get_since_until_dt(feed_vj)
+        vj = self._get_navitia_vj(vj_name, line_ref, since_dt, until_dt)
+        if vj is None:
+            return None
+
+        trip_update = model.TripUpdate(vj=vj, contributor_id=self.contributor.id)
+        if feed_vj.find("./FramedVehicleJourneyRef/Cancellation").text == "true":  # TODO: AttributeError
```
kadhikari, line 197: This verification should better be done just after getting the value of 'feed_vj' to avoid unnecessary treatment.

kadhikari (commented):
Errors handling for AttributError, IndexError, ...

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
woshilapin: In theory, yes… but I'm thinking we're going to have a lot of other use cases that just vehicle journey cancellation. Some of them may need information from the `trip_update` to be processed. I'd rather keep all the parts of  “processing the business value of the feed” at the same place here, instead of having some above (as early return) and some below. Does that make any sense?

kirin/queue_workers/siri_et_xml_tn/siri_et_xml_tn_model_maker.py
woshilapin: > If it's in second = 1500/60 = 25 minutes, i think it's a huge value. 

I basically copy/paste this value from other model makers, so I didn't even realize. That said, I believe this value is the time of retention of the cache not a timeout to query a value from the cache (although, the documentation is not very clear about that :shrug:).

> we can also use configured value non?

We could indeed put that in a configuration value, I don't see that value being specific to one connector or another.

woshilapin: I was able to find documentation about `memoize` and it is indeed the duration until which the cache is going to be renewed. Usually, this kind of parameter is named TTL for Time-To-Live. Using `timeout` here is misleading... but it is the same.
https://github.com/pallets-eco/flask-caching/blob/e3440f3601061e3732f920a59f31d154c11d73ba/src/flask_caching/__init__.py#L850-L851

kadhikari (approved):
:+1:

PR #452, opened 2022-03-25
refactor: instantiate logger in AbstractKirinModelBuilder

woshilapin:
When querying navitia, kirin do not need disruptions. Disabling
disruptions have shown significant improvement in response time (from
1.2s down to 500ms).

Must be merged after #451.


woshilapin:
Let's hold on this evolution for now. Maybe @pbougue will help us evaluate if this is a good idea or not.

